Issue #1056, opened 2015-01-07
Implement a way of data clean-up for a particular survey


iperdomo:
The current approach for deleting collected data related to a survey is using the _test harness_ with the action `deleteSurveyInstances`. This doesn't process doesn't work, see #730 

We can just execute the same process from a _data script_ program. Eventually, we need to recompute counters, but since this process is followed by a data import, the latter takes care of it.

PR #1057, opened 2015-01-07
[#1056] Deletes collected data for a particular survey id

iperdomo:


iperdomo:
Sample parameters call:
````
org.akvo.gae.remoteapi.DeleteData localhost test@example.com yyy 123
`````

jonase:
:+1:

Issue #1059, opened 2015-01-08
Identifier search not working on monitoring tab


mtwestra:
When a user enters an identifier in the identifier search field in the Monitoring tab, the server responds with a 500 internal server error, quoting a missing datastore index for this search.

Proposed solution: add identifier index to datastore-indexes.xml

mtwestra:
Example: 
instance: ctbburundi
survey: BDI_BISEM_VF
a data point with identifier chqg-0r6r-g7an exists, but is not found.

Cause: the first time 'Find' is clicked, a cursor is added to the search, which makes the search fail

mtwestra:
Test plan:
1) Go to the monitoring tab, select a survey that has a lot of data (more than a screenful), and click 'Find'
2) Copy one of the identifiers of the list in the identifier list, and click 'Find'
3) The right element should be found straight away.

rumca:
@mtwestra the search functionality on this tab seems to have been broken. I can see Monitored data being submitted on the "Inspect Data" tab and can export a raw data report for the forms, however when I am on the "Monitoring" tab and select the survey - nothing is returned once I select search.

rumca:
scratch that, identifier search works fine - it's the drop downs which don't work.

PR #1073, opened 2015-01-13
[#1059] Don't set since cursor parameter

mtwestra:
* When 'Find' is clicked, we don't need to set the cursor (the 'since' parameter).


iperdomo:
:+1:

PR #1077, opened 2015-01-14
Feature/1012 geoshape question type

mtwestra:


jonase:
License/Copyright notice is missing from the js files and the Java files needs to have the notice updated to include 2015

Issue #325, opened 2013-08-01
Search results should be returned considering the "To Date " to be inclusive


joycarpediem:
Currently the To date provided in the search parameters in exclusive.
So if i want to fetch all records in till  31-Jul-2013, the "To" Date that i've to provide is 1-Aug-2013 which is not intuitive to the user.

We could append 23:59:59 to the To Date while firing the query or as Mark suggested increment the ToDate by one.

rumca:
Searching todays date as the "To" date brings up today's results as expected - passes

PR #1080, opened 2015-01-14
[#325] Add a day to the end date search

mtwestra:
* by default, when you pick a day using the 
date picker, it has a time of 00:00. This means
that searching for the period 2014-1-1 to 2014-1-5
will actually search 2014-1-1 to 2014-1-4 inclusive.
As this is counter intuitive, we make the end date 
inclusive by adding 24 hours.


jonase:
:+1:

Issue #730, opened 2014-08-19
deleteSurveyInstance tries too many things and fails in the attempt


iperdomo:
The method `deleteSurveyInstance` from `SurveyInstanceDAO` tries to have a synchronous execution and control too many things that fails with a relatively _small_ dataset (~1K SurveyInstances). This makes the attempt to clearing out the data related to a particular `Survey` a real pain.

muloem:
### Test Plan

**Pretest prep:**
Identify or create existing data/response for a survey.  The survey should  comprise a `GEO` question and an `OPTION` type question. 

**Test Execution:**
1.  Select one of the responses under the data tab and note the response values for the geo location and the option question. Under the reports tab, generate a chart for the option question and note the total responses indicated, as well as the percentage of the total response corresponding to the specific option noted earlier.  Under the maps tab try to identify the point and/or cluster under which the responses noted earlier falls.

2. Go back to the data tab and delete the response

**Expected:**

* Under the reports tab, the total values shown by the chart of the option question should be updated correctly (i.e. one less).  Also the percentage for the specific option value should be recomputed appropriately.

* Under the map, the clusters that are close to the identified point should have been updated to indicate that a point has been removed.

rumca:
The above test plan passes - clustering needs a page refresh to update correctly but this is expected :+1:

PR #1086, opened 2015-01-16
[#730] Refactor delete survey instance functionality

muloem:
* Adds a new queue for updating summary counts
* Delete a `SurveyedLocale` when there are no more related survey instances
* Remove test harness call

### DIFF
--- a/GAE/src/com/gallatinsystems/gis/map/MapUtils.java
+++ b/GAE/src/com/gallatinsystems/gis/map/MapUtils.java
@@ -33,6 +33,7 @@ import com.gallatinsystems.survey.domain.Survey;
 import com.gallatinsystems.survey.domain.SurveyGroup;
 import com.gallatinsystems.survey.domain.SurveyGroup.PrivacyLevel;
 import com.gallatinsystems.surveyal.dao.SurveyedLocaleClusterDao;
+import com.gallatinsystems.surveyal.dao.SurveyedLocaleDao;
 import com.gallatinsystems.surveyal.domain.SurveyedLocale;
 import com.gallatinsystems.surveyal.domain.SurveyedLocaleCluster;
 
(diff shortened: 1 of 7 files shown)


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

GAE/src/com/gallatinsystems/gis/map/MapUtils.java
```diff
@@ -174,6 +175,11 @@ public static void recomputeCluster(Cache cache, SurveyedLocale locale, int delt
             }
         }
 
+        // delete locale if the Delta was a subtraction
+        SurveyedLocaleDao slDao = new SurveyedLocaleDao();
+        if (delta < 0) {
```
iperdomo: Are we sure we want to delete a `SurveyLocale` on each substraction? I think we need to check for available `SurveyIntance` for this locale?

iperdomo:
We need to update the copyright header for those modified files

GAE/src/com/gallatinsystems/gis/map/MapUtils.java
muloem: The deletion only happens when the deleted `SurveyInstance` is the only one connected to the `SurveyedLocale`

iperdomo: This code is aware of that? The comment states:

     // delta can be +1 or -1, depending on if we want to add a point or delete a point.

So `delta` can be `-1`, but a `SurveyedLocale` can have more `SurveyInstance`s ?

muloem: Yep. The task with the -1 delta is created here:
https://github.com/akvo/akvo-flow/pull/1086/files#diff-a521baebc25d5bbfb464614e3eb1f5daR684

See comment.

iperdomo:
:+1:

Issue #1072, opened 2015-01-13
Cascading questions are not shown in preview


mtwestra:
Cascading questions are not yet part of the survey preview.

mtwestra:
Test plan:
1) create a cascade resource, with a number of levels
2) create a survey including a cascade question with this resource
3) show preview of survey
4) the preview should display the cascade question, with each level shown as a text item plus empty text field.

mtwestra:
At the moment, we do not use the values of the text fields.

rumca:
Tested and works as intended on release/1.8.1 :+1:

PR #1092, opened 2015-01-19
[#1072] Show cascade questions in preview

mtwestra:
* shows cascade levels as simple text fields


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

mtwestra:
@iperdomo I think we should show some kind of input field there? I agree a textfield is not ideal, but I don't quite know an alternative, apart from implementing it completely with dropdowns which are filled from the cascade definition.

Second comment: 
As far as I know, we initialize the cascade names to a default ['level1', 'level3', ... 'level10'] (something like that), and use numLevel to keep track of the number actually used. But perhaps this was already refactored to make the numLevel uncessary?

iperdomo:
:+1: code reviewed

Issue #1153, opened 2015-02-17
Correct `SurveyGroup` and `Survey` paths via RemoteAPI


muloem:
In the current implementation of user roles and permissions #818, the restriction on user permissions relies heavily on the presence of a correct `path` property in the `SurveyGroup` and `Survey` entities.  Without a correct `path` a number of issues are experienced e.g folders are not visible, users are not able to create new folders/surveys. Due to a number of issues reported regarding visibility of folders when assigning roles and permissions, we create this script to correct the `path` property for all `SurveyGroup` and `Surveys` on an instance.

PR #1154, opened 2015-02-17
[#1153] Correct `Folder`, `Survey`, and `Form` path

muloem:


scripts/data/correct_folder_survey_path.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
+
+# Assumes the script is executed in the /path/to/scripts/data/ directory
+# and the sources have been compiled in a bin directory also located
+# in the /path/to/scripts/data directory
+
+java -cp bin:lib/appengine-api-1.0-sdk-1.8.8.jar:lib/appengine-remote-api.jar org.akvo.gae.remoteapi.RemoteAPI CorrectFolderSurveyPath $1 "$2" "$3"
```
iperdomo: We could add more dependencies to the `lib` folder. I suggest we do:

````java -cp bin:lib/* org.akvo.gae.remoteapi.RemoteAPI <rest>````

PR #1167, opened 2015-02-27
Issue/1117 update cljs deps

jonase:
In addition to updating to latest versions of all dependencies this changeset simplifies the build process enabled by recent additions to cljs compiler. Thanks to the `:main` entry point we no longer need two versions of the `html` file. The only transformation needed is to load `users.min.css` in production builds which is trivial enough to be done with a simple `sed` command.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Dashboard/app/cljs/cp-html.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+OUT_ROOT=../../../GAE/war/admin/frames/
+
+mkdir -p ${OUT_ROOT}
+
+if [[ $* == *--production* ]]; then
+    cat users.html | sed 's/users\.css/users\.min\.css/' > ${OUT_ROOT}users.html
```
iperdomo: There is no need to `cat <file> | sed '<expr>'`, simply just used `sed` in the original file: `sed '<expr>' <file>`

Dashboard/app/cljs/users.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
iperdomo: Is the `lang` attribute changed when with change language to `fr` or `es` ?

iperdomo:
@jonase the `*.cljs` files require a copyright header update

iperdomo:
:+1: code reviewed

Issue #1175, opened 2015-03-03
Locate and correct visibility of private data points


muloem:
Some data points for surveys marked private are visible on the public map.  

Proposed solution: A data script to correct the setting of private data points for surveys that are publicly visible.

PR #1176, opened 2015-03-03
[#1175] App to scan for and fix public private data points

muloem:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

muloem:
@iperdomo please review as well.

scripts/data/src/org/akvo/gae/remoteapi/HidePrivateSurveyData.java
```diff
@@ -0,0 +1,148 @@
...
+                privateSurveyIds.add(survey.getKey().getId());
+            }
+        }
+
+        // retrieve survey instances for private surveys
+        Query surveyResponses = new Query("SurveyInstance").setFilter(
+                new Query.FilterPredicate("surveyId", FilterOperator.IN, privateSurveyIds))
+                .setKeysOnly();
+        List<Long> surveyInstanceIds = new ArrayList<Long>();
+        for (Entity surveyInstance : ds.prepare(surveyResponses).asList(
```
iperdomo: I think we should use `asIterable` when we think there is a large resultset .. 

> When iterating through the results of a query using the PreparedQuery.asIterable() and PreparedQuery.asIterator() methods, the Datastore retrieves the results in batches. By default each batch contains 20 results, but you can change this value using FetchOptions.chunkSize(). You can continue iterating through query results until all are returned or the request times out.

scripts/data/src/org/akvo/gae/remoteapi/HidePrivateSurveyData.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Override
+    public void execute(DatastoreService ds, String[] args) throws Exception {
+
+        if (args.length == 0) {
+            System.err.println("Usage: " + RemoteAPI.class.getName()
+                    + "HidePrivateSurveyData <appid> <username> <password> <appid> [--dry-run]");
+            System.exit(1);
+        }
+        String appId = args[0];
```
jonase: You need to specify the appId twice? If it's just for printing you could also print it in the shell script. In general I think we should supply the appId as a third argument to `execute`.

muloem: yeah I know its silly but thats because of the current setup of the RemoteAPI invocation. Let me see if there is a way to resolve it without messing up the other scripts

Issue #1177, opened 2015-03-04
Optimize cascade question deletion


jonase:
CascadeNodes are deleted on background tasks recursively. New deletion tasks are scheduled for each child node before they are deleted. 

We should not schedule new tasks if the child nodes are at the leaf level since there will be no nodes to delete but the task will still take up a slot in the task queue. Given that the vast majority of nodes in a cascade are leaf nodes and the tasks are executed at a rate of 5 tasks/s this will result in a huge performance boost.

jonase:
## Test Plan

This is an optimization so no user visible changes should be observed

* Create two cascading resources a few levels deep
* Delete one of them, make sure that the other is intact
* Go back to the (undeleted) cascading resource
* Delete a node in the middle of the hierarchy.
* The child nodes should get deleted as well (this will require a refresh, or at least some navigation in the node hierarchy tree to be visible)
* Nodes not part of the deleted subtree (and other cascading resources) should remain intact.

rumca:
tested on release/1.8.3 with several cascades and the above test plan passes :+1:

PR #1178, opened 2015-03-04
[#1177] Optimize cascade node deletion by not scheduling tasks for leaf ...

jonase:
...nodes


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

GAE/src/org/waterforpeople/mapping/app/web/DataProcessorRestServlet.java
```diff
@@ -1443,13 +1443,27 @@ private void deleteCascadeNodes(Long cascadeResourceId, Long parentNodeId) {
...
+            }
         }
 
         dao.delete(nodes);
     }
 
+    private boolean areLeafNodes(CascadeNodeDao dao, Long cascadeResourceId,
+            List<CascadeNode> nodes) {
+        if (nodes.size() == 0) {
+            return true;
```
muloem: Perhaps you can move the `nodes.size() == 0` branch to just before the `if(!areLeafNodes())`.  This way the meaning of the `areLeafNodes()` function is not mixed with checking if the list of nodes is empty.

muloem:
:+1: code reviewed!

Issue #1168, opened 2015-03-02
Cascade question export bug


mtwestra:
BTC reports a problem with the export of data (using question id as headers) for cascade question data.

The issue is that the cascade data is split correctly, but then placed into the wrong columns. See image below:
![screen shot 2015-03-02 at 14 55 19](https://cloud.githubusercontent.com/assets/478954/6441937/32a21b4a-c0ec-11e4-90de-9d82ba2f15d0.png)

instance: http://ctbburundi.akvoflow.org/
surveygroup: Burundi / DBM_BISEM_VF
form: 01_identification

It could be related to a misunderstanding of the number of columns.

jonase:
## Test plan

### Background

When we export a report using the "Use Question Ids" option we split the cascading answers into separate columns (one for each level). The problem is that the number of columns can change (if the form is republished). 

### Test

* Create **three** different cascading resources with different number of levels each (for example 1, 3, and 5 levels)
* Create a form with a single free text question and pick a questionId (**note**: the form should not contain a cascading question at this point)
* Publish the form and create an assignment for your form/phone
* Fill in a few answers.
* Go back to the form and create new a cascading question **before** your free text question. Remember to **pick a questionId**. Selected the resource with the **least** number of levels
* Republish the form, and fill in some answers on your phone.
* Go back to the form and change the cascading question to point to the resource with the **most** number of levels
* Republish the form, and fill in some answers on your phone.
* Go back to the form and change the cascading question to point to the last cascading resource
* Republish the form, and fill in some answers on your phone.

---

* Go to Exports > Export Reports, select your form, tick "use question id" and click "Export"
* Analyze the report. The columns should all line up correctly. All rows should contain the same number of columns.

rumca:
tested on release/1.8.3 and everything works as expected.

@jonase I think we need to present an award for "Best Test Plan" at some point :laughing:

jonase:
@rumca, @muloem :smile:  Here's my nomination: https://github.com/akvo/akvo-flow/issues/818#issuecomment-65885873

rumca:
@jonase @muloem haha I'll start keeping a shortlist and perhaps we can have an award ceremony at the developer week :+1:

PR #1179, opened 2015-03-05
[#1168] Fix cascading question padding issues in raw data reports

jonase:
* Ensure that we pad with empty string when the answer is null
* Ensure that we don't split the answer into more parts than the level count


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

muloem:
:+1: code reviewed!

Issue #1187, opened 2015-03-12
Moving questions from one form to the other 


JosjeS:
Hi All,

A problem occurs when you move a question from one form to another (possible with registration / update surveys). 

When you move the questions from one survey to another the following happens:
1. The Form ID does not change but keeps the ID of the form where it was originally created.

This has the following consequences:
1. The moved questions are not seen as part of the form so are not counted in the 'number of questions' in the forms list.
2. When trying to make dependent questions this is not possible, as when you tick the dependent box, it won't find the option questions because the system does not see it as part as the form but as part of the original form.
3. The moved questions are still counted in the original form as questions. Example you have 12 questions in a form, you move 4 to another form. So now you have 7 questions in the original form. But the counter of the original form will still say 12 questions.

This issue was found by a tender issue: http://flowhelp.akvo.org/discussions/problems/3680-fwd-arghh
And I recreated it in the training instance. 
I created 8 questions in the form 'Josje Test' then I moved 3 questions to the form "Update form".

Josje_Test -> Josje -> Josje Test (survey ID 36866001) ("original" form ID: 37606001) ("moved questions" form ID: 44506003)

mtwestra:
For the moment, we will just fix the UI to reflect the fact that you can't copy across forms.

mtwestra:
Test plan:
1) create a survey with two forms
2) in form 1 create a question group and a question
3) trying to copy or move the question group to the other form should fail
4) trying to copy or move the question to the other form should fail

rumca:
the above test plan passes as described :+1:

PR #1188, opened 2015-03-12
[#1187] disable moving/copying across forms

mtwestra:
* This was never implemented, the fix just fixes the UI
to be consistent with that.

* A proper implementation of copying across forms 
will need to be done later.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1203, opened 2015-03-24
Remove _copy from question ID after duplicating a survey.


m1025:
**problem**
After duplicating a form,  the text in the question ID is accompanied by _**_copy**_

![screen](https://cloud.githubusercontent.com/assets/1729842/6797746/c7e160cc-d216-11e4-8481-b5d625ec2015.png)

**Proposed Solution**
After duplicating a survey, the questions do not come with the text _**_copy**_. Is it possible to use the same logic in the question ID?

muloem:
@mtwestra This depends on the scope of question ID uniqueness. If it is still system-wide as I recall, then this solution cannot work.  The options left are either to not copy the question ID or leave it with the *_copy* suffix.

mtwestra:
@muloem question-IDs are only unique inside a survey (meaning, at most among the forms inside a single survey), so this should not be a problem.

mtwestra:
We will take the simple approach and not add the "_copy". For copying a survey, this will work fine.

The only occasion where this might not work is when we copy a form inside a survey, something we do not support at the moment. However, it is not clear in what case people would want to have two copies of a form in a single survey.

Furthermore, this edge case could be tackled by the foreseen 'survey integrity check' that we want to build in.

mtwestra:
Test plan:
1) create a survey, and include some questionIds
2) copy the survey
3) the question ids in the questions should be identical to the source questions.

janagombitova:
(For 1.9.5 release this is a low priority issue)

janagombitova:
Hi @muloem and @mtwestra what is the status of this fix? There seems to be some referenced code but the issue is not fixed yet. 

Secondly I would like to clarify for future reference: does the question ID have to be unique (1) only in a form but can be repeated in another form in the same survey or does it have to be unique (2) across all forms of the survey? Thank you both

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/copying-survey-question-id-changes
Assignee: Unassigned

janagombitova:
## Problem 

As a dashboard user I want to reuse a survey we have implemented in one region to collect data in another region. We import the collected data into our national database where we link the data set to our structure using question IDs. I want to make a copy of the original survey (with all its forms and questions with question IDs) to create a new survey for the 2nd region. Now if I do so, I have to then go through all the questions and remove the "_copy" from the question IDs. 

### Current rules in place 
[...]

PR #1235, opened 2015-04-10
[#1203] remove '_copy' when copying a questionId

mtwestra:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

muloem:
As @jonase mentioned, this function is also used for copying questions *within* a survey and in these cases the question ID needs to be unique. e.g copying question groups.

mtwestra:
@muloem I see what you mean. So we will need to know when we are copying within a survey, and when not. Depending on that, the behaviour could change?

muloem:
@mtwestra yes. exactly.

jonase:
Here's another scenario that is currently problematic:

1. Create a question with question identifier `foo``
2. Copy the question, the new identifier will be `foo_copy` (so far so good)
3. Copy the same question one more time, this question will get the same identifier as the previous copy.

muloem:
Implementation needs some rework based on comments.

Issue #1130, opened 2015-02-04
Send events to an event log


mtwestra:
We want to get data from Google App Engine into a Unified Log (External to GAE). This Log can then be used by other systems that needs the data for further processing or reporting. 

As an initial prototype we will try to keep a CartoDB Postgres database up to date with the Log as new data continously arrives.

**Initial approach** 

* To get data out of GAE simply hook into the various save methods in [BaseDAO](https://github.com/akvo/akvo-flow/blob/7f4b189413a174d57ae8cc927cfa14cada289b2c/GAE/src/com/gallatinsystems/framework/dao/BaseDAO.java) and post each change to the log.
* Use a table in a postgres db as a "unified log". The table can possibly be as simple as a single column with JSON data.
* Other processes can either read the transaction log (https://github.com/eulerto/wal2json) or query the table for new data periodically.
* We will use json-schema to define the event schema used.

Implementation of https://github.com/akvo/akvo-product-design/issues/67

Continuation of https://github.com/akvo/akvo-flow/issues/766

mtwestra:
When a question group is copied, a nullPointerException is thrown by the event code, at at org.akvo.flow.events.EventLogger.logPut(EventLogger.java:161)

iperdomo:
## Test plan

* Event notification disabled
  * If the `enableChangeEvents` property key is not prent in `appengine-web.xml` system properties, or its value is `false`, the event notifiction is disabled.
  * All the operations in the application must not reflect any new entity in `EventQueue` kind
* Event notification disabled
  * Set `enableChangeEvents` to `true` in `appengine-web.xml` properties (redeploy the application)
  * For all data change events, a new entity is created in `EventQueue` kind, and then a notification is sent to the configured server

rumca:
The above test plan passes fine now. I can see the events being created, there are no errors on the GAE side, just not 100% sure that they're being recorded correctly at the db level - as I'm not sure how to check this.

PR #1249, opened 2015-04-22
[#1130] Use the datastore api directly when storing events

jonase:


GAE/src/org/akvo/flow/events/EventLogger.java
```diff
@@ -129,14 +132,24 @@ private void notifyLog() {
...
+
+            Entity entity = new Entity("EventQueue");
+            entity.setProperty("createdDateTime", timestamp);
+            entity.setProperty("lastUpdateDateTime", timestamp);
+
+            String payload = w.toString();
+            if (payload.length() > Constants.MAX_LENGTH) {
+                entity.setProperty("payloadText", new Text(payload));
+            } else {
+                entity.setProperty("payload", w.toString());
```
iperdomo: we're calling `w.toString()` 2 times, `payload` local variable already has the String value, apologies for not catching it earlier

iperdomo:
:+1: looks good

Issue #1217, opened 2015-03-31
`QuestionGroup` does not resolve all dependent questions correctly when copying `Survey`


muloem:
When copying a `Survey`, copying of `QuestionGroup` entities is executed as an asynchronous task for each `QuestionGroup`.  This results in the failure of the system to correctly resolve a dependency of a `Question` when it depends on another one in a different `QuestionGroup` than the one being copied.

Steps to reproduce:

1. Create a survey and add two question groups

2. Create an option question in group 1 with a number of options.

3. In the first group, create a free text question and make it dependent on one of the options in the question from step 2.

3. In the second group, create a free text question and make it dependent on the option question from Step 2.

4. Copy the survey and check the dependencies for both questions. The dependency for the question in the second group does not resolve correctly.

I have marked this as a regression as it seems this was introduced when copying `QuestionGroup` as tasks was introduced in #435 

Was reported in this issue http://flowhelp.akvo.org/discussions/problems/3700-urgent-copying-surveys

muloem:
### Test Plan

1. Create a survey and add two question groups

2. Create an option question in group 1 with a number of options.

3. In the first group, create a free text question and make it dependent on one of the options in the question from step 2.

4. In the second group, create a free text question and make it dependent on the option question from Step 2.

6. Within the survey copy the second question group. Verify that dependencies are correctly handled

7. Copy the survey and check the dependencies for both questions. Verify that dependencies are correctly handled.

rumca:
Step 5 of the above test plan does not pass successfully. When you copy the second group, the contained questions are copied correctly however their dependencies are not preserved.

rumca:
Test plans pass as expected now :tada:

muloem:
Reopening issue.  The current implementation still has problems and after discussions with @iperdomo, we will implement the copying of surveys as a task and eliminate the tasks for copying survey groups. that way we can preserve or correct dependencies survey wide.

rumca:
Re-using the above test plan - behaviour as expected :+1:

PR #1258, opened 2015-04-25
[#1217] Fix dependent question resolution

jonase:
* If we are copying a single question group and the target option question
  is in some previous group (i.e. not in the same group) we look up the dependant
  question id via the original question (using the sourceQuestionId property).


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

iperdomo:
:+1: looks good

Issue #1282, opened 2015-05-28
Multiple `SurveyedLocale` entities created for the same `surveyedLocaleIdentifier`


muloem:
During submission of survey responses, a new `SurveyedLocale` entity is created for every `surveyedLocaleIdentifier` that is present in the data files.  However, in cases where a device is offline during data gathering and then all data submitted in one big batch to the backend, multiple `SurveyedLocale` entities are created for the same `surveyedLocaleIdentifier`.  As a result we have multiple form responses that should be related to each other through a common `SurveyedLocale` but are not.

Proposed solutions:  For the existing data we will create a data script to correct the wrongly created surveyed locales.  We will also create a fix for the data processing to ensure this does not happen in future releases.

ichinaski:
**Test plan:**
* Download a monitored group in the app
* Enable airplane mode in the device
* Create a bunch of datapoints, each of which containing several form instances (a reg. form plus a few regular forms)
* Disable airplane mode, and let all the data get synced into the dashboard.
* Ensure no duplicate datapoints (same identifier) are found in the dashboard

rumca:
Test plan passes as described :tada:

PR #1284, opened 2015-05-29
[#1282] Correct surveyedLocaleIds for related survey instances

muloem:
* A RemoteAPI script to correct the surveyed locale for responses with the same surveyed locale identifier. Especially relevant for monitoring surveys.
* Added utility functions for batch saving and deleting against remote API

### DIFF
--- a/scripts/data/src/org/akvo/gae/remoteapi/CorrectSurveyedLocale.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/CorrectSurveyedLocale.java
@@ -1,0 +1,195 @@
+/*
+ *  Copyright (C) 2015 Stichting Akvo (Akvo Foundation)
+ *
+ *  This file is part of Akvo FLOW.
+ *
+ *  Akvo FLOW is free software: you can redistribute it and modify it under the terms of
+ *  the GNU Affero General Public License (AGPL) as published by the Free Software Foundation,
+ *  either version 3 of the License or any later version.
+ *
+ *  Akvo FLOW is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *  See the GNU Affero General Public License included below for more details.
+ *
+ *  The full license text can also be seen at <http://www.gnu.org/licenses/agpl.html>.
+ */
+
+package org.akvo.gae.remoteapi;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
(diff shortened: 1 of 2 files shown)


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

scripts/data/src/org/akvo/gae/remoteapi/CorrectSurveyedLocale.java
```diff
@@ -0,0 +1,195 @@
...
+     * @param surveyId
+     * @return
+     */
+    private List<Entity> retrieveSurveyForms(DatastoreService ds, Long surveyId) {
+        Filter surveyFilter = new Query.FilterPredicate("surveyGroupId", FilterOperator.EQUAL,
+                surveyId);
+        Query q = new Query(FORM_KIND).setFilter(surveyFilter);
+
+        List<Entity> surveyForms = new ArrayList<Entity>();
+        for (Entity survey : ds.prepare(q).asIterable()) {
```
iperdomo: Since the list of forms is not huge, we can just do: `ds.prepare(q).asList()` instead of getting an `Iterable` and adding it to a `List` ?

scripts/data/src/org/akvo/gae/remoteapi/CorrectSurveyedLocale.java
```diff
@@ -0,0 +1,195 @@
...
+     * @param formId
+     * @return
+     */
+    private List<Entity> retrieveSurveyInstances(DatastoreService ds, Long formId) {
+        Filter formFilter = new Query.FilterPredicate("surveyId", FilterOperator.EQUAL,
+                formId);
+        Query q = new Query(FORM_INSTANCE_KIND).setFilter(formFilter);
+
+        List<Entity> formInstances = new ArrayList<Entity>();
+        for (Entity formInstance : ds.prepare(q).asIterable()) {
```
iperdomo: Idem

scripts/data/src/org/akvo/gae/remoteapi/CorrectSurveyedLocale.java
muloem: Makes sense. Will do.

scripts/data/src/org/akvo/gae/remoteapi/CorrectSurveyedLocale.java
muloem: In this case there is potentially a larger list i.e. all survey instances.

iperdomo: Correct, my fault

iperdomo:
looks good :+1:

PR #1291, opened 2015-06-04
[#1217] Copy question group tasks

muloem:
* When copying surveys, handle copying of question groups in the same task, i.e. no additional tasks for each question group as was the case before.
* Handle dependency resolution for questions *only* for copying of surveys.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1290, opened 2015-06-03
Accept optional fields in JSON-formatted responses


ichinaski:
# Overview

The current Jackson set up fails when an unexpected (undeclared) field is encountered. For gradual upgrades in our schema, and keeping in mind the miscoordination between some of the FLOW instances (currently only Cardno), we should have a slightly more flexible approach, simply ignoring extra fields found in the `data.json` file.

This means the a non up-to-date FLOW instance, can safely ingest newer versions of the data format, just ignoring the new fields.

ichinaski:
**Test plan:**
* Run tests in `akvo-flow/tests/java/src/org/akvo/flow/test/DataSerializationTest.java`

ichinaski:
See https://github.com/akvo/akvo-flow/pull/1292#issuecomment-109055858

PR #1292, opened 2015-06-04
[#1290] Accept optional fields in JSON responses

ichinaski:
* Skip undeclared fields
* Add test case to existing unit test

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/serialization/SurveyInstanceHandler.java
+++ b/GAE/src/org/waterforpeople/mapping/serialization/SurveyInstanceHandler.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

ichinaski:
After a chat with @muloem, we decided to disallow optional parameters, and instead fail (catch IOException) upon unknown fields parsing. This means the schema is strict and any unspecified attribute will report and log an error. Nonetheless this situation should never be encountered, since all of the instances *should be* up to date. Closing the issue and PR for now.

iperdomo:
> Nonetheless this situation should never be encountered, since all of the instances should be up to date.

In an ideal world this is the case, but we'll face out of sync situations. IMHO, our systems should be robust to accept backward and forward compatibility.

Issue #1307, opened 2015-06-15
Skip QAS referencing an invalid question


ichinaski:
# Overview

While ingesting form responses, we need to ensure all the referenced questions are found in the datastore. If a form is submitted after its has been altered (removing some questions), and subsequently sent to the server, we need to skip invalid responses.

ichinaski:
**Test plan:**
* Create a form and sync it in the app, preferably w/o assignments.
* Delete a question from the form, but do not sync it back onto the app.
* Submit an instance of the synced form.
* The data ingestion should skip the invalid question (checking the logs should suffice, but you can check the datastore to be 100% sure the QAS is not there).

rumca:
The above test plan passes as described :+1:

PR #1308, opened 2015-06-15
[#1307] Ensure questions are found in the DB

ichinaski:
* Attempt to find the question, and proceed if found.

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
+++ b/GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
@@ -69,15 +69,15 @@ import com.google.appengine.api.taskqueue.TaskOptions;
 
 public class SurveyInstanceDAO extends BaseDAO<SurveyInstance> {
     private static final String DEFAULT_ORG_PROP = "defaultOrg";
+    private final QuestionAnswerStoreDao qasDao = new QuestionAnswerStoreDao();
+    private final DeviceDAO deviceDao = new DeviceDAO();
+    private final QuestionDao questionDao = new QuestionDao();
 
     public SurveyInstanceDAO() {
         super(SurveyInstance.class);
     }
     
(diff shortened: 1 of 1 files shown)


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
```diff
@@ -99,21 +99,17 @@ public SurveyInstance save(SurveyInstance si, DeviceFiles deviceFile) {
...
-                                + surveyInstanceId + ", " + si.getSurveyId() + ", " + qas.getQuestionID());
-                continue; // Already stored
-            }
-            
-            if (Question.Type.GEO.toString().equals(qas.getType()) && isNew) {
-                locations.add(qas);
-            } else if ("IMAGE".equals(qas.getType())) {
-                // The device send values as IMAGE and not PHOTO.
-                images.add(qas);
+            if (isValid(qas, si)) {
```
muloem: A bit of a pedantic point here but I find naming the function `isValid` a bit confusing, i.e. what its doing is checking whether or not the response should be skipped rather than whether or not its a valid response.

ichinaski: Sure, I'm open to suggestions. What about `isProcessable()` ? (I unconsciously prepend the *is* prefix for all boolean functions if possible).

muloem: How about `saveResponse()`?

ichinaski: I'd argue `saveResponse(Response response)` kind of suggests side effects (saving the response), rather than merely checking whether the response is valid or not.

muloem:
:+1: code reviewed!

PR #1336, opened 2015-07-15
Issue/872 instance id filter

jonase:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Dashboard/app/js/lib/controllers/data-controllers.js
```diff
@@ -211,8 +211,9 @@ FLOW.surveyInstanceControl = Ember.ArrayController.create({
     this.set('content', FLOW.store.findQuery(FLOW.SurveyInstance, {}));
   },
 
-  doInstanceQuery: function (surveyId, deviceId, since, beginDate, endDate, submitterName, countryCode, level1, level2) {
+  doInstanceQuery: function (surveyInstanceId, surveyId, deviceId, since, beginDate, endDate, submitterName, countryCode, level1, level2) {
     this.set('content', FLOW.store.findQuery(FLOW.SurveyInstance, {
+			'surveyInstanceId': surveyInstanceId,
```
muloem: Would it perhaps make sense to use the direct query instead here? `FLOW.store.find()`?

Dashboard/app/js/lib/models/models.js
```diff
@@ -347,6 +347,7 @@ FLOW.Placemark = FLOW.BaseModel.extend({
 FLOW.SurveyInstance = FLOW.BaseModel.extend({
   approvedFlag: DS.attr('string'),
   approximateLocationFlag: DS.attr('string'),
+  surveyInstanceId: DS.attr('number'),
```
muloem: The `keyId` is the same as `surveyInstanceId` so perhaps we can use that instead?

GAE/src/org/waterforpeople/mapping/app/web/rest/SurveyInstanceRestService.java
```diff
@@ -111,6 +112,19 @@
...
+            List<SurveyInstanceDto> dtoList = new ArrayList<>();
+            if (surveyInstance != null && surveyInstance.getSurveyId().equals(surveyId)) {
+                SurveyInstanceDto dto = new SurveyInstanceDto();
+                DtoMarshaller.copyToDto(surveyInstance, dto);
+                dtoList.add(dto);
+            }
+            response.put("survey_instances", dtoList);
+            return response;
+        }
+
```
muloem: If you move this to the `GET /{id}` remember to filter for authorised surveys

muloem:
Two more general comments:
1. what happens if somebody accidentally types in the string identifier instead? like `skjh-sjhf-xkju`
2. what happens if the instance id is not found?  Maybe we can paste a small text in the results window saying its not found?

jonase:
> what happens if somebody accidentally types in the string identifier instead? like skjh-sjhf-xkju

Currently we return status code "400 Bad Request" and there is no indication in the UI that the user did something "wrong". I'm not sure what kind of feedback we should give to the user. Maybe some kind of instant form validation where we change the color of the input box to red, show a warning icon, or something similar when non-numeric characters are used in the input. Do we have something similar elsewhere on the dashboard?

> what happens if the instance id is not found? Maybe we can paste a small text in the results window saying its not found?

Same thing as for the other filters: No instances are shown in the list. I agree that some kind of message would be helpful, a simple "No results found" would be nice. I'll look into it.

GAE/src/org/waterforpeople/mapping/app/web/rest/SurveyInstanceRestService.java
jonase: I'm not sure if it is better/clearer to go via `GET /{id}` only for the instanceId filter when all other filters are handled in the `rest/survey_instances` endpoint. 

Only accessing authorized surveys is handled by requiring the user to first select a folder/survey/form. I can see how this might seem unnecessary but it's the same for the other filters as well AFAICT.

Another argument for not going via `GET /{id}` is that the return type is different. The `/rest/survey_instances` returns a list and `/rest/survey_instances/{id}` returns a single element. Since we put the result in a list view it makes more sense that the return type is also a list.

janagombitova:
I like the "kind of instant form validation where we change the color of the input box to red" idea and maybe add a sentence under saying "This is an invalid Instance ID." But still allow the user to type in the new ID. 

Inspiration from ING backing 
screen shot 2015-07-17 at 13 57 23

muloem:
As far as I can tell the only place on the dashboard where we do some kind of input validation is the question id.

janagombitova:
@muloem This is from the question ID. Then I think we should do these notifications in a consistent way. Thus reuse the UI from the question ID also here. 

screen shot 2015-07-17 at 14 06 13

muloem:
@jonase any luck with including a short message in case there were no results returned?

muloem:
:+1: code reviewed!

Issue #1335, opened 2015-07-15
Skip QAS referencing an invalid question


ichinaski:
# Overview

Although this problem will not arise on new data (see https://github.com/akvo/akvo-flow/issues/1307), the datapoint sync process does trigger a NPE when fetching an invalid response.

The quick fix here is to skip any response referencing a null/invalid question. Additionally, any survey instance pointing to a null/invalid survey id must be skipped as well.

ichinaski:
**Test plan:**
* Submit data for a monitoring survey
* In the datastore, manually create a SurveyalValue with the same **surveyedLocaleId** and **surveyInstanceId** as the ones that just came in, but setting both the **surveyQuestionId** and **questionType** to `null`. (Simply copy one value, and delete these 2 properties from the new one).
* Attempt to sync the whole survey in the app. No unexpected error should be triggered.

*Note: Although manually updating the datastore is not ideal, this is the quickest way to replicate the context we are fixing here, which is related to data ingested in older versions of FLOW.*

rumca:
the above test plan passes (at least I think I followed it correctly) :+1:

PR #1338, opened 2015-07-16
[#1335] Skip invalid responses

ichinaski:
* Don't process SV w/o a valid questionID
* Simplify type assignment

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/app/web/SurveyedLocaleServlet.java
+++ b/GAE/src/org/waterforpeople/mapping/app/web/SurveyedLocaleServlet.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2010-2014 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2010-2015 Stichting Akvo (Akvo Foundation)
  *
  *  This file is part of Akvo FLOW.
  *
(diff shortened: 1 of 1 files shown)


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

GAE/src/org/waterforpeople/mapping/app/web/SurveyedLocaleServlet.java
```diff
@@ -182,19 +179,16 @@ protected SurveyedLocaleResponse convertToResponse(List<SurveyedLocale> slList,
...
+                                questionTypeMap.put(sv.getSurveyQuestionId(), type);
                             }
                         }
                     }
-                    if (!sv.getQuestionType().equals("IMAGE")) {
-                        // add question type
-                        siDto.addProperty(sv.getSurveyQuestionId(),
-                                sv.getStringValue() != null ? sv.getStringValue() : "",
-                                deviceQuestionType);
-                    }
```
ichinaski: We should probably double check this change with @mtwestra, as I fail to see the reason for having this check in the first place.

muloem: We do sync images as well now right?

ichinaski: We always sync images, yes. The bit that kept me puzzled was that a `SurveyalValue`'s question type, as we saw the other day, is always *PHOTO*, not *IMAGE*, which is exactly the mapping happening a few lines above. So my concerns with that conditional check were:
* If question type is always *PHOTO*, the condition *always* holds true.
* Even if the question type is set to *IMAGE*, why would we cut it out from the responses?

muloem: Perhaps in previous versions of the app images were not synced?

ichinaski: Image *responses* have always been synced. Images themselves were synced in more recent versions, but this feature is completely unrelated.

muloem:
:+1: code reviewed!

Issue #1215, opened 2015-03-31
Replace `path` with Object ID for roles and permissions authorisation


muloem:
The implementation of roles and permissions uses the `path` of a `Survey` or `Folder` as the means to verify the authorisation of a user against that entity.  However, the paths are brittle as they break of an upper level `Folder` or `Survey` is renamed.  The consequence of this change is that users in some cases no longer have access to the `Survey` or `Folder` entities, and in other cases the user cannot even see these entities.  

**Proposed solution:** revise the implementation to use the object ids, which never change, as a means for checking authorisation.

muloem:
### TODO list

- [ ] Revise domain model, `Surveys`, `SurveyGroup` and `SurveyInstances` entities
    - [ ] Create, Read, Update, Delete Entities

- [ ] Create, Update user authorization based on Object Id

- [ ] Filter all `GET` requests internally
    - [ ] /rest/survey_instances
    - [ ] /rest/survey_groups
    - [ ] /rest/surveys

- [ ] Revise RequestUriVoter functions for authorisation check

- [ ] Authorization for `/rest/survey_instances` 
    - [ ] `POST` / `PUT`
    - [ ] `DELETE`
    - [ ]  UI enable/hide delete link (inspect data and monitoring tabs)

- [ ] Authorization for `/rest/survey_groups` 
    - [ ] `POST` / `PUT`
    - [ ] `DELETE`
    - [ ]  UI Show/Hide Icons (Create, Edit, Delete)

- [ ] Authorization for `/rest/surveys` 
    - [ ] `POST` / `PUT`
    - [ ] `DELETE`
    - [ ]  UI Show/Hide Icons (Create, Edit, Delete)

muloem:
@loicsans @Kiarii @mtwestra @janagombitova @iperdomo @jonase 

When we delete a `Survey` or `SurveyGroup` the associated permission will no longer be valid as its associated to the ID.  I'd like opinions on the preferable course of action in this case?  Delete the associated permission, or highlight it with a message that the associated entity has been deleted?

iperdomo:
In traditional RDMBS you can describe the preferred action, `ON DELETE CASCADE`, `ON DELETE RESTRICT` See: http://www.postgresql.org/docs/current/static/ddl-constraints.html#DDL-CONSTRAINTS-FK

[...]

PR #1342, opened 2015-07-16
[#1215] Replace path with objectIds

muloem:
* Backend changes to use the objectId to verify that a user has permissions to perform certain actions
* UI changes to show/hide/disable elements based on user permissions


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

muloem:
@iperdomo @ichinaski if you are able to review it would also come in very handy.

GAE/src/org/waterforpeople/mapping/app/web/CurrentUserServlet.java
```diff
@@ -83,8 +83,12 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
 
         final VelocityContext context = new VelocityContext();
         final UserDao uDao = new UserDao();
-        final String currentUserEmail = UserServiceFactory.getUserService()
-                .getCurrentUser().getEmail().toLowerCase();
+        final com.google.appengine.api.users.User currentGoogleUser = UserServiceFactory
+                .getUserService().getCurrentUser();
+        if (currentGoogleUser == null) {
```
iperdomo: If we're going to `return` there is no point on preparing a template that we're not going to use. I propose to move this check as the first thing to do in the method

scripts/data/src/org/akvo/gae/remoteapi/PathToIdMigration.java
```diff
@@ -0,0 +1,116 @@
...
+            Long entityId = entity.getKey().getId();
+
+            if (entity.getProperty("parentId") == null) {
+                entity.setProperty("parentId", 0L);
+            }
+            idToParentId.put(entityId, (Long) entity.getProperty("parentId"));
+
+            String path = (String) entity.getProperty("path");
+            if (path != null) {
+                pathToId.put(path, entityId);
```
muloem: @jonase Perhaps we can also log here if the path is `null` so that we know which `SurveyGroup` entities are not going to be considered.

jonase: @muloem :+1: done

jonase:
:+1: Code reviewed

Issue #1367, opened 2015-07-30
Moving folder onto itself (estimated: 4)


jonase:
It's currently possible to move a folder onto itself (or some folder further down the hierarchy). This will result in a situation where the parentId path introduces a loop and never reaches the root folder. Immediate result of this is that the folder dissappears from the dashboard.

We should make sure that this is not a possibility:
- Disable the target folder `<a>` tag so it's not possible to navigate into the target folder
- Possibly add a check on the backend

muloem:
@rumca the script for checking for folders has been created but we also need to include a fix to prevent this from happening.  Thats why I moved it to the next release.

rumca:
:+1:

akvotender:
Assigned to [Tender discussion #6247](http://flowhelp.akvo.org/discussions/problems/6247-missing-folder-in-our-flow-instance).

janagombitova:
So I tried to move the folder on itself and tried and tried and could not. But I could move it somewhere else, so all good 👍  Ready to get out there

PR #1368, opened 2015-07-30
[#1367] Check parent path loops datascript

jonase:


scripts/data/src/org/akvo/gae/remoteapi/CheckParentPath.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
+    private static final String MISSING_PARENT = "Missing parent for #%s (chain: %s)\n";
+    private static final String PATH_LOOP = "Path loop for #%s with parentId #%s and chain %s\n";
+
+    private static void checkParentPath(final Map<Long, Long> idToParentId,
+            List<Long> parentPath, final Long id) {
+        Long parentId = idToParentId.get(id);
+        if (parentId == null) {
```
muloem: Will this branch ever execute given that you set the `parentId` on line 27?  Then perhaps the log should be just after line 27?

jonase: Yes it will. This is for the cases where the parent folder does not exist (the id is not found in `idToParentId` since it was never added when building the maping). 

Unfortunately there are such cases where the `parentId` of a survey group is pointing to a non-existing survey group id.

muloem:
:+1: code reviewed!

PR #1374, opened 2015-08-05
[#1346] Move inline styles to main.scss

### LINKED ISSUE #1346: UX Design and CSS styling of data tab search text
Under the `Data > Inspect Data` tab, we currently return a message *No results found*. The styling is directly in the template and needs to move to CSS.

no-results-found-text-message

@Kiarii @gabemart @loicsans

gabemart:
Fixes #1346 by moving inline styles from the Inspect Data tab to main.scss


muloem:
:+1: code reviewed!

PR #1377, opened 2015-08-10
Feature/1280 cartodb maps

jonase:


Dashboard/app/js/lib/views/maps/map-views-common-public.js
```diff
@@ -0,0 +1,205 @@
+FLOW.NavMapsView = FLOW.View.extend({
```
muloem: I noticed this is declared twice and I was not sure whether or not that is accepted by the framework.

Dashboard/app/js/lib/views/maps/map-views-common-public.js
```diff
@@ -0,0 +1,205 @@
...
+
+      // adapt the points shown on the map
+      FLOW.placemarkController.adaptMap(bestBB, this.map.getZoom());
+    },
+
+  /**
+    Once the view is in the DOM create the map
+  */
+  didInsertElement: function () {
+    //console.log(FLOW.Env);
```
muloem: I guess you will probably clear these out.

Config/instanceConfigurator/src/org/akvo/flow/InstanceConfigurator.java
```diff
@@ -327,6 +328,32 @@ private static Options getOptions() {
...
+                "Url endpoint for the cartodb sql api");
+        cartodbSqlApi.setLongOpt("cartodbSqlApi");
+        cartodbSqlApi.setArgs(1);
+        cartodbSqlApi.setRequired(false);
+
+        Option cartodbHost = new Option("ch", "Cartodb host");
+        cartodbSqlApi.setLongOpt("cartodbHost");
+        cartodbSqlApi.setArgs(1);
+        cartodbSqlApi.setRequired(false);
+
```
muloem: Lines `343 - 345` should refer to the `cartodbHost`, right?

GAE/src/org/waterforpeople/mapping/app/web/EnvServlet.java
```diff
@@ -110,14 +113,18 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
...
         }
 
         if (props.get("googleMapsRegionBias") == null) {
             props.put("googleMapsRegionBias", "");
         }
 
+        if (props.get("cartodbHost") == null) {
+            props.put("cartodbHost", "");
+        }
+
```
muloem: Do we also need default `cartodbApiKey` and `cartodbSqlApi` properties here?

GAE/src/org/waterforpeople/mapping/app/web/rest/CartodbRestService.java
```diff
@@ -0,0 +1,305 @@
...
+        InputStream result = connection.getInputStream();
+
+        @SuppressWarnings("unchecked")
+        Map<String, Object> resultMap = objectMapper.readValue(result, Map.class);
+
+        return resultMap;
+    }
+
+    @RequestMapping(method = RequestMethod.POST, value = "update_map")
+    @ResponseBody
```
muloem: Till now we have had the convention of `PUT` for updates.  I was wondering what was your reason to choose the `POST` here

GAE/war/Env.vm
```diff
@@ -12,8 +12,11 @@ loader.register('akvo-flow/flowenv', function(require) {
     mandatoryQuestionID: $env.mandatoryQuestionID,
     showExternalSourcesFeature: $env.showExternalSourcesFeature,
     countries: $env.countries,
-    useGoogleMapsLayers: $env.useGoogleMapsLayers,
-    googleMapsRegionBias: '$env.googleMapsRegionBias'
+    mapsProvider: '$env.mapsProvider',
+    googleMapsRegionBias: '$env.googleMapsRegionBias',
+    cartodbHost: '$env.cartodbHost',
+    hereMapsAppId: '$env.hereMapsAppId',
```
muloem: If you are adding the properties mentioned previously they will probably need to go here as well, I guess.

muloem:
@jonase @kymni As a more general comment I noticed a lot of styling and markup within the code.  I guess this is a consequence of cartoDB implementation. Do you think its too much effort to extract this at some point in the future to external templates?

Dashboard/app/js/lib/views/maps/map-views-common-public.js
jonase: This should be fine for now since the *-public is only used when the public site is requested. When we include the cdb integration for the public maps as well we'll need to unify the implementations.

GAE/src/org/waterforpeople/mapping/app/web/rest/CartodbRestService.java
jonase: I think this endpoint is not used currently. @kymni ?

GAE/war/Env.vm
jonase: Not all of them. For example we don't want to expose the cartodb api key to `FLOW.Env`.

jonase:
@muloem Yes, we will need to move some of the styling to external stylesheets. In general it's been quite liberating to not have to deal with (our version of) Ember so if we move to external templates I'd like to look at options where we perhaps use Handlebar templates but minimize any dependency on the rest of Ember.

I would like to keep the overall architecture such that it only really depends on cartodb tables and not on GAE/Datastore/Ember/"Legacy" backend. This way it will be easier to move the maps to e.g. Dash when we get such a request. I'm not sure if it'll be possible to keep the implementation independent of the Datastore as the next milestone might include features like hierarchical survey selection and integration with roles & permissions.

Config/instanceConfigurator/src/org/akvo/flow/InstanceConfigurator.java
jonase: Good catch, fixed.

GAE/src/org/waterforpeople/mapping/app/web/rest/CartodbRestService.java
kymni: @jonase yes we can get rid of the "update_map" end point for now

muloem:
:+1: code reviewed!

Issue #1332, opened 2015-07-10
Include in-line help in the dashboard in more places 


janagombitova:
Related to https://github.com/akvo/akvo-product-design/issues/88

gabemart:
"Repeat question group" is not yet implemented in develop, so I will not add inline-help item 2 at present.

gabemart:
I have updated transifex with the new text. I will PR once the translated text has been added to transifex.

gabemart:
## Test plan

### Test languages

1) Edit a survey. Under "Survey basics", select "Show advanced settings", then tick "Enable monitoring features". Check that for each language, hovering the cursor over the '?' icon brings up the following help text:

English: "With the registration form you fill in the general information that identifies the entity you want to monitor. You fill in the registration form only once, at the beginning of your monitoring. The registration form creates the data point."

Spanish: Con el formulario de registro Ud. registrará información general que identifica la entidad que desea monitorear. Ud. rellena el formulario de registro sólo una vez, cuando empieza el monitoreo. El formulario de registro crea un nuevo punto de datos.

French: Con el formulario de registro Ud. registrará información general que identifica la entidad que desea monitorear. Ud. rellena el formulario de registro sólo una vez, cuando empieza el monitoreo. El formulario de registro crea un nuevo punto de datos.

### Test Icons Appear

**1)** i) In Edit Survey, go to Forms > Question Group > Add new question > Edit question.

ii) Set the "question type" to "Free text"

iii) Verify a help icon appears after "Use as data point name" and that, on hover, the text reads: 

"When checked, the answer to this question will become part of the ‘name’ of the data point. This will be shown in lists on the device, and can be used to search data points. This should only be enabled for one or two questions which will help to identify the data point, such as an id, or a name."

**2)** Set the question type to "Geolocation". Verify the "use as data point location" tooltip reads:

"When checked, the location captured by this question will be used as the main location of the data point."

**3)** Set the question type to "Date". Verify a help icon appears after "question type" and the text reads:

"The date format is DD-MM-YYYY."

**4)** Set the question type to "Geographic shape". Verify the help icon after "question type" has the text:

"A geographic shape can be a point, a line or a polygon which you mark on the map. This question type enables you to map a set of points, a certain area, or lines, and view that information as a layer on the map. It is useful if you want to map out an area such as a group of water taps, walking paths, a protected forest, a palm oil plantation or an urban area."

[...]

PR #1397, opened 2015-08-20
[#1332] Include in-line help in the dashboard in more places

gabemart:
#1332


Dashboard/app/js/templates/navSurveys/question-view.handlebars
```diff
@@ -33,18 +33,36 @@
...
       {{#if view.attribute}}<label class="labelcheckbox">{{view Ember.Checkbox checkedBinding="view.includeInMap"}}{{t _include_in_map}}</label>{{/if}}
   {{/if}}
-    <label class="selectinLabel"><span>{{t _question_type}}:</span> {{view Ember.Select
-      contentBinding="FLOW.questionTypeControl.content"
-      optionLabelPath="content.label"
-      optionValuePath="content.value"
-      selectionBinding="view.type" }} </label>
+    {{#if view.amDateType}}
+      <label class="selectinLabel"><span>{{t _question_type}}: {{tooltip _question_type_date_tooltip}}</span> {{view Ember.Select
+        contentBinding="FLOW.questionTypeControl.content"
```
muloem: @gabemart given that the tooltip is the main difference here perhaps you could have only a single `{{Ember.Select ...}}` and the `{{#if ... }}` statements would only apply specifically to the tooltip part? that way we avoid duplicating the same code.

gabemart:
@muloem Agreed!

muloem:
:+1: code reviewed!

Issue #1400, opened 2015-08-21
Importing data into a registration form does not create corresponding `SurveyedLocale` entities


muloem:
When data is entered into the RAW DATA spreadsheet without corresponding instance ids or the hash column, and then imported via the *Data Cleaning* tab, new corresponding `SurveyInstance` entities are created.  In the case where data is imported with the registration form of a monitoring survey as a target, we expect that additionally `SurveyedLocale` entities are created.  At the moment this does not happen and a consequence is that it is not possible to synchronise data imported via excel sheets to the devices as this depends on the presence of a `SurveyedLocale`

akvotender:
Assigned to [Tender discussion #5668](http://flowhelp.akvo.org/discussions/problems/5668-fwd-data-upload).

muloem:
### Test Plan

#### Pre test prep
1. Create a regular survey with a number of test questions including a geo question and publish it, as well as a number of the questions selected as "Data point name" questions.

2. Assign to a device and fill out some test responses

3. Generate a raw data report for the survey and ensure it contains all the responses. Save report

#### Test 1: Data Transformation
1. Copy the survey created during test prep and rename it appropriately

2. Edit the advanced settings of the survey to make it a monitoring survey with the main form set as the registration form

3. Export an empty raw data report for that form and copy data from the first raw report into this one, (remember to leave out the columns containing instance ids and digest values)

4. Import this raw report against the newly created monitoring form via data cleaning

5. Check that the data is imported under the inspect data tab

6. Check that there are newly created corresponding data points under the *Data > Monitoring tab* with the data point names set correctly

7. Regenerate the raw data report and check that there are appropriate *identifiers* and correct display names set.

#### Test 2: Data Synchronisation
1. Assign the monitoring survey to the device and synchronise data

2. Check that all new data points under the *Monitoring* tab are synced to the device

muloem:
@iperdomo @jonase @ichinaski @janagombitova I created the third test case (even though its not strictly connected to this issue) after a discussion with Jonas where we figured that importing monitoring data is a bit more nuanced and we have to think it through a bit more.  In the meantime we should prevent users from taking this action and make it clear to them that its not supported at the moment.

rumca:
Test cases pass as described :+1: 

@muloem with Test 3 is there a way of suppressing the original message that is sent in this situation? The spreadsheetProcessed message is still sent along with the warning:

[...]

PR #1412, opened 2015-09-15
[#1400] Import registration forms

muloem:
* Process date correctly when importing date responses
* Process responses according to question type


ichinaski:
Code reviewed!

Issue #269, opened 2013-06-28
Feature Request: Associated Repeatable Questions


bhmann:
If numerical dependencies become a new feature, having those dependencies linked to repeatable questions would be a great way to cut down on the number of questions someone has to build into a survey.

For example, when doing a water point survey, a question (call it [Q1]) is asked "How many water samples were taken?" A photo is needed for each individual water sample, showing the quality of water at the time of collection, or maybe a barcode question is needed to link each sample with the water point (call it [Q2]). Using the numerical dependencies linked with associated repeatable questions, the appropriate number of follow-up [Q2] questions appear based upon your answer to [Q1]. If you responded "3 samples", then 3 [Q2] questions appear to collect the barcodes or photos for each sample.

Through this process, less training is required for enumerators to collect more complex data while survey creators can make sure required information is gathered.

ichinaski:
Relevant documents for the feature:
* [Functional design](https://github.com/akvo/akvo-product-design/blob/master/FLOW/Features/35-RepeatQuestionGroups/FunctionalDesign.md)
* [Technical design](https://github.com/akvo/akvo-product-design/blob/master/FLOW/Features/35-RepeatQuestionGroups/TechnicalDesign.md)
* [Visual design](https://github.com/akvo/akvo-product-design/blob/master/FLOW/Features/35-RepeatQuestionGroups/VisualDesign.md)

ichinaski:
Question not belonging to the same group cannot depend on a repeatable group's questions. Allowed dependencies:
  * [x] Repeatable question on an external group question
  * [x] Repeatable question on a repeatable question **within the same group**
  * [ ] External question (different group) on a repeatable question

When prompting the dependency question dropdown, we need to ensure the aforementioned requirements are fulfilled.

ichinaski:
## Dashboard integration

* Add `repeatable` flag on question groups: This is a fairly trivial action but depends on some aesthetic changes. For the moment being there's a temporary implementation that does the trick to set this flag, though is just squeezed in the existing layout. Perhaps @loicsans has a better understanding of this UI issue.

* Code management: Issue **authenticated** requests to flow-services, and perform CRUD operations for managing codes. This feature belongs initially on the *Devices* tab, though it's UI/UX is pretty much the same as the API key management in the *Users* tab. For bringing about this task's auth part, we need to make use of the JWT claims, specifying which instances the user has access to.

muloem:
#### Test Plan


[...]

PR #1413, opened 2015-09-17
[#269] Repeat save button translation

muloem:
* Fix repeat question group `Save` button interaction
* Fix missing translations


ichinaski:
Code reviewed!

PR #1420, opened 2015-10-01
[#269] Fix monitoring tab template

muloem:
* Fix template error under the monitoring tab
* Render the responses under monitoring tab by question group


jonase:
Looks good :+1:

PR #1423, opened 2015-10-02
[#269] Restore previous monitoring tab view

muloem:
* Retored QuestionAnswerView to be previous state
* Using extended QuestionAnswerView class with specialised template to be used exclusively under the inspect data tab


jonase:
:+1:

PR #1447, opened 2015-10-21
[#269] Raw data Import repeat question groups

jonase:


GAE/src/org/waterforpeople/mapping/app/web/RawDataRestServlet.java
```diff
@@ -143,35 +144,64 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
...
+                            answer = existingAnswers.get(questionId).get(iteration);
+                        }
+                    }
+
+                    // New answer/iteration
+                    if (answer == null) {
+                        answer = new QuestionAnswerStore();
+                        answer.setQuestionID(questionId.toString());
+                        answer.setSurveyId(instance.getKey().getId());
+                        answer.setSurveyId(s.getKey().getId());
```
ichinaski: Why is the survey ID set up twice? Isn't the second call just overriding the first value?

jonase: Good catch! The first call shoud set the `surveyInstanceId`. Fixed.

muloem:
@janagombitova do we already have documentation for the repeat question groups import?

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -136,17 +126,81 @@ public void runImport(File file, String serverBase, Map<String, String> criteria
         Map<Long, Long> columnIndexToQuestionId = processHeader(sheet);
```
muloem: I took a look into the `processHeader()` method and noticed that you add 1 to the column index when you create this map but I couldn't figure out why.

muloem: As you pointed out there will be issues processing the header row of  `GEO` questions.  These will fail because of the split columns.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -136,17 +126,81 @@ public void runImport(File file, String serverBase, Map<String, String> criteria
...
+     * @param questionIdToQuestionDto
      * @return
      */
-    public List<InstanceData> parseSheet(Sheet sheet) throws Exception {
+    public List<InstanceData> parseSheet(Sheet sheet,
+            Map<Long, QuestionDto> questionIdToQuestionDto, Map<Long, Long> columnIndexToQuestionId)
+            throws Exception {
+
+        List<InstanceData> result = new ArrayList<>();
+        int md5Column = Collections.max(columnIndexToQuestionId.keySet()).intValue();
```
muloem: We should also probably take into account the possibility of the last question being a `GEO` question and therefore the `MD5` column will be later.

GAE/src/org/waterforpeople/mapping/dataexport/ExportUtils.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Date;
+import java.util.List;
+
+import org.apache.log4j.Logger;
+import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.Row;
+
+import com.gallatinsystems.common.util.StringUtil;
+
+public class ExportUtils {
```
muloem: This is a very pedantic comment from a readability perspective this is used for import as well but the name suggests otherwise.

jonase: Pedantic comments welcome :+1: 

What would be a better name? `ImportExportUtils`? The general idea is to share as much as possible between Import and Export for things that needs to be kept in sync (md5 calculation, date formatting etc.)

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -154,44 +208,97 @@ public void runImport(File file, String serverBase, Map<String, String> criteria
...
+            return null;
+        }
+        String surveyedLocaleIdentifier = ExportUtils.parseCellAsString(baseRow.getCell(0));
+        String surveyedLocaleDisplayName = ExportUtils.parseCellAsString(baseRow.getCell(2));
+        String deviceIdentifier = ExportUtils.parseCellAsString(baseRow.getCell(3));
+        String surveyInstanceId = ExportUtils.parseCellAsString(baseRow.getCell(4));
+        Date collectionDate = ExportUtils.parseDate(
+                ExportUtils.parseCellAsString(baseRow.getCell(5)));
+        String submitterName = ExportUtils.parseCellAsString(baseRow.getCell(6));
+        String surveyalTime = ExportUtils.parseCellAsString(baseRow.getCell(7));
```
muloem: For this section where we identify the first 8 columns, we need to use a similar construct as in the previous implementation in order to be backwards compatible. See https://github.com/akvo/akvo-flow/blob/develop/GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java#L285-L313.  Or alternatively we need to identify the column in question by using the header text.

GAE/src/org/waterforpeople/mapping/dataexport/ExportUtils.java
muloem: Yep that captures it.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
muloem: I just realised that the first question column index is already accessible here through `columnIndexToQuestionId`

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -154,44 +208,97 @@ public void runImport(File file, String serverBase, Map<String, String> criteria
...
 
         int iterations = 1;
 
         // Count the number maximum number of iterations for this instance
         while (true) {
             Row row = sheet.getRow(startRow + iterations);
-            if (row == null || row.getCell(7).getStringCellValue().equals("1")) {
+            if (row == null || row.getCell(1).getNumericCellValue() == 1.0) {
```
muloem: If we are to keep backward compatibility, then perhaps an extra check could be if its a report that does not contain the repeat column i.e the first question column index is `< 7` or `<8` depending on the column indexing you are going for.

muloem: In case we have an empty row like what appears in the reports with blank rows where the iterations were deleted, then `row == null` means the next survey instance will start processing but on a row that actually contains data for another iteration of repeat group responses, right?

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -154,44 +208,97 @@ public void runImport(File file, String serverBase, Map<String, String> criteria
...
 
-        return null;
+        // question-id -> iteration -> response
+        Map<Long, Map<Long, String>> responseMap = new HashMap<>();
+
+        for (Entry<Long, Long> m : columnIndexToQuestionId.entrySet()) {
+            long columnIndex = m.getKey();
+            long questionId = m.getValue();
+
+            for (int iter = 0; iter < iterations; iter++) {
```
muloem: I may be mistaken but when I was trying to calculate whether this will process all rows in an iteration, seems it skips the last row from what I could calculate. Can you cross check that?

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -154,44 +208,97 @@ public void runImport(File file, String serverBase, Map<String, String> criteria
...
+
+                Row iterationRow = sheet.getRow(startRow + iter);
+
+                long iteration = (long) iterationRow.getCell(1).getNumericCellValue();
```
muloem: Also with this `iteration` variable we should probably check whether the report contains a repeat column or not.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
jonase: Fixed by looking instead at the header row: sheet.getRow(0).getLastCellNum();

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
jonase: Fixed by using the first column (Identifier) instead. Increase the iteration as long as the Identifier = `""`

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
jonase: I checked this and that's not what I'm seeing. The first instance in my test file has 4 repeats and the calculated response map also has four iterations.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
jonase: Yes. I think it should be safe to assume that `iteration == 1` if the iteration column is missing.

jonase: Fixed. I left a `TODO` note that we should remove this special case after a few releases (when we no longer expect to see reports without the Repeat column)

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
muloem: You're right. I miscalculated it.

GAE/src/org/waterforpeople/mapping/dao/QuestionAnswerStoreDao.java
```diff
@@ -215,6 +215,29 @@ public QuestionAnswerStoreDao() {
...
+            // default iteration = 0
+            iteration = iteration == null ? 0 : iteration;
+
+            Map<Integer, QuestionAnswerStore> iterationMap = responseMap.get(questionId);
+
+            if (iterationMap == null) {
+                iterationMap = new HashMap<>();
+            }
+            iterationMap.put(iteration, a);
+            responseMap.put(questionId, iterationMap);
```
muloem: This line doesn't need to be executed for every response, right?

GAE/src/org/waterforpeople/mapping/app/web/RawDataRestServlet.java
```diff
@@ -143,35 +144,64 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
...
+            for (Long questionId : existingAnswers.keySet()) {
+                for (Integer iteration : existingAnswers.get(questionId).keySet()) {
+                    if (incomingResponses.containsKey(questionId)) {
+                        if (!incomingResponses.get(questionId).containsKey(iteration)) {
+                            // Iteration has been deleted
+                            deletedAnswers.add(existingAnswers.get(questionId).get(iteration));
+                        }
+                    } else {
+                        // All iterations has been deleted
+                        deletedAnswers.addAll(existingAnswers.get(questionId).values());
```
muloem: I was thinking we should maybe remove the `else` branch as it may simply be the case that the column with that particular question was deleted from the excel sheet, which doesn't necessarily mean the intention was to delete the data from the backend.  Perhaps we can just ignore if we don't find the question id

GAE/src/org/waterforpeople/mapping/dao/QuestionAnswerStoreDao.java
jonase: correct. It can be moved into the `if` branch. Fixed.

GAE/src/org/waterforpeople/mapping/app/web/RawDataRestServlet.java
jonase: done.

muloem:
:+1: code reviewed!

Issue #1323, opened 2015-07-01
Add codes to cascade questions


ichinaski:
# Overview

Cascade resources **do** store codes already in GAE. However this properties are not reflected in the serialised version of the resource (SQLite database), therefore the app cannot submit the code as the value.

**Actions:**
* All the required data is already present in the datastore, thus is up to FLOW services to add the codes in the SQLite database. In this step no action is needed neither in GAE nor Dashboard.
* The most critical part is to determine the new response serialization format, since response will potentially include both the code and the value (for each cascade level). This decision needs to be aligned with the app implementation.
* Store new values. The current cascade responses, if I'm not mistaken, are fully stored as a single value in the datastore. Depending on the previously mentioned format, some further action will be needed at this stage.
* Data reports. Same as above: the required actions will be dependent on the serialization format we agree on.

**Format options:**

* Although most response types are held in a single *property* (string serialized value), this code is an interesting attribute, since it will be shared by other response types as well (option codes to begin with, but could be extended as product needs evolve). Taking this into account, adding a new property to the **Response** objects could be very useful, as the codes storage will be homogeneous among all response types. However, the cascade responses contain the whole *path* value, thus multiple codes will be needed. This slightly complicates this approach. Option question can contain multiple values (and therefore codes) too, when the `multiple responses` flag is enabled. Should we take this approach further, it'd make sense to use code arrays, instead of single codes, per response.

* Packed code-values pairs. As opposed to adding new properties, we can extend the current paradigm of *packing* values, either with JSON-like structures, or with pipe separated strings. Even though the JSON version is more flexible, it will require more time to visualize data properly, as every single *view* (dashboard, reports, etc) will need to accordingly parse and display a human-readable version of the response.

janagombitova:
Duplicate of #1209

ichinaski:
Sample response representing the first approach (`code` property):

```json
// Cascade response

{
    "questionId": 0001,
    "answerType": "CASCADE",
    "value": "United Kingdom|England|Bristol",
    "code": "acb|def|ghi"
}

// Option question (single response)

{
    "questionId": 0002,
    "answerType": "OPTION",
    "value": "United Kingdom",
    "code": "acb"
}


// Option question (multiple responses)

{
    "questionId": 0003,
    "answerType": "OPTION",
    "value": "United Kingdom|Netherlands|Germany",
    "code": "acb|xyz|jkl"
}
```

ichinaski:
[...]

PR #1448, opened 2015-10-22
[#1323] Render JSON response format cascade codes

muloem:


muloem:
### Test Plan 

#### Pre test prep

Define a cascade resource. Create a monitoring survey with a registration form containing at least a cascade question, a geo question and a simple free text question.  Assign to a device and fill in responses and ensure they are submitted to the dashboard


#### Test 1: Inspect data

1. Check the inspect data tab to make sure the cascade response is correctly rendered as a pipe separated value

2. Edit the cascade response - fill in a different pipe separated string that corresponds to another one existing in the cascade resource.   Make sure this is correctly saved and rendered in the dashboard.

3. Edit the cascade response - fill in a response that doesn't exist in the cascade resource.  Make sure a popup message appears notifying that the response is invalid.

#### Test 2: Monitoring tab 

1. Check the monitoring tab to ensure that the cascade is correctly rendered there


#### Test 3: Map data 

1. Select the data point in the map and ensure that the cascade response is correctly rendered there.

Dashboard/app/js/lib/views/data/question-answer-view.js
```diff
@@ -74,6 +78,46 @@ FLOW.QuestionAnswerView = Ember.View.extend({
...
+      } else {
+        cascadeJson = JSON.parse(c.get('value'));
+        for (var i = 0; i < cascadeJson.length; i++) {
+          cascadeString +=  cascadeJson[i].name + "|";
+        }
+
+        // strip last pipe
+        if (cascadeString.charAt(cascadeString.length - 1) === "|") {
+          cascadeString = cascadeString.substr(0, cascadeString.length - 1);
+        }
```
jonase: I wonder if it would be clearer to use `join` here. Something like
```
cascadeString = cascadeJson.map(function(c) {
  return c.name;
}).join("|");
```
which takes care of not appending a trailing pipe.

Dashboard/app/js/lib/views/data/question-answer-view.js
```diff
@@ -137,7 +181,12 @@ FLOW.QuestionAnswerView = Ember.View.extend({
       }
     }
     FLOW.store.commit();
-    this.set('inEditMode', false);
+    if (FLOW.metaControl.status === 'failed') {
+      this.set('inEditMode', true);
+    } else {
+      this.set('inEditMode', true);
```
jonase: Both the `if` and `else` branch sets the `inEditMode` to `true`. Typo?

Dashboard/app/js/lib/views/views.js
```diff
@@ -93,12 +93,30 @@ Ember.Handlebars.registerHelper('tooltip', function (i18nKey) {
...
   answer = Ember.get(this, 'stringValue').replace(/\|/g, ' | '); // geo data
   answer = answer.replace(/\//g, ' / '); // also split folder paths
 
+  if (Ember.get(this, 'questionType') === 'CASCADE') {
+
+      if (answer.indexOf("|") > -1) {
+        // ignore
+      } else {
+        cascadeJson = JSON.parse(answer);
+        for (var i = 0; i < cascadeJson.length; i++) {
```
jonase: Could we use [Array.join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) here as well?

jonase:
Code reviewed.

PR #1461, opened 2015-11-02
Issue/269 import rqg

jonase:


GAE/src/org/waterforpeople/mapping/app/web/RawDataRestServlet.java
```diff
@@ -180,10 +188,12 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
 
                     answer.setValue(response);
                     answer.setType(type);
+                    answer.setIteration(iteration);
```
muloem: I was just wondering if strictly speaking `iteration` shouldn't only be set when `answer == null`.  In fact thinking about it, perhaps `type` should also not be changed because this creates possibilities for inconsistencies..

muloem:
:+1: code reviewed!

PR #1465, opened 2015-11-04
[#269] Validate that the repeat contains valid values (if it exists).…

jonase:
… Allow empty Instance column for new data


muloem:
:+1: code reviewed!

Issue #1471, opened 2015-11-05
Hide private data script


muloem:
The script to hide private data skips some data points that have the old `pointType` parameter defined.

muloem:
In order to ensure that any changes to the datastore are reflected in the unified log, we are deprecating the data scripts that use the remote API to update datastore entities. Therefore, closing this issue.

muloem:
See https://github.com/akvo/akvo-flow/issues/1568

PR #1472, opened 2015-11-05
[#1471] Private data script

muloem:
* Enable hiding data for `Survey` entities that contain old `pointType` property that is not set to *Household*


muloem:
@jonase this one is not urgent so whenever you get a chance.

jonase:
If possible we should wait for #1492

muloem:
Closing this one. see https://github.com/akvo/akvo-flow/issues/1492 & https://github.com/akvo/akvo-flow/issues/1568

Issue #1415, opened 2015-09-28
Remove column 'form' in Inspect data tab


Geerts:
- After copying a survey the path shown under 'form' is showing the old path
- The information is redundant with the selection you made on the top

akvotender:
Assigned to [Tender discussion #5910](http://flowhelp.akvo.org/discussions/problems/5910-ewb-usa-questions-on-a-couple-of-flow-glitches).

janagombitova:
Secondly, can you change in Monitoring subtab after a user clicks to 'view' collected data, the column header 'Survey' to 'Form'? Thank you 

![screen shot 2015-09-28 at 16 30 55](https://cloud.githubusercontent.com/assets/12456965/10138352/52ad1db0-65fe-11e5-9943-e9975187ed38.png)

janagombitova:
**Related issue:** 

In Monitoring subtab, in the Survey column, the form string does not distinguish between a registration form and the other forms. All the names (strings) are identical.

janagombitova:
@muloem as we will be migrating the Inspect Data tab and the Monitoring tab for the work related to webforms, is this issue still relevant?

muloem:
@janagombitova this is what it looks like 


<img width="1257" alt="no_form_column" src="https://cloud.githubusercontent.com/assets/533336/10970921/c2f399e6-83d1-11e5-98ee-a39a885ee5fa.png">

muloem:
### Test Plan

1. Visit the *Inspect data* tab and preview responses for a particular survey 

2. The *Form* column should **not** be showing

janagombitova:
There is one more issue added here:

**Related issue:**
In Monitoring subtab, in the Survey column, the form string does not distinguish between a registration form and the other forms. All the names (strings) are identical.

@muloem Are we going to leave this one out for now, because of the work related to Webforms (Combining Inspect data and Monitoring)?

muloem:
@janagombitova yes I think indeed we shall leave this one for now and focus on the  refactor

rumca:
Test plan passes as described.

PR #1473, opened 2015-11-05
[#1415] Remove form column

muloem:


jonase:
@janagombitova Do we need to update some documentation screenshots for this change?

janagombitova:
@jonas, yes we do - Once we have the PR ready for Ruarcc I will take care of the documentation changes, as we need to log them down as well for maintaining the translated versions of Read the Docs.

Issue #1477, opened 2015-11-06
Enhance answerCreated and answerUpdated events


jonase:
* With the new Repeatable Question Groups we will need to add `iteration` to `answerCreated` and `answerUpdated` events for the Unilog
* The answers value can be in either the `value` property or the `valueText` property.

rumca:
i'm not sure there's an impact on the dashboard from this, so closing (correct me if I'm wrong though!)

PR #1478, opened 2015-11-06
[#1477] Enhance answerCreated and answerUpdated events

jonase:
* Add 'iteration' if it exists
* Look up the answer value using both 'value' and 'valueText'


muloem:
:+1: code reviewed!

Issue #1125, opened 2015-02-02
No `Previous` button when browsing responses under the Data tab


muloem:
**Steps to reproduce:**
1. Go to the `Data > Inspect Data` tab

2. Select a survey known to have over 20 responses and select `Find`

3. Click on the `Next` button to see next batch of 20 responses.

**Observed:**
There is no `Previous` butting to look at the previous set of records.

muloem:
### Test Plan

1. Create a survey monitoring or not with a number of questions

2. Fill in over 20 responses. (over 40 responses even better)

3. Go to  the inspect data tab and browse the responses and ensure that you are correctly able to browse back and forth between the pages using the `Next` and `Previous` buttons

rumca:
Test plan passes as described. If it's a quick fix it might be nice to sort the styling on the ``next`` and ``previous`` links, otherwise we can address it next release I guess:

![screen shot 2015-11-25 at 17 37 28](https://cloud.githubusercontent.com/assets/3003031/11404781/a1c5f244-939b-11e5-975d-9b6ff23b97bf.png)

rumca:
@muloem I'm not sure if it's the commits related to this or #1299 but the messages tab now throws an error on loading: 

```
Uncaught TypeError: Cannot read property 'indexOf' of undefined
DS.FLOWRESTAdapter.DS.RESTAdapter.extend.setQueryCursor	@	app.js:1
DS.FLOWRESTAdapter.DS.RESTAdapter.extend.sideload	@	app.js:1
d	@	ember-1.0.0.pre-2-36.min.js:1
DS.RESTAdapter.DS.Adapter.extend.didFindQuery	@	ember-data-rev10.min.js:2
DS.FLOWRESTAdapter.DS.RESTAdapter.extend.didFindQuery	@	app.js:1
d	@	ember-1.0.0.pre-2-36.min.js:1
ajax.success	@	ember-data-rev10.min.js:2
k	@	jquery.min.js:2
l.fireWith	@	jquery.min.js:2
y	@	jquery.min.js:2
d
```

rumca:
both of the above comments have been addressed now :+1: 

@gabemart could you possibly add the same alignment tweaks for the ``next`` button on the Messages tab also? It currently appears the same way the inspect data one did.

gabemart:
[...]

PR #1481, opened 2015-11-10
[#1125] Fix previous button on inspect data tab

muloem:
* Refactor use of `since` cursor variable while navigating pages


jonase:
@iperdomo any idea why the @akvobot comment is missing for this issue (and #1481)

PR #1490, opened 2015-11-16
[#1323] Codes for cascades in reports

jonase:


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -616,10 +620,11 @@ private void writeAnswer(
             Row row,
             int startColumn,
             QuestionDto questionDto,
-            final String value,
+            String value,
             boolean useQuestionId) {
 
         assert value != null;
+        value = value.trim();
```
muloem: Just a matter of preference here but perhaps its better to have the `trim()` right at the first point of data entry into the system.. in this case the `BulkDataServiceClient`?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -665,27 +671,73 @@ private void writeAnswer(
...
+                if (value.startsWith("[")) {
+                    try {
+                        parts = OBJECT_MAPPER.readValue(value,
+                                new TypeReference<List<Map<String, String>>>() {
+                                });
+                    } catch (IOException e) {
+                        log.warn("Unable to parse CASCADE response - " + value, e);
+                    }
+                } else {
+                    for (String part : value.split("\\|")) {
```
muloem: Should we maybe use `value.split("\\|", levelCount)` like it was previously?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -665,27 +671,73 @@ private void writeAnswer(
...
                     int padCount = levelCount - parts.size();
 
-                    for (int p = 0; p < padCount; p++) { // padding
-                        parts.add("");
+                    for (Map<String, String> map : parts) {
+                        String code = map.get("code");
+                        String name = map.get("name");
+                        String nodeVal = (code == null ? "" : code + ":") + name;
+
+                        if (cells.size() == levelCount) {
```
muloem: Will this branch of the if statement ever be executed.  Infact thinking about it further even if it is, this is for the export to external system branch so It should probably not contain any `|` characters

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -307,18 +310,50 @@ public InstanceData parseInstance(Sheet sheet, int startRow,
...
+                            break;
+                        case CASCADE:
+                            // Two different possible formats:
+                            // With codes: code1:val1|code2:val2|...
+                            // Without codes: val1|val2|...
+                            String cascadeString = ExportImportUtils.parseCellAsString(cell);
+                            String[] cascadeParts = cascadeString.split("\\|");
+                            List<Map<String, String>> cascadeList = new ArrayList<>();
+                            for (String cascadeNode : cascadeParts) {
+                                String[] codeAndName = cascadeNode.split(":");
```
muloem: just to be absolutely :) sure we should also use `split(":", 2)` here. so we don't get something like `code:`

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
jonase: This is the issue where the number of cascade levels defined is the question < levels in the response. We need to do something here  otherwise we "spill over" into the columns of other questions. Our options are limited:

* don't display more levels than the question specifies => we lose some of the response
* fall back to pipe separated rendering (this is the current approach)

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
jonase: I wanted to build a full in-memory representation of the response similar to the one we would get if the response was JSON encoded. I handle the `levelCount` issue for both response formats below.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
jonase: Good point. That way the value can contain `:` characters as well. I'll add this

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
muloem: Ok. yes that makes sense indeed!

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
muloem: ok. hadn't seen it that way but sure then lets leave as is i.e. fall back to pipe separated until we are able to do better validation

muloem:
:+1: code reviewed!

Issue #1124, opened 2015-02-02
A user with view permissions can still edit forms


joycarpediem:
Steps:
1. Create a role which has permission to only View Folders and Forms.
2. Add a user to the dashboard and assign him this role for a folder/survey

Issue:
1. When the new user logs in , only the assigned Survey/folder is visible but he CAN EDIT  the forms. i.e., the **Edit**, **Copy** and **Move** buttons are shown.

akvotender:
Assigned to [Tender discussion #6133](http://flowhelp.akvo.org/help/discussions/problems/6133-new-permission-setting-view-only-role-needed-requested-by-lars).

muloem:
### Test Plan


#### Pre test prep
Ensure there are at least two users present on the system. One `AdminUser` and `ReadOnlyUser`.  Using the `AdminUser`, create a survey with a few questions.  Create a new role `ReadOnly` and select the permissions `View Folders` and `View Forms` under this role.  Assign the `ReadOnlyUser` to this folder with the `ReadOnly` role.

#### Test 1: View only surveys

1. Login as the `ReadOnlyUser` and browse the survey

2. The Edit survey button should be hidden.

3. Open the survey

4. Survey details should be shown but disabled i.e not possible to edit

5. If there are no forms, then the add form button should appear disabled

6. If there are forms the `Publish` and `Delete` button should be disabled

7. The `Manage Translations` and `Manage Notifications` buttons should be hidden if there are no edit permissions for a form

8. Insert question group button should be disabled

9. the `Edit`, `Move` `Copy` and `Delete` buttons should be hidden

10. `Repeat question` checkbox should be shown disabled

11. `Add new question` should be hidden

12. For each question the text should be shown but the `Edit` `Move` `Delete` and `Copy` buttons should hidden

rumca:
I think these defects are from this issue, but it could be #1460 also not sure.

When a super admin user creates a new form the ``INSERT GROUP HERE`` link no longer appears by default:

![screen shot 2015-11-25 at 09 23 26](https://cloud.githubusercontent.com/assets/3003031/11393258/9bfa827a-9356-11e5-852c-3ae5e68f8edd.png)

it does however appear after a refresh:

![screen shot 2015-11-25 at 09 24 26](https://cloud.githubusercontent.com/assets/3003031/11393267/a87bb064-9356-11e5-8c0e-1b222af91cc9.png)

Also previously when a user created a new survey, the new survey item had a list of option icons as below:

[...]

PR #1505, opened 2015-11-26
[#1124] Ignore `since` meta property for all other request types

muloem:
* We only consider the `since` property for SurveyInstance and SurveyedLocale entities for pagination


jonase:
:+1:

PR #1514, opened 2015-12-01
[#1124] Add new form tab hide

muloem:
* Hide new form tab item for read only
* Hide folder move icon for read only


jonase:
:+1: code reviewed.

Issue #828, opened 2014-10-05
Add labels to option questions options


mtwestra:
The aim is to add a value to question options, in the same way as HTML options have a label and a value. The main purpose for this is to make it easier to import FLOW data into existing databases. The label would be used for display, the value will be used for in API's or optionally in data exports.

In many cases, existing databases use coding schemes to codify answers to questions. For example, 'yes' might be 1, and 'no' might be 0. Another example could be 'banana'=01, 'orange'=02, 'grapefruit'=03.

Implements https://github.com/akvo/akvo-product-design/issues/34

muloem:
@mtwestra did you already push the code done so far for this one?

mtwestra:
@muloem, no, not yet. In need to clean it up and merge it to the current develop. That will take about 0.5-1 day of work

jonase:
## (pre) Test Plan

This issue is not feature complete yet but we have added some initial support for Option codes in exports. There should be no user visible changes when exporting reports with option questions and if there are, it should be regarded as a regression.

rumca:
@jonase is this for the 1.9.4 milestone?

jonase:
@rumca Yes we pushed the report part of the codes for options to 1.9.4.

jonase:
## Test Plan (reporting)

Option questions can be configured to `Allow other?`, `Allow multiple?`, and `Use codes?` (9 combinations in total). 
* For each combination create a separate option question
* Submit a few responses. 
  * For `Allow other?` questions make sure you add an “other” response for some of the submissions (but not all). 
  * Leave some questions unanswered
* Generate a raw data report
  * Multiple responses should be pipe separated: `Foo|Bar|Baz`
  * Responses with codes should be colon separated: `1:ABC|4:JKL`
  * There’s no definite way of knowing if the last response is an “other” or not.
* Clear the md5 hash column in order to import all responses
* In the data cleaning tab, choose the form and upload the excel file
* Check the datastore,
  * Each response should be JSON encoded with a “text” property (and a “code” property for the “Use codes?” questions)
  * The `isOther` flag should be preserved for responses where “other” was used.
  * Any changes you made to the responses should be visible in the datastore.

rumca:
@muloem I think one of the above commits has borked the option codes & saving options in general. It was definitely working before, but now it appears when I add some options to an option question and then click save that they're all just discarded?

rumca:
[...]

PR #1519, opened 2015-12-04
[#828] Codes for options in reports import and export

jonase:


muloem:
@jonase the `SurveyRestRequest` and `xml/Option.java` copyright need updating. ;)

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -623,6 +623,7 @@ private void writeAnswer(
             boolean useQuestionId) {
 
         assert value != null;
+        value = value.trim();
```
muloem: This may not be necessary since we trim the value at `BulkDataServiceClient.parseSurveyInstanceResponse`

jonase: Good catch. This was accidentally reintroduced when on `git merge`. I'll remove it

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -632,155 +633,205 @@ private void writeAnswer(
...
-                    // +------------+------------+-----
-                    // |code1:value1|code2:value2| ...
-                    // +------------+------------+-----
+    private static String dateCellValue(String value, Long questionId) {
+        try {
+            return ExportImportUtils.formatDate(ExportImportUtils.parseDate(value));
 
-                    int levelCount = questionDto.getLevelNames().size();
-                    int padCount = levelCount - cascadeNodes.size();
+        } catch (Exception e) {
```
muloem: Is this exception ever thrown? Seems you already designed all the date parsing functions to already catch and handle the exceptions.  If there are no exceptions then maybe we can also leave out the `questionId` parameter?

jonase: This could happen if the user updates the cell with some random (non-date) value

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -302,7 +313,12 @@ public InstanceData parseInstance(Sheet sheet, int startRow,
 
                 long iteration = 1;
                 if (hasIterationColumn) {
-                    iteration = (long) iterationRow.getCell(1).getNumericCellValue();
+                    Cell cell = iterationRow.getCell(1);
+                    if (cell == null) {
+                        iteration = 1;
```
muloem: `iteration` is already set to `1` by default, so maybe you only need the `cell != null` case here?

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -348,6 +364,65 @@ public InstanceData parseInstance(Sheet sheet, int startRow,
...
+                            // With codes: code1:val1|code2:val2|...
+                            // Without codes: val1|val2|...
+                            String optionString = ExportImportUtils.parseCellAsString(cell);
+                            if (optionString.isEmpty()) {
+                                break;
+                            }
+                            String[] optionParts = optionString.split("\\|");
+                            List<Map<String, Object>> optionList = new ArrayList<>();
+                            for (String optionNode : optionParts) {
+                                String[] codeAndText = optionNode.split(":");
```
muloem: Perhaps here you can add the split limit just to take into account erroneous input like `code:|` or `:text|`?

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -348,6 +364,65 @@ public InstanceData parseInstance(Sheet sheet, int startRow,
...
+                                Map<String, Object> optionMap = new HashMap<>();
+                                if (codeAndText.length == 1) {
+                                    optionMap.put("text", codeAndText[0]);
```
muloem: Maybe we could be overly paranoid and `trim()`?

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -348,6 +364,65 @@ public InstanceData parseInstance(Sheet sheet, int startRow,
...
+                            if (Boolean.TRUE.equals(questionDto.getAllowOtherFlag())
+                                    && !optionList.isEmpty()) {
+                                Map<String, Object> lastNode = optionList
+                                        .get(optionList.size() - 1);
+                                String lastNodeText = (String) lastNode.get("text");
+                                boolean isOther = true;
+                                List<QuestionOptionDto> existingOptions = optionNodes
+                                        .get(questionId);
+                                if (existingOptions != null && lastNodeText != null) {
+                                    // Only compare text. Is that enough?
```
muloem: I believe comparing text is enough. We probably don't want a situation where two options have the same text and different codes since the app user sees the text.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
jonase: Good catch! Will add split limit = 2

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
jonase: :+1:

muloem:
:+1: code reviewed!

Issue #1510, opened 2015-12-01
Copying question groups sometimes fails


jonase:
There have been a number of support requests where question group copying has failed. This can be due to several reasons. We copy question groups in at least the following scenarios:

* Copy question group
* Copy survey
* Copy survey from instance -> instance.

One known issue with "Copy question group" is that you can select to copy a question group, navigate to another survey, and attempt to "Paste group here". This is an "accidental feature" which we don't currently support (@janagombitova should we?). If the user does this the copy will end up in the original form. (https://github.com/akvo/akvo-flow/issues/1404)

Another issue I was able to reproduce is copying the same question groups twice (quickly!). One of the question groups will end up with the status of "COPYING" which it will never get out of.

There might be other issues as well which I haven't been able to find yet.

muloem:
In addition to copying the same question group quickly, copying the same survey quickly also creates the same error.

janagombitova:
@jonase for now, I think we should not support this and we need to make that also visible in the dashboard and later when we redesign the survey management part we can look into it all together

akvotender:
Created via Reamaze:

Link: https://akvoo.reamaze.com/admin/conversations/copying-probleme-occure
Assignee: Jonas Enlund

Message:
Dear colleague,
I wanted to copy data from one folder to copy them into another, but since I am unable to handle anything.
I thought it was probably a problem of internet connection, but after trying several different networks that still persists.
I did a screen capture of your understanding.
the instance name: http://dnhmali.akvoflow.org/admin/
survey ID: 8371913
Thanks for your support,
Best regards.
Banzou

muloem:
### Test Plan

**Test 1: Copying a question group multiple times**

1. Create a test survey containing a form with a question group with a number of questions
2. Copy the question group multiple times within the same form
3. Verify that the copying is successful

**Test 2: Copying a question group to a different form**

1. Create a test survey containing multiple forms. One form should have a question group with a number of questions
2. Select a question group to *copy* and then open a different form.  The `Paste Group Here` link should not be visible. 
3. Go back to the original form, the link should be visible there.
4. Select a question group to *move* and then open a different form.  The `Move Group Here` link should not be visible.
5. Go back to the original form, the link should be visible there.

**Test 3: Copying a survey**

1. Create a survey with a form and a question group containing multiple questions
2. Copy the survey and paste multiple times within  the same and different folders
3. Check the resulting copies that they are correctly executed.

rumca:
[...]

PR #1521, opened 2015-12-08
[#1510] Copying question group issues

muloem:


jonase:
We should add 
``` 
.header("Host",  BackendServiceFactory.getBackendService().getBackendAddress("dataprocessor”)
```
when copying question groups.

jonase:
:+1: Code reviewed

PR #1523, opened 2015-12-09
[#1462] Remove double quotes from french IMEI tooltip

### LINKED ISSUE #1462: Text in in-line help box for French and Spanish versions does not fit
Under the Devices tab we have added an in-line help text next to the IMEI, for the Spanish and French versions the last parts of the text is missing. This is probably due to the size of the box. 

screen shot 2015-11-03 at 15 46 45

gabemart:
Remove double quotes from string to resolve #1462


muloem:
@gabemart did you check for similar occurrences in the files?  Perhaps we can go with your solution to replace everything with `&quot;`

gabemart:
@muloem added this now - I have taken the most conservative approach and only changed the `"` strings to entities for the actual translated strings, rather than changing them in the identifiers too. The identifiers never end up in HTML so changing them seems not strictly necessary

gabemart:
Superseded by https://github.com/akvo/akvo-flow/pull/1537

Issue #1512, opened 2015-12-01
Date type not rendering correctly on overlay for Cartodb maps


rumca:
related to #571 except with Cartodb and not mapbox

Date question responses are currently appearing as below:

![screen shot 2015-12-01 at 09 47 34](https://cloud.githubusercontent.com/assets/3003031/11497405/b7324c94-9810-11e5-8fb2-12e3bc75ba46.png)

cc @jonase

janagombitova:
Display the date in the same way as the Submission date is.

Once we fix the issues around displaying dates in the dashboard in general we will make it all consistent

rumca:
@kymni there's something not quite right with the date display code, it's currently rendering blanks were there should be dates:

![screen shot 2016-01-03 at 23 58 15](https://cloud.githubusercontent.com/assets/3003031/12081630/cb5d35c8-b276-11e5-9d64-070fdee49ea9.png)

kymni:
@rumca issue should be [fixed now](https://github.com/akvo/akvo-flow/pull/1546)

rumca:
Date now appearing as expected :+1:

Issue #1495, opened 2015-11-18
Geoshape - expand the rendering to support 'line' and 'points'


janagombitova:
Currently we only support rendering of polygons in the new geoshape feature. This following step will involve supporting also 'lines' and 'points' when viewing the geoshapes in the Map tab and both sub-tabs of the Data tab (Inspect data and Monitoring)

kymni:
## Test Plan
* Create a "geoshape" question and on the survey collect multi-points, lines, and polygon on a survey then submit
* Inspect the collected data on inspect data tab, monitoring tab, and cartodb map
  * Objects should be drawn on the map
* On cartodb map, click on a point with at least one "geoshape" answer then click on show on map
 * All object types should be projected on the main map

rumca:
Test plan passes as described.

@kymni @janagombitova would it be desirable to support displaying shapes which have multiple features? For example I have submitted some responses which have lines, areas and points as part of their shape. We currently just display one part of that (the points). Please disregard if this is already on the road map or not required :+1:

jonase:
@rumca I think this is actually easy to support in our maps but there were some concerns about exporting data to external systems which only support one kind of feature/column. @joycarpediem might know more details.

kymni:
@rumca @jonase displaying multiple features on a map is indeed possible. Adding support for this

janagombitova:
@rumca Indeed, as @jonase pointed out the entire discussion on if we do support multiple shapes in one answer or not is related to exporting the geoshapes. 

@joycarpediem what is your opinion on this? Do we need to change the entire workflow and not enable anymore multiple shapes in one answer (of the same type of geoshapes or different types)? Or can the GIS people when exporting our shapes to this change themselves?

joycarpediem:
Well there are two aspects to this. One functional and one technical
Functional: The collection of multiple geoshapes against one question is an inherently complex exercise. Also usage wise there seems to be hardly any usecase where organisations may want to do so. 

Technical: In general nested shapes are difficult to work with. Some GIS tools may require additional formatting of data.

IMHO, we can do away with multi shape capture per question.
I think this is a good time to do it as geoshapes are sporadically used right now and we'll not run into too much backward compatibility issues.

rumca:
As per discussion this morning I will open a new issue to address the change in workflow for a future release - making it so that only one feature type/shape can be used per response. We will leave the display functionality as is for now.

janagombitova:
[...]

PR #1524, opened 2015-12-10
Fixes issues [#1495] and [#1512]

kymni:


jonase:
@kymni Do we have a test plan for this?

kymni:
@jonase test plan created [#1495]

Issue #1534, opened 2015-12-18
Data cleaning NUMBER type can result in datastore values in scientific notation 


jonase:
If the number is large enough we currently import it in scientific notation: `1.234E+8`. We need to parse numbers correctly using [org.apache.poi.ss.usermodel.DataFormatter](https://poi.apache.org/apidocs/org/apache/poi/ss/usermodel/DataFormatter.html)

jonase:
## Test Plan

1. Create a form of type `NUMBER`.
2. Submit a few responses with varying numeric size
3. Export a report. Some of the values will be displayed in scientific notation
4. Clear the md5 hash column in order to reimport all responses
5. Import the excel file via the `Data Cleaning` tab.
6. Check the responses in the GAE datastore (Kind: `QuestionAnswerStore`, Sort by `lastUpdateDateTime`) and make sure that none of them are in scientific notation.

rumca:
Test plan passes as described

PR #1536, opened 2015-12-18
[#1534] Use poi DataFormatter to parse cells as String

jonase:


muloem:
:+1: code reviewed!

iperdomo:
@jonase @muloem is `DataFormatter` thread safe? 

> Internally, formats will be implemented using subclasses of Format such as DecimalFormat and SimpleDateFormat.

https://poi.apache.org/apidocs/org/apache/poi/ss/usermodel/DataFormatter.html

> Decimal formats are generally not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.


https://docs.oracle.com/javase/7/docs/api/java/text/DecimalFormat.html#synchronization

PR #1538, opened 2015-12-23
[#828] Support codes for option questions on the dashboard

muloem:


Dashboard/app/js/lib/controllers/survey-controllers.js
```diff
@@ -899,11 +899,260 @@ FLOW.questionControl = Ember.ArrayController.create({
...
+        order: c.get('length') + 1,
+        questionId: this.get('questionId'),
+    }));
+  },
+
+  /*
+   *  Persist all the newly added options to the data store.
+   *  Options with empty code and empty text fields are dropped
+   *  from the list.  If they were already persisted in the datastore
+   *  they are deleted
```
jonase: > If they were already persisted in the datastore they are deleted

What does this mean?

Dashboard/app/js/lib/controllers/survey-controllers.js
```diff
@@ -899,11 +899,260 @@ FLOW.questionControl = Ember.ArrayController.create({
...
+    // reset ordering and persist
+    options.forEach(function (option, index) {
+      var code = option.get('code') && option.get('code').trim();
+      var text = option.get('text') && option.get('text').trim();
+      if (!code) {
+        option.set('code', null); // do not send empty string as code
+      }
+
+      // trimmed whitespace
+      option.set('code', code);
```
jonase: The `if` statement above is undone here?

Dashboard/app/js/lib/controllers/survey-controllers.js
muloem: The idea is if we have any option in the list with blank text and code, we delete it. But if the blank text and code were from an option that had previously been saved to the datastore, we need to remove it from the datastore as well.  Its a bit of a corner case.

Dashboard/app/js/lib/controllers/survey-controllers.js
muloem: good catch! I'll fix that one.

Dashboard/app/js/lib/controllers/survey-controllers.js
jonase: I see, so this is basically how we can delete previously defined options from the datastore :+1:

jonase:
Code reviewed :+1: Well done!

Issue #1453, opened 2015-10-26
Signature question type


janagombitova:
https://github.com/akvo/akvo-product-design/issues/117

muloem:
### Test Plan

#### Test 1: Create and edit a signature question.

1. Create a survey and add a few questions. At least one should be a set as a signature question and another as geo location question.

2. Verify that the question is saved correctly and there are no additional options for the question type

#### Test 2: Preview signature responses on inspect data, monitoring and map (including cartodb) tabs

1. Publish the survey and assign to a device and submit a few responses

2. Verify that the signature is correctly rendered on the inspect data, monitoring and maps tabs

rumca:
@kymni re: the signature questions appearing on the maps part of this, is that something which has already been implemented on the cartodb maps? At the moment all I am seeing is a blank line where the signature should be.

rumca:
@muloem the signature image is being rendered correctly on the inspect data tab, and on the mapbox overlay - however it seems it's not working as intended on the monitoring tab. Instead the below appears:

```{"image":"blah","name":"squiggly "}```

rumca:
Signature questions now render correctly on monitoring tab. As per Skype discussion with @jonase the cartodb issue might take a bit more time to fix, so closing this issue for now (potentially to be address in next release).

janagombitova:
@rumca What is the issue with CartoDB? Can we create a new issue for it then and link it to this one for reference?

jonase:
I suspect that the issue is that we are hitting some cartodb sql-api size limits due to the big payload (due to the base64 encoded signature image)

PR #1540, opened 2015-12-29
[#1453] Signature questions on the dashboard

muloem:
* Add the signature question type
* render signature question responses on the dashboard


jonase:
code reviewed :+1:

PR #1552, opened 2016-01-04
[#1453] Render signature question type on monitoring tab

muloem:
* This is specifically for monitoring forms.  We use the response type value to render since the questions for monitoring forms are *not* yet loaded when browsing the monitoring tab and its therefore not possible to use the question.type property.


jonase:
:+1:

PR #1556, opened 2016-01-05
[#828] Handle JSON option responses in comprehensive reports

jonase:


rumca:
@muloem I had a look at this and think it's OK, so stealing your assignment :smile:

PR #1574, opened 2016-02-02
Feature/1570 strip test

mtwestra:
This code takes care of the following things:
1) correctly show strip-test data in the inspect data tab
2) correctly show strip-test data in the monitoring tab
3) not export img key values (base64 image data) when exporting
4) ignore strip test results when reading in an excel file

I've had some trouble testing point 4 on my machine, so this needs particular attention. 

I also didn't understand how the digest is computed - it seems like this is done on all columns in GraphicalSurveySummaryExporter, while I suppose Signature and Strip test values need to be exempt., so this will need some checking as well.

This is actually a temporary implementation: what is really needed is that Caddisfly gets its own question type. This will allow us to clean up the code, and follow the way it is done for the Signature question type.

### DIFF
--- a/Dashboard/app/js/lib/views/data/question-answer-view.js
+++ b/Dashboard/app/js/lib/views/data/question-answer-view.js
@@ -170,14 +170,16 @@ FLOW.QuestionAnswerView = Ember.View.extend({
 	result=Ember.A();
 	if (c && c.get('value')) {
 	  striptestJson = JSON.parse(c.get('value'));
-	  striptestJson.result.forEach(function(item){
-		  image = 'data:image/png;base64,' + item.img;
-		  newResult = {"name":item.name,
+	  if (striptestJson.result && !Ember.empty(striptestJson.result)){
+		  striptestJson.result.forEach(function(item){
+			  image = 'data:image/png;base64,' + item.img;
+			  newResult = {"name":item.name,
 				  "value":item.value,
 				  "unit":item.unit,
 				  "image":image};
-		  result.push(newResult);
-	  });
+			  result.push(newResult);
+		  });
+	   }
 	}
 	this.set('striptestResult',result);
   },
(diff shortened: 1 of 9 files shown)


Dashboard/app/js/lib/views/data/question-answer-view.js
```diff
@@ -156,6 +162,46 @@ FLOW.QuestionAnswerView = Ember.View.extend({
...
+	   }
+	}
+	this.set('striptestResult',result);
+  },
+
+  /*
+   * Get out the strip test name
+   */
+  striptestName: function(){
+	var c = this.content, signatureJson;
```
muloem: I guess you should rename the signatureJson variable to striptest.

Dashboard/app/js/lib/views/data/question-answer-view.js
```diff
@@ -156,6 +162,46 @@ FLOW.QuestionAnswerView = Ember.View.extend({
     return null;
   }.property('this.content'),
 
+  /*
+   * parse the strip test JSON result
+   */
+  parseStriptestJson: function(){
+	var c = this.content, signatureJson, newResult, image;
+	result=Ember.A();
```
muloem: Same here for the `signatureJson` variable renaming as below

Dashboard/app/js/lib/views/data/question-answer-view.js
```diff
@@ -156,6 +162,46 @@ FLOW.QuestionAnswerView = Ember.View.extend({
...
+	if (c && c.get('value')) {
+	  striptestJson = JSON.parse(c.get('value'));
+	  if (striptestJson.result && !Ember.empty(striptestJson.result)){
+		  striptestJson.result.forEach(function(item){
+			  image = 'data:image/png;base64,' + item.img;
```
muloem: In this line if you have more than one set of results, won't you end up having at the `image` element of the subsequent `newResult` objects containing the same image twice?

mtwestra: I don't think so, as the forEach function iterates over the result array from the JSON, which contains the different images as values. The image is created using item.img.

muloem:
Closing this one.  Was deployed in a side branch as needed. No need to merge to develop.

PR #1581, opened 2016-02-18
Issue/1576 datapoint name

ichinaski:


GAE/src/org/akvo/flow/domain/DataUtils.java
```diff
@@ -37,10 +36,9 @@
         if (optionResponse == null || optionResponse.trim() == "") {
             return new String[0];
         }
-
-        List<Map<String, String>> optionNodes = new ArrayList<>();
-        if (optionResponse.startsWith("[")) {
```
muloem: This original branch structure was to avoid processing a string with the `ObjectMapper` unless it is certain its a JSON string.

GAE/src/org/akvo/flow/domain/DataUtils.java
```diff
@@ -75,24 +99,20 @@ public static String jsonResponsesToPipeSeparated(String optionResponses) {
...
+     * Convert a JSON string response to a list containing corresponding maps
      *
-     * @param optionResponse
-     * @return
+     * @param data String containing the JSON-formatted response
+     * @return List of maps with response properties
      */
-    public static List<Map<String, String>> jsonStringToOptionList(String optionResponse) {
-        List<Map<String, String>> optionNodes = new ArrayList<>();
-        if (optionResponse.startsWith("[")) {
```
muloem: Same here re processing strictly JSON strings.

GAE/src/org/akvo/flow/domain/DataUtils.java
ichinaski: My intention was to rely on the Jackson validator, which throws an `IOException` if the data is not a JSON string, rather than a `startsWith(...)` check. I was unaware of this being an expensive operation. Perhaps I'm missing something?

GAE/src/org/waterforpeople/mapping/app/web/DataProcessorRestServlet.java
```diff
@@ -1504,5 +1510,59 @@ private void scheduleCascadeNodeDeletion(Long cascadeResourceId, Long parentNode
...
+            locales = new ArrayList<>();
+            locales.add(slDao.getById(surveyedLocaleId));
+        } else {
+            // get locales by createdSurveyId
+            locales = slDao.listLocalesByCreationSurvey(surveyId, Constants.ALL_RESULTS, null);
+        }
+
+        List<Question> nameQuestions = qDao.listDisplayNameQuestionsBySurveyId(surveyId);
+        
+        SurveyInstance si;
```
muloem: Maybe we can have this one inside the `for` scope?

ichinaski: The variable is initialized each iteration so yes, it's probably more readable :+1:

GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
```diff
@@ -714,5 +714,27 @@ public void deleteSurveyInstance(SurveyInstance surveyInstance) {
...
+        paramMap = new HashMap<String, Object>();
+
+        appendNonNullParam("surveyedLocaleId", filterString, paramString,
+                "Long", locale.getKey().getId(), paramMap);
+        appendNonNullParam("surveyId", filterString, paramString,
+                "Long", locale.getCreationSurveyId(), paramMap);
+        query.setFilter(filterString.toString());
+        query.declareParameters(paramString.toString());
+        query.setUnique(true);
+        query.setOrdering("collectionDate desc");
```
muloem: Since we are probably retrieving (hopefully) only a single instance, maybe we do not need the unique and ordering filters.

GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionAnswerRestService.java
```diff
@@ -227,12 +228,20 @@
...
                         sval.setStringValue(qa.getValue());
                         slDao.save(sval);
+                        // Populate locale id from the only entity containing this attribute
+                        surveyedLocaleId = sval.getSurveyedLocaleId();
                     }
-
+                    
+                    // Update datapoint names for this survey, if applies
+                    if (q.getLocaleNameFlag() && surveyedLocaleId != null) {
+                        DataProcessorRestServlet.scheduleDatapointNameAssembly(q.getSurveyId(), surveyedLocaleId);
```
muloem: When updating the display name for single responses, do we also take into account to only do this for registration forms?

ichinaski: In order to avoid an extra query for checking the creation survey in the SurveyedLocale entity, I use the `getLocaleNameFag()` check as a preliminary check - this should only be set in registration surveys. If for whatever reason the question is mistakenly flagged, the task will find out this is not the registration form, and therefore just do nothing (name won't be updated).

GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
ichinaski: I expect these two filters to make the query more efficient in case we have more than one registration form:

- `setUnique(true)` will make sure only the first result is returned, rather than a Collection, and `null` if it doesn't exist. This way we don't have to fetch a List to check it's first element. Just read the query response in a SurveyInstance variable.

- The ordering one is just to ensure we get the latest registration form, avoiding any subtle surprise we can have if multiple registration forms are present.

muloem:
:+1: code reviewed!

PR #1588, opened 2016-03-01
Issue/1576 datapoint name

ichinaski:


GAE/src/org/waterforpeople/mapping/app/web/DataProcessorRestServlet.java
```diff
@@ -1520,18 +1524,32 @@ private void assembleDatapointName(Long surveyId, Long surveyedLocaleId) {
...
+            // Fetch all locales for this survey group
+            locales = slDao.listLocalesBySurveyGroupId(surveyGroupId);
         } else {
-            // get locales by createdSurveyId
-            locales = slDao.listLocalesByCreationSurvey(surveyId, Constants.ALL_RESULTS, null);
+            return;
+        }
+        
+        SurveyGroup sg = sgDao.getByKey(surveyGroupId);
+        if (sg == null) {
```
muloem: I was thinking perhaps to make this cleaner you can already check for `surveyedlocaleid == null` , `surveyGroupId == null` and `sg == null` at the start of the method. Preconditions.

ichinaski: We can check that either `surveyedLocaleId` or `surveyGroupId` is not null, yes. Just note that we don't require both to be defined. Also `sg` is not known until later on, once the `SurveyGroup` object is fetched.

muloem: also if `sg.newLocaleSurveyId == null`

muloem:
:+1: code reviewed!

Issue #1553, opened 2016-01-04
Ignore blank last cell in raw data reports


muloem:
In the raw data reports used for data cleaning, excel seems to add a blank cell in the header row of the last (*hash*) column of the report.  This results in the error `The header "" can not be imported` during raw data import. We can simply ignore this header if we know that its the hash column. So we need to modify the report import to correctly identify the hash column

akvotender:
Linked to the Issue of Jennie (SNV)

muloem:
### Test Plan

1. Export a raw data file with test data from a survey

2. In the header cell of the last column type in a space, then save the document and reimport it

3. The import should not fail i.e the *spreadsheet processed* message should appear

rumca:
@muloem does this mean the header column that the hash is in, or the column to the right of that? Currently getting a ``Invalid RAW DATA file: blah.xlsx - Errors: The header " " can not be imported`` when I add blank spaces to the hash column and also the one to the right.

muloem:
@rumca my mistake. I hadn't yet deployed the latest version to FLOW services. please recheck. thanks!

rumca:
yep works as intended now :+1:

PR #1595, opened 2016-03-07
[#1553] Ignore blank cells at the end of header row when importing reports

muloem:


GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -758,6 +767,26 @@ protected void invokeUrl(String serverBase, String urlString,
...
+    /**
+     * When a blank or null cell is incurred while processing, make sure that this is the last cell
+     * in the row and ignore any other "phantom cells" that may occur. We only allow this for the
+     * header row. If the blank cell occurs in between valid header cells we return true.
+     *
+     * @param cell
+     * @return
+     */
+    private boolean isMissingHeaderCell(Cell cell) {
+        assert cell.getRow().getRowNum() == 0; // only process header rows
```
ichinaski: I tend to be wary of assertions in non-tests, production code. Why would we want an AssertionError being thrown here? As far as I can tell, we are not handling that particular error. Also, are assertions enabled in GAE?

ichinaski:
Just raised a small concern re: assertions (s/GAE/FLOW-services if this code is meant to be run elsewhere). Otherwise all looks good.

Edit: copyright header

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
muloem: The code will run under FLOW services so no worries about GAE support.  In this case the main motivation is to ensure the method is only used for checking cells in the header row.

ichinaski:
Code reviewed!

PR #1606, opened 2016-03-22
Feature/1527 geotagging

ichinaski:


Dashboard/app/js/lib/views/data/question-answer-view.js
```diff
@@ -349,7 +349,8 @@ FLOW.QuestionAnswerView = Ember.View.extend({
   photoUrl: function(){
     var c = this.content;
     if (!Ember.empty(c.get('value'))) {
-      return FLOW.Env.photo_url_root + c.get('value').split('/').pop();
+      jImage = JSON.parse(c.get('value'));
+      return FLOW.Env.photo_url_root + jImage.filename.split('/').pop();
```
muloem: it should not happen ideally but perhaps we can add a check here to make sure filename is not `undefined` or `null` etc

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.logging.Logger;
+
+import org.codehaus.jackson.map.ObjectMapper;
+import org.waterforpeople.mapping.domain.response.value.Media;
+
+public class MediaResponse {
+    private static final Logger log = Logger.getLogger(MediaResponse.class.getName());
+    private static final ObjectMapper JSON_OBJECT_MAPPER = new ObjectMapper();
+    
+    public static final int VERSION_INITIAL    = 0;
```
muloem: Maybe these can be more descriptive towards what the formatting for the version is e.g. `VERSION_STRING`

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
```diff
@@ -0,0 +1,65 @@
...
+     * Format media value with the given format. If the data is 
+     * already in the output format, no transformations are applied.
+     */
+    public static String format(String value, int version) {
+        Media media;
+        int savedVersion;
+        try {
+            media = JSON_OBJECT_MAPPER.readValue(value, Media.class);
+            savedVersion = VERSION_GEOTAGGING;
+        } catch(IOException e) {
```
muloem: Maybe its an idea to distinguish more precisely between whats an exception because `value` is a filename string and when its possibly a malformed JSON string?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -66,6 +66,7 @@
 import org.waterforpeople.mapping.app.gwt.client.surveyinstance.SurveyInstanceDto;
 import org.waterforpeople.mapping.app.web.dto.SurveyRestRequest;
 import org.waterforpeople.mapping.dataexport.service.BulkDataServiceClient;
+import org.waterforpeople.mapping.serialization.response.MediaResponse;
```
muloem: Can you cross check that this class will be added to the jar when building the exporter.jar file?

muloem:
We have to think of how to handle situations where image location data is stored as JSON and we import excel responses.

Dashboard/app/js/lib/views/data/question-answer-view.js
ichinaski: ok. Would a `if (jImage && jImage.filename)` suffice or have you got something else in mind?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
ichinaski: Is there any reason for which it wouldn't be added to the jar?
What's wrong with having this dependency?

I may be missing something of this build process...

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
ichinaski: ok

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
ichinaski: The rationale behind the original implementation is to assume a `value` if the JSON parsing fails. I am currently not contemplating a malformed JSON (not sure how that could happen). How is that handled in other response types?

ichinaski:
> We have to think of how to handle situations where image location data is stored as JSON and we import excel responses

This is currently solved by generating a JSON-formatted value when parsing the data. Obviously the location is left empty.

GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionAnswerRestService.java
```diff
@@ -152,13 +154,24 @@ private void processApiV1Responses(List<QuestionAnswerStoreDto> responses,
         for (QuestionAnswerStoreDto qDto : responses) {
             String optionResponseValue = qDto.getValue();
```
muloem: I had missed out this one.  `optionResponseValue` is confusing now since its processing image responses as well.

muloem:
:+1: code reviewed!

PR #1607, opened 2016-03-23
Issue/257 date responses

ichinaski:


muloem:
the copyright headers.. :-|

Dashboard/app/js/lib/views/surveys/survey-group-views.js
```diff
@@ -197,7 +197,7 @@ FLOW.ProjectItemView = FLOW.View.extend({
   formatDate: function(datetime) {
     if (datetime === "") return "";
     var date = new Date(parseInt(datetime, 10));
-    return date.getDate() + "." + (date.getMonth() + 1) + " " + date.getFullYear();
+    return date.getDate() + "-" + (date.getMonth() + 1) + "-" + date.getFullYear();
```
muloem: Shows ok except we need to make adjustment to `yyy-MM-dd`

Dashboard/app/js/lib/views/views.js
```diff
@@ -555,11 +555,10 @@ FLOW.DateField2 = Ember.TextField.extend({
     this._super();
 
     this.$().datepicker({
-      dateFormat: 'yy/mm/dd',
+      dateFormat: 'yy-mm-dd',
```
muloem: For some reason the date picked popup doesn't work under the inspect data tab when editing date responses

GAE/src/org/waterforpeople/mapping/dataexport/ExportImportUtils.java
```diff
@@ -86,27 +93,62 @@ public static String md5Digest(List<Row> rows, int lastColumnIndex)
...
+
+    /*
+     * Format DateQuestion response in a human-readable way.
+     * This date will not contain time details (yyyy-MM-dd)
+     */
+    public static String formatDateResponse(String value) {
+        try {
+            Date date = new Date(Long.valueOf(value));
+            return DATE_RESPONSE_FORMAT.get().format(date);
+        } catch (NumberFormatException e) {
```
muloem: In some cases we had incidents in the past where the datastore contained date strings instead of timestamps e.g `21-03-2016 16:21:05`  So thats why the previous formatting would fall back to parsing for this format if a `NumberFormatException` occurred.  And we should probably keep it that way for the time being.

ichinaski: The previous date format, as far as I understand, used the String-based formatting as it handled both timestamps and imported data values (the format you point out). The function was explicitly called from both sides if I remember correctly https://github.com/akvo/akvo-flow/commit/750e261bb4a5d0d2eaef8ce81c8405c55d95f171

Dashboard/app/js/lib/views/views.js
ichinaski: You are right. I switched back to master and the problem persists. I'm not sure where the root issue resides

GAE/src/org/waterforpeople/mapping/dataexport/ExportImportUtils.java
muloem: Yes. exactly. and this was by design to handle the possibility of having date strings coming from the datastore and going out to the exported report as you can see in the comment of that commit.

ichinaski: I see. I thought it was merely to handle export/import in the same function. Will add this fallback...

muloem:
:+1: code reviewed!

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -428,7 +428,7 @@ public InstanceData parseInstance(Sheet sheet, int startRow,
 
                         case DATE:
                             String dateString = ExportImportUtils.parseCellAsString(cell);
-                            Date date = ExportImportUtils.parseDate(dateString);
+                            Date date = ExportImportUtils.parseSpreadsheetDate(dateString);
                             if (date != null) {
                                 val = String.valueOf(date.getTime());
```
muloem: Just realised that incase it happens that the format here is the old one i.e. `dd-MM-yyyy HH:mm:ss Z`, we should set the timestamp to 00h00 that day.

ichinaski: I'm not sure that would fix the issue. Note that TZ would still be present, so the `Date` parsing is taking that into account. Resetting the hours/minutes/seconds does not guarantee the output date will be the same as the one introduced in the spreadsheet

Issue #1611, opened 2016-04-05
NPE on non-monitored survey publishing


ichinaski:
# Overview

`newLocaleSurveyId` is not guaranteed to be non-null. In non-monitored surveys we can safely assume the form being published is the registration one, for the aforementioned attribute does not exist.

rumca:
non monitored surveys no long throw the npe on publish :+1:

PR #1612, opened 2016-04-05
[#1611] Prevent NPE on non-monitored surveys

ichinaski:
* Assume registration form on non-monitored surveys
* Check newLocaleSurveyId on monitored ones


muloem:
:+1: code reviewed!

Issue #1602, opened 2016-03-21
Load dashboard languages on demand


muloem:
At the moment we are loading strings for all available dashboard languages even when these will not be used.  This adds a considerable fraction to the dashboard loading time.  We need to refactor to load only the language strings that the current user will need and only load another language on-demand.

muloem:
#### Test Plan

1. Load the public page of the dashboard.  The default language appearing should be English.  Any changes to the language should results in only English being returned

2. Log in to the dashboard and switch between the different languages and verify that the correct language is shown after the page is reloaded.

rumca:
The first part of this test plan doesn't pass - changes to language on the public page result in the language actually changing (which is the behaviour I would expect).

Second part passes.

muloem:
@rumca true. ignore the first test please. it should only happen that way if the public page is visited when *not* logged in.

PR #1614, opened 2016-04-06
[#1602] Users tab loading

muloem:
* Do not use local storage for language configuration when loading user tab


ichinaski:
Code reviewed

Issue #1644, opened 2016-05-05
Device identification (duplicates)


ichinaski:
# Overview

We are aware of the device identification system being weak, but the issue seems to be rampant as of late. The main problem is twofold:

* Lack of IMEI (i.e. tablets) causes the device to be registered multiple times, namely each time the survey assignment servlet is hit.
* Concurrent connections to various servlets (surveys, device notification) may register the device twice, since none of them is able to fetch it from the datastore upon receiving the request.

After a chat with @muloem, the quickest way to resolve this is:

* Ensure all servlets reuse the same registration policy, gracefully handling the lack of IMEI.
* Stop registering the device in the device notification servlet. This connection mostly happens at the same time as the survey assignment one. Only the latter should be able to register new entities if the provided device is not found in the datastore.

*Note: This is still a temporary measure, and device recognition should be more robust to avoid this situations further down the road*

ichinaski:
**Test plan:**

* Issue multiple requests (sequentially) to the survey assignments servlet without containing the IMEI value. For example: `https://akvoflow-dev2.appspot.com/surveymanager?action=getAvailableSurveysDevice&phoneNumber=15555215554&androidId=dc2ea33bc8dd5c33&imei=&ver=2.2.3&devId=n4`

* Connect a new device to the server. This will trigger a couple concurrent connections to different servlets.

* In both scenarios, the given device should only get registered **once**.

rumca:
test plan passes as described (i think)

PR #1646, opened 2016-05-05
[#1644] Avoid multiple registration for devices

ichinaski:
* Small refactor on device data update
* Only update, not register devices on DeviceNotificationRestServlet


muloem:
:+1: code reviewed!

Issue #1527, opened 2015-12-15
Geotagging in photos 


janagombitova:
All details on the implementation of the geotag and decisions can be found here: https://github.com/akvo/akvo-product-design/issues/74

For the first implementation we will show the geotag in the Data tab.

Providing this information in reports and visualising the location of the photos in the maps, will be dealt with in the following implementations. https://github.com/akvo/akvo-product-design/issues/74

ichinaski:
**Test plan:**

Test plan:

* Create a monitoing form containing media questions (image, video)

* Provide a response from both old and new versions of the app (i.e. v2.2.3.1 and v2.2.4)

* Ensure image/video URLs are visualized correctly in the inspect data tab. They should not be editable.

* Monitoring tab should work the same way.

* Check data exports. Format should be the same as usual, only containing the URL for now.

rumca:
The discussed URL issues in reports seem to be fixed (for all but one response - which could just be down to something funky going on before the fix).

I'm going to redo the reports testing just to double check and then will close this issue.

rumca:
The fixes for video URLs on monitoring & inspect data tab look good. As do the fixes for loading images on the maps tab, and displaying the URLs correctly instead of JSON strings.

rumca:
temporarily closing this one (until we tag v1.9.6)

janagombitova:
Based on @ichinaski discussion with Loic - https://github.com/akvo/akvo-product-design/issues/74#issuecomment-209368877 we will add the location information under the media link in the dashboard. 

This information needs to be shown in the: 
- Inspect data tab
- Monitoring tab
- Maps (current and CartoDB) 

**The outstanding questions are:**
- How to add the location information in raw data reports? (should a new issue be created to handle reports?) 
   - Inigo mentions in the above linked comment, to add the location into the same cell as the media link. 
   - after speaking with some of our users, the preferred option is to add one extra column following the media column that will hold the location information, to not change (and complicate) the way photos are access via the reports  
   - What shall we implement? 
   - Will we allow editing via data cleaning? 
- Will we allow editing in the Inspect data tab? 

@ichinaski @muloem what do you think? 

Following implementation steps will be to (1) calculate and show the distance between the media and the data point location and (2) add pins on the CartoDB maps once previewing a specific data point. These steps are not n1 priority and thus will be tackled later on in time. https://github.com/akvo/akvo-product-design/issues/141

ichinaski:
[...]

PR #1670, opened 2016-05-12
[#1527] adding check for IMAGE type as well as PHOTO

rumca:
it seems that sometimes we have photo question answers which don't have PHOTO as their type - instead IMAGE. When this happens the logic for displaying lat/lon and parsing the JSON responses breaks down.


muloem:
:+1: code reviewed!

Issue #1655, opened 2016-05-11
Change forms endpoint on CartoDB maps from cartodb storage to dashboard storage


kymni:
The current end point pulls surveys (forms) from the form table in cartodb. At the moment if a survey is deleted from the dashboard, the event is not executed on cartodb so users will see surveys that they already deleted on the form selector.
Quick fix is to change the end point to use the dashboard's surveys endpoint and pass the surveyGroupId. 
@muloem any additional thoughts on this?

muloem:
Deleted surveys no longer appear in the list under the maps tab. 👍

PR #1679, opened 2016-05-24
#1655 Change surveys end point

kymni:


kymni:
## Test plan
When filtering by folders and surveys on cartodb maps, the endpoint that pulls forms that belong to a survey should point to Flow dashboard instead of cartodb

PR #1681, opened 2016-05-27
Issue/998 android

ichinaski:


GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
```diff
@@ -139,13 +139,8 @@ public SurveyInstance save(SurveyInstance si, DeviceFiles deviceFile) {
...
                     qas.getValue().lastIndexOf("/") + 1);
 
-            Device d = null;
-            if (deviceFile.getImei() != null) {
-                d = deviceDao.getByImei(deviceFile.getImei());
-            }
-            if (d == null && deviceFile.getPhoneNumber() != null) {
-                d = deviceDao.get(deviceFile.getPhoneNumber());
-            }
+            Device d = deviceDao.getDevice(deviceFile.getAndroidId(),
```
muloem: I noticed that this is within a loop. Perhaps you can shift it out and instead use a hash to store the device (if thats necessary at all).

GAE/src/org/waterforpeople/mapping/app/web/TaskServlet.java
```diff
@@ -164,6 +165,7 @@ public TaskServlet() {
         deviceFile.setProcessDate(getNowDateTimeFormatted());
         deviceFile.setProcessedStatus(StatusCode.IN_PROGRESS);
         deviceFile.setURI(url);
+        deviceFile.setPhoneNumber(androidId);
```
muloem: `setAndroidId()`

muloem:
Also perhaps we need to add the AndroidId to this set of params for the rescheduling.. https://github.com/akvo/akvo-flow/blob/ec09ea90ada7e21cc39181f8dbafb3e57c479d2c/GAE/src/org/waterforpeople/mapping/app/web/TaskServlet.java#L278-L303

GAE/src/org/waterforpeople/mapping/app/web/TaskServlet.java
ichinaski: ouch!

muloem:
So regarding the `/surveyedlocale` endpoint change, it was mainly because the whole block will possibly (even though highly unlikely) be skipped because of this.. 
https://github.com/akvo/akvo-flow/blob/0ec6905ff029b155c535f303b97906c6833ccca9/GAE/src/org/waterforpeople/mapping/app/web/SurveyedLocaleServlet.java#L83

ichinaski:
The `DeviceSurveyJobQueueDAO` [catches breath again], already checks the androidId, IMEI, and phone number params. Perhaps we can fully remove this condition here?

muloem:
Yep. sounds like a plan. clean away!

muloem:
:+1: code reviewed!

Issue #987, opened 2014-12-12
Improve performance of /survey_groups loading (estimated: 16)


iperdomo:
Add a `parentId` parameter to load only a portion of the `SurveyGroups`

muloem:
#### Test Plan
* Verify the metrics working 
* Verify cascade resources working as expected
*

iperdomo:
@janagombitova @muloem why not adding the _(estimate: 16)_ as a tag instead of changing the issue title? just asking :wink:

muloem:
We considered having them as tags but didn't agree yet whether we actually want this info here or use waffle. so @janagombitova left like this for now. 👍

ichinaski:
Cascade resources working as expected. 
@muloem What exactly is the metrics test targeting?

muloem:
@ichinaski Metrics test should be with the showStatistics flag set. And you should be seeing the stats from data collection.

ichinaski:
On the metrics test, there's a small glitch when browsing _Statistics_:
The first dropdown displays both **folders** and **surveys**. Selecting a survey works fine (second dropdown shows forms within it). However, selecting a folder results in an empty form dropdown.

PR #1682, opened 2016-05-30
[#987] Improve performance of loading dashboard

muloem:
* Get rid of BeanUtils for copying properties
* Rearranged order of requests to dashboard. Cascade resources and metrics end points loaded after surveys
* Reduced `surveyList` to array of IDs instead of array of Survey entities


ichinaski:
Code reviewed, looks good.

PR #1699, opened 2016-06-07
[#987] Load metrics when loading surveys

muloem:
* This fixes a regression that causes question properties not to be loaded.


ichinaski:
Code reviewed

Issue #1402, opened 2015-08-24
Editing survey twice in same session results in blank page


gabemart:
To reproduce:

1) Create a new survey

2) Add a form

3) Hit "Surveys" link to enter Surveys tab

4) Edit the form you just created

Expected result: navigate to survey editor

Actual result: Blank screen

Console reads: `Uncaught TypeError: Cannot read property 'get' of null`

caused by: `return form.get('status') === 'PUBLISHED'` in 
survey-group-views.js - [view it here](https://github.com/akvo/akvo-flow/blob/develop/Dashboard/app/js/lib/views/surveys/survey-group-views.js#L83-L86)

rumca:
Nice! I've encountered this a few times, but could never workout the exact steps to reproduce it consistently. I think there may be a few other places that a similar issue occurs, I've seen the initial surveys tab blank screen a few times too.

muloem:
# Test Plan

1) Create a new survey

2) Add a form

3) Hit "Surveys" link to enter Surveys tab

4) Edit the form you just created

You should be able to go to the form without hitting a blank screen

ichinaski:
Tested on dev-2. Works as expected.

janagombitova:
This is still a problem and it would be good to give it a second look. It is quite an annoying bug

janagombitova:
@kymni can you prioritise this issue forward on your to do list once back from holidays?

kymni:
To reproduce:
1. Leave the survey tab and select a survey using the survey selectors on any of the other parts of the dashboard
1. Navigate to the survey tab and select the same survey selected above
1. You should see a blank screen

Navigating away to a different part of the dashboard then back to the survey tab and selecting a different survey doesn't produce a blank screen

janagombitova:
@kymni have you tried with selecting the **same** survey?

kymni:
@janagombitova yes, basically selecting the same survey reproduces the the problem but selecting a different survey doesn't

PR #1702, opened 2016-06-08
[#1402] Roll back blank page fix.

muloem:
* This has to be fixed as refactor of overall state management strategy for dashboard.

### DIFF
--- a/Dashboard/app/js/lib/views/surveys/survey-group-views.js
+++ b/Dashboard/app/js/lib/views/surveys/survey-group-views.js
@@ -82,7 +82,7 @@ FLOW.Project = FLOW.View.extend({
 
   isPublished: function() {
     var form = FLOW.selectedControl.get('selectedSurvey');
-    return form && form.get('status') === 'PUBLISHED'
+    return form.get('status') === 'PUBLISHED'
   }.property('FLOW.selectedControl.selectedSurvey.status'),
 
   disableFolderSurveyInputField: function() {


Dashboard/app/js/lib/views/surveys/survey-group-views.js
```diff
@@ -82,7 +82,7 @@ FLOW.Project = FLOW.View.extend({
 
   isPublished: function() {
     var form = FLOW.selectedControl.get('selectedSurvey');
-    return form && form.get('status') === 'PUBLISHED'
+    return form.get('status') === 'PUBLISHED'
   }.property('FLOW.selectedControl.selectedSurvey.status'),
```
ichinaski: Not sure I follow the reasoning for this rollback?

ichinaski: Discussed on Skype:

> The rollback was fixing the display of a blank screen... but the root cause of that is some convoluted system of state management with lots of global variables that at some point go out of sync and there is no survey data

> addressing the symptom by checking for a null form solves the appearance.. i.e. no blank screen but later on results in issues like options missing from questions or question types incorrectly set etc... and I felt that is way more dangerous because then a user may save the wrong settings on a question and we are in trouble

> therefore rollback... let the blank page issue persist that forces user to reload the page and have correct state... and then we can sort out the proper state management issues later

ichinaski:
Code reviewed!

PR #1713, opened 2016-06-17
Feature/1605 option split

mtwestra:


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -848,6 +871,92 @@ private String optionCellValue(String value) {
...
+	            			optionFound[i] = true;
+	            			found = true;
+	            			// put in cache
+	            			optionCache.put(cacheId, i);
+	            			break;
+	            		}
+	            	}
+        		}
+
+        		// finally, try to match on code
```
muloem: Perhaps we should try to match by code first and if that fails then match by text.  The current setup is such that there will always be text but the codes may or may not be set.  Also the codes are all or nothing so if there is no code for the first option we don't bother checking for codes for the rest of the responses.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -848,6 +871,92 @@ private String optionCellValue(String value) {
...
+        		if (!found){
+        			for (int i = 0; i < numOptions; i++){
+                		if (code != null && code.equalsIgnoreCase(options.get(i).getCode())  && code.length() > 0){
```
muloem: I think to be more secure we should flip the conditions and check `code.length() > 0` before the match with existing codes.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -848,6 +871,92 @@ private String optionCellValue(String value) {
...
+        		cacheId = qId + text + code;
+        		if (optionCache.containsKey(cacheId)){
+        			optionFound[optionCache.get(cacheId)] = true;
+        			found = true;
+        		}
+
+        		// If it is not in the cache, try to match on text
+        		if (!found){
+	        		for (int i = 0; i < numOptions; i++){
+	            		if (text != null && text.equalsIgnoreCase(options.get(i).getText()) && text.length() > 0){
```
muloem: To be more secure perhaps we should check `text.length() > 0` before `text.equalsIgnoreCase(options...)`

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -291,6 +292,10 @@
     private boolean lastCollection = false;
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
+    private Map<Long,List<QuestionOptionDto>> optionMap = new HashMap<Long,List<QuestionOptionDto>>();
+    private Map<Long,Boolean> allowOtherMap = new HashMap<Long,Boolean>();
+    private Map<String,Integer> optionCache = new HashMap<String,Integer>();
```
muloem: May I suggest we rename this to `optionsPositionCache`  while reading through I was wondering what exactly we are caching as it was a bit confusing with the name

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -951,7 +1060,8 @@ private String sanitize(String s) {
                                         + " - " + level;
                                 createCell(row, offset++, levelName, headerStyle);
                             }
-                        } else {
+                        }
+                        else {
```
muloem: perhaps for readability it could be an idea to create another `if` branch for `OPTION` questions just like we do for `GEO` and `CASCADE` on lines 1029 & 1054 respectively.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -968,11 +1078,34 @@ private String sanitize(String s) {
...
                             }
+
                             createCell(
                                     row,
                                     offset++,
                                     header,
                                     headerStyle);
+
+                            // check if we need to create columns for all options
+                            if (QuestionType.OPTION == q.getType()  && useQuestionId) {
```
muloem: Can you cross check but maybe we have some extra conditions that are used in the `useQID` that you may be missing by only using `useQuestionId`?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -692,6 +699,14 @@ private void writeAnswer(
         for (String cellValue : cells) {
             if (questionType == QuestionType.NUMBER) {
                 createCell(row, col, cellValue, null, Cell.CELL_TYPE_NUMERIC);
+            } else if (questionType == QuestionType.OPTION){
```
muloem: this is more personal style issue but if you used..
 `(questionType == QuestionType.OPTION && (cellValue.equals("0") || cellValue.equals("1")))`
you would skip having the same line repeated on 708 and 711 below.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
mtwestra: I don't think so, for the following reason:

1) when there is a match on the text, the situation is clear.
2) if there is no match on the text, this could be caused by the fact that someone made a small change to the text, for example to fix a typo. In that case, matching on the code acts like a backup.

On the other hand, we can have the situation where options were created first without codes, and given codes later. In that case, the answers sometimes will have a code, and sometimes not. First matching on codes will just create extra work in that case.

I also think codes are more sensitive to errors. Say we have the answer 'yes' and 'no', and we first have the codes 1 and 0, and later we change this to 0 and 1. In that case, answers that have the right text but the wrong code should preferably matched correctly - which is accomplished by matching first on text.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
mtwestra: agreed, change made.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
mtwestra: agreed, done

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
mtwestra: Yes, but the problem is that the first part needs to be done for all columns, including for the first column of an option question. The check for the option question comes after that, and if valid, creates the additional option columns.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
mtwestra: useQID includes the fact that there is a non-null question identifier. However, we don't use this in the creation of the column names, so we don't care about that. In fact, we need to use useQuestionId here.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
mtwestra: Yes, much better. Done.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
muloem: Good point. Makes sense.

muloem:
:+1: code reviewed!

PR #1719, opened 2016-06-22
1637 unify date format on CartoDB maps

kymni:


Dashboard/app/js/lib/views/maps/map-views-common.js
```diff
@@ -599,7 +599,7 @@ FLOW.NavMapsView = FLOW.View.extend({
               +'<li>'
               +'<span>'+Ember.String.loc('_collected_on') +':</span>'
               +'<div class="placeMarkCollectionDate">'
-              +date.toUTCString()
+              +date.toISOString().slice(0,-8).replace("T", " ")
```
muloem: I think you can reuse the same `self.formatDate()` function here as you used later on for the question content, right?

muloem:
:+1: code reviewed!

PR #1722, opened 2016-06-27
Feature/1652 media geodata in reports

stellanl:


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -698,8 +705,14 @@ private void writeAnswer(
 
         int col = startColumn;
         for (String cellValue : cells) {
-            if (questionType == QuestionType.NUMBER) {
+        	if (questionType == QuestionType.NUMBER) {
```
muloem: Seems your environment is using tabs instead of spaces for formatting. Can you check and change this?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -716,9 +729,26 @@ private static String dateCellValue(String value) {
...
-        return ExportImportUtils.formatImage(imagePrefix, value);
-    }
+    private static List<String> mediaCellValues(String value, boolean useQuestionId, String imagePrefix) {
+        List<String> cells = new ArrayList<>();
+        if (value.startsWith("{")) { //JSON, possibly geotagged
+            Media media = MediaResponse.parse(value);
+            String filename = media.getFilename();
+            final int filenameIndex = filename != null ? filename.lastIndexOf("/") + 1 : -1;
+            if (filenameIndex > 0 && filenameIndex < filename.length()) {
+            	cells.add(imagePrefix + filename.substring(filenameIndex));
```
muloem: From the line 735 to here is same as whats happening in `ExportImportUtils.formatImage()`.  May I propose we eliminate the use of `ExportImportUtils.formatImage()` to make it simpler? They we would also not need two branches.  Just one if branch for the case where the geolocation is present.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -266,6 +269,10 @@
         CODE_LABEL.put("en", "Geo Code");
         CODE_LABEL.put("es", "Código Geo");
 
+        ACC_LABEL = new HashMap<String, String>();
+        ACC_LABEL.put("en", "Accuracy (m)");
+        ACC_LABEL.put("es", "Precisi�n (m)");
```
muloem: I am getting an error when compiling this line `error: unmappable character for encoding UTF8`

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
stellanl: Of course. ( I made a new Eclipse workspace just for Flow and got vanilla editor settings...)

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -716,9 +729,23 @@ private static String dateCellValue(String value) {
         return ExportImportUtils.formatDateResponse(value);
     }
 
-    private static String photoCellValue(String value, String imagePrefix) {
-        return ExportImportUtils.formatImage(imagePrefix, value);
-    }
+    private static List<String> mediaCellValues(String value, boolean useQuestionId, String imagePrefix) {
+        List<String> cells = new ArrayList<>();
+        if (value.startsWith("{")) { //JSON, possibly geotagged
```
muloem: I didn't execute but this seems like it will skip the non-JSON formatted responses altogether... right? Perhaps we don't even need this if block. Only check later if there is location data and create the relevant cells.

stellanl: The previous code assumed that photo answers were always in JSON, so yes, the if clause is probably redundant.

muloem:
:+1: code reviewed!

Issue #1741, opened 2016-07-25
Devices w no Phone No are matched with wrong assignments


stellanl:
When looking for an assignment the legacy-handling code tries to find a match for the devicePhoneNo even if it is the empty string. Only null is handled.
This results in devices downloading extra surveys.
It has been exposed recently by Android 6 not allowing MAC address access and the App hotfix akvo/akvo-flow-mobile#448 which reports phoneNo as empty string also on Android 5.1

stellanl:
Test plan: Using an (emulated) Android 6 phone, run the App against an instance where legacy assignments with empty phoneNo exist, such as  dnhmali.akvoflow.org
Check that no surveys are downloaded before an asssignment is made.

janagombitova:
Created an assignment on UAT1 for Kasper's Android 6 phone and then asked Geert to install the app on his Android 6 phone and see if Kasper's assigned survey was downloaded to his phone and it wasn't.

👍  fix passes test

janagombitova:
passes 2nd round of regression tests 👍

PR #1742, opened 2016-07-25
[#1741]Do not try to match an empty devicePhoneNo to a legacy assignment

stellanl:
This should help with most of the ghost survey assignments


muloem:
:+1: code reviewed!

Issue #1657, opened 2016-05-11
Remove language dropdown from 'public' page (estimated: 2)


janagombitova:
Translations do not work on 'Log in = public page'. When a user is not logged in the dashboard and selects a language from the dropdown the language does not change and remains in English. 

Once he logs into the dashboard, he can change the language, lets say to Spanish. After logging out the public page is shown and it is in English again. 

We either need to enable translations on the public page as well (not sure if this is a good idea considering the changes implemented in the last release regarding loading translations) or we need to hide the language selector from the dropdown in the public page to avoid confusion.

muloem:
We should probably hide the language selection button. The reason is that language is now user specific. i.e. if there is no logged in user, we only use the default since we can't tell which language the anonymous user wants.

janagombitova:
That makes perfect sense and if the option is not visible to the user, they will not get confused.

gabemart:
# Test Plan

1) Open the logged-out 'public' page and verify that the language menu does not appear

2) Log in, and verify the language menu does appear and behaves correctly

janagombitova:
passed test plan 👍

janagombitova:
passes 1st and 2nd regression test 👍

PR #1746, opened 2016-08-01
[#1657] Remove language dropdown from public page

gabemart:
#1657


muloem:
@gabemart looking at the entire `headers-common-public` none of the markup in there seems to be visible infact. Perhaps we can eliminate it altogether?

gabemart:
@muloem I had that thought too - I'm not sure why it was there in the first place so I was hesitant to remove it. I will do some more investigation to see if removing it breaks anything

gabemart:
@muloem Everything seems to work with that removed :+1:

Issue #1689, opened 2016-06-06
Development environment setup/cleanup


muloem:
- Improvements and cleaning up of the development environment

janagombitova:
In the planning poker meeting we decided to not estimate this issue as it is about updating the documentation on setting up a dev environment, making sure the templates are in place > this is an ongoing process and we progressively need to make sure all information is in place when doing so

PR #1748, opened 2016-08-03
[#1689] Update remote api scripts for local datastore access without service accounts

muloem:


kymni:
Code reviewed and tested. Works fine!

Issue #1684, opened 2016-05-30
Invalid empty repeat question group error (estimated: 8)


muloem:
When importing data for spreadsheets, in some cases there are empty phantom rows after the last valid row of data.  In these cases the raw data importer is throwing the error of an empty repeat question group column, and preventing valid reports from being imported.  We need to fix the check to ensure that phantom row are ignored.

stellanl:
Test plan:
1. export a survey for data cleaning.
2. Open the spreadsheet and create some content in a cell below the last row with content.
3. Delete the cell again.
4. Save the spreadsheet and import it.

janagombitova:
👍  passed test plan

muloem:
@janagombitova can you also update the test plan to include a repeat question group set of responses with an empty row in between? i.e with even the number in the repeat column missing. thanks!

janagombitova:
@muloem yes sir!

janagombitova:
passes 2nd round of regressions as well (with all the odd possible things people can do with RQGs) 👍

stellanl:
The RawDataSpreadsheetImporter.validate() function still does not cope with phantom rows at the end. It ierates ove allr rows and expects to find a repeat number on all of them.

janagombitova:
👍 implementation passes test plan successfully and the validation is the repeated column is truly empty still works as well as with empty repeated group rows

PR #1758, opened 2016-08-30
[#1684]Use same criterion for end of sheet (connect #1684)

stellanl:


muloem:
@stellanl I reviewed this branch. Its ready to merge.  I made three small changes that you can perhaps cross check and then merge if you are ok with them.

PR #1760, opened 2016-09-06
Issue/1759 enable monitoring data transfer

stellanl:
Uses surveyedLocaleIdentifier to connect monitored surveys together.

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
@@ -71,6 +71,7 @@ import org.waterforpeople.mapping.domain.response.value.Media;
 import org.waterforpeople.mapping.serialization.response.MediaResponse;
 
 import com.gallatinsystems.common.util.JFreechartChartUtil;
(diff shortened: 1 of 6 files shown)


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -1632,7 +1632,7 @@ public static void main(String[] args) {
         Map<String, String> criteria = new HashMap<String, String>();
         Map<String, String> options = new HashMap<String, String>();
         options.put(LOCALE_OPT, "en");
-        // options.put(TYPE_OPT, RAW_ONLY_TYPE);
+        options.put(TYPE_OPT, RAW_ONLY_TYPE); //For monitored-transfer test
```
muloem: Do we need this for the time being?

muloem:
@janagombitova given that we do not want to include the monitoring data import for the time being, lets say we close this PR until we move that decision further?

janagombitova:
@muloem could not agree more. But if we have the very exeptional cases where we might need to do this in a very controlled way, would that still be possible (as we have done the tests on SNV instances)?

Issue #1772, opened 2016-09-30
Fix NPE when stored counts are null


iperdomo:
```
Failed to process zip file:bb54c05e-ea73-4a83-ab48-94ca047f76f8.zip : null
java.lang.NullPointerException
    at org.waterforpeople.mapping.domain.SurveyInstance.updateSummaryCounts(SurveyInstance.java:396)
    at org.waterforpeople.mapping.dao.SurveyInstanceDAO.save(SurveyInstanceDAO.java:159)
    at org.waterforpeople.mapping.app.web.TaskServlet.processFile(TaskServlet.java:204)
    at org.waterforpeople.mapping.app.web.TaskServlet.ingestFile(TaskServlet.java:388)
    at org.waterforpeople.mapping.app.web.TaskServlet.handleRequest(TaskServlet.java:354)
    at com.gallatinsystems.framework.rest.AbstractRestApiServlet.executeRequest(AbstractRestApiServlet.java:85)
    at com.gallatinsystems.framework.rest.AbstractRestApiServlet.doPost(AbstractRestApiServlet.java:61)
   [...]
```

janagombitova:
@iperdomo could you add a test plan for the issue? Thanks

iperdomo:
@janagombitova unfortunately i don't have a way of reproducing the error. It just annoying to have those `NullPointerException` in the email :wink: 

Some of us are recipients of error stack traces when something goes wrong processing a data file (the _stacktrace_ copy i pasted in the issue).

I can forward some emails with the same error.

osg74:
Small enough fix, which should stop error emails from this particular code fragment.

janagombitova:
@osg74 I do not know how to check this one as I am not in the email list (was one way how I thought I can test it out). Can you please decide on if this implementation is ready for releasing? Thanks

osg74:
Should be ready for release -- I've added myself to the mailing list to catch any similar errors in future.

PR #1773, opened 2016-09-30
[#1772] Fix NPE when stored count is null

iperdomo:


GAE/src/org/waterforpeople/mapping/domain/SurveyInstance.java
```diff
@@ -393,7 +393,7 @@ public void updateSummaryCounts(boolean increment) {
                 }
 
                 // update and save or delete
-                long count = questionSummary.getCount();
+                long count = questionSummary.getCount() == null ? 0 : questionSummary.getCount();
```
stellanl: Looks good

Issue #1800, opened 2016-10-27
Caddisfly results render as string in Carto maps


janagombitova:
In Carto maps the results are rendered as a string. The answers need to be shown in the same way across the entire dashboard. Ideally in the same way they are shown on non-Carto maps. 

<img width="1250" alt="screen shot 2016-10-24 at 15 06 03" src="https://cloud.githubusercontent.com/assets/12456965/19646831/8fc19e88-99fb-11e6-93e0-c59861359f6b.png">

<img width="1227" alt="screen shot 2016-10-24 at 15 09 28" src="https://cloud.githubusercontent.com/assets/12456965/19646899/eed44eb6-99fb-11e6-896d-c3c7917a5b69.png">
_(rendering on non-Carto maps)_

You can find the data points here https://uat1.akvoflow.org/admin/ > Map tab > 1.9.11 UAT Folder > Jana UAT tests > 1.9.11 UAT Non MF Survey (All questions) 1st test > points are located in Slovakia in the North part of the country

janagombitova:
👍  implementation passes test plan

PR #1844, opened 2016-11-28
#1800 issue fix (connect #1800) 

kymni:
## Checklist
* [ ] Click on point with caddisfly response
* [ ] Response should appear as an image in the side panel same as in issue definition


Dashboard/app/js/lib/views/maps/map-views-common.js
```diff
@@ -774,6 +774,15 @@ FLOW.NavMapsView = FLOW.View.extend({
                         clickedPointContent += srcAttr + signatureJson.image +'"/></div>';
                         clickedPointContent += '<div class="signedBySection">'+Ember.String.loc('_signed_by') +': '+signatureJson.name+'</div>';
                         break;
+                      case "CADDISFLY":
```
muloem: I was just wondering if this section isn't doing the same thing as `FLOW.renderCaddisflyAnswer()` in `views.js`. Can you review and if its the case then reuse the function. 👍

kymni: You are right, change applied

Issue #1845, opened 2016-11-28
Copyright change - 2012-2017


janagombitova:
As the next Flow release is planned in for January 12, we can already prepare the copyright to reflect the year change. 

## Test plan
1. Log into a dashboard
1. Check that the copyright shows 2012-2017

janagombitova:
@stellanl could you take care of this one please as well so we are ready for the January release? Thank you

janagombitova:
👍  ready to go

PR #1884, opened 2016-12-13
[1845]Update copyright headers to 2012-2017 for January release. (Connect #1845)

stellanl:
Copyright year span 2012-2016 would be outdated by the time of the release.
#### The solution
Changed to 2012-2017
## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 LICENSE.md, v1.0, 24 May 2012 [co, ogl]
 
-Copyright (C) 2010-2012 Stichting Akvo (Akvo Foundation)
+Copyright (C) 2010-2016 Stichting Akvo (Akvo Foundation)
 
 This file is part of Akvo Flow.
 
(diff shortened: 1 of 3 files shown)


osg74:
Should we not also update the copyright dates in comment headers and the main license file?

stellanl:
Main license file, sure. The dates on individual files only applies to that file and should only be updated if we change that file.

stellanl:
Actually, the copyright notice in the license file applies only to the license file itself. We changed it this year, but only to turn AKVO into Akvo...

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 LICENSE.md, v1.0, 24 May 2012 [co, ogl]
 
-Copyright (C) 2010-2012 Stichting Akvo (Akvo Foundation)
+Copyright (C) 2010-2016 Stichting Akvo (Akvo Foundation)
```
muloem: @stellanl Can you make this quick fix? thanks!

Issue #1846, opened 2016-11-28
Improve UX around Log in/Log out link


janagombitova:
Currently the Log in/Log out button is located in the footer of the dashboard, hidden with links to download the Flow app and support, etc. The need is to find a more straightforward location for this link, so the user can find it easily and intuitively.  

![screen shot 2016-11-28 at 15 19 53](https://cloud.githubusercontent.com/assets/12456965/20671585/2c365e14-b57e-11e6-8f48-c2af1e8f12ae.png)

loicsans:
Swap language dropdown with logout call to action and move login to header on public page

janagombitova:
I like the swap, but there is one thing that you need to change. Now the language dropdown in visible when the user is not logged in. This should not be shown, as the public page is only in English and language option are only available after logging in. 

I am a bit uncomfortable that we now have documentations, terms and conditions and language all packed in the footer together and I agree with your idea for a "admin" page in the future. For now, could we somehow divide the footer (in a separate issue)?

janagombitova:
I love the new location of the log in/log out. ready to roll 👍

Issue #1836, opened 2016-11-16
Do not rely on default locale for machine-readable dates


stellanl:
As touched on in #1835, when creating a SimpeDateFormatter to format or parse dates read or written by another device, it is bad practice not to specify a locale. Potentially, if the sender and receiver are set to different locales, communication could fail or data be corrupted.
The standard solution is to specify a locale that exists on all Android devices, en_US.
This is true even if the format specified is purely numeric. Not all locales count years from the same beginning...

PR #1902, opened 2017-01-12
[#1836] swap login and language selector (connect #1846)

loicsans:


muloem:
* When I look at the screen, the login and logout links are almost imperceptible... Maybe its too far aligned to the right. Could it be perhaps not entirely against the right hand border?

* The language drop down also seems a little out of place among these links at the bottom... at least based on what I have seem as a pattern from other applications

* The _akvo.org_ button just before the language button seems to be out of place.  If we want a link to akvo.org shouldn't we just take advantage of the copyright notice and create a link there?


_Login_ 

screen shot 2017-01-24 at 11 31 56


_Logout_

screen shot 2017-01-24 at 11 26 30

muloem:
Good job! This is ready to go.  We have to merge the refactor branch first and then follow with this one. 👍

Issue #1643, opened 2016-05-04
Export Caddisfly data to reports


mtwestra:
We need to export the caddisfly data to reports. 
Related issue: [#1577]

Items to consider:
- format of output when using questionIds
- format of output when not using questionIds
### Format of Caddisfly result data:

JSON string with properties:
- type: always 'caddisfly'
- name: display name of test
- uuid: unique identifier of test
- image: image url (optional)
- result: array of result objects, containing a result object for each specific result returned by the test.

Result object:
- name: display name of result 
- value: numeric value of result
- unit: unit of result
- id: id of result

Example:
{
  "type":"caddisfly",
  "name":"Chlorine and Free Chlorine",
  "uuid":"bf1c19c0-9788-4e26-999e-1b5c6ca28111",
  "image":"45e61553-8568-4719-b922-7f678903e594.png",
  "result":[{
    "name":"Total Chlorine (ppm)",
    "value":10,
    "unit":"ppm",
    "id":1
  },{
    "name":"Free Chlorine (ppm)",
    "value":1.6,
    "unit":"ppm",
    "id":2
  }]
}
### Proposal:
- Caddisfly data is always split into multiple columns
- Columns created are one for each result, and one for the image URL
- Caddisfly data cannot be changed by importing excel files
- The names of the columns contain the name of the result, and the unit

mtwestra:
Say, we have the question text = "Chlorine measurement"
Taking the above example result, this would be exported as follows:

1) using questionId
* 28772937|Chlorine measurement|TEST_TYPE
* --CADDISFLY--|Total Chlorine (ppm)
* --CADDISFLY--|Free Chlorine (ppm)
* --CADDISFLY--|image-URL

2) not using questionId
* Chlorine measurement|TEST_TYPE
* --CADDISFLY--|Total Chlorine (ppm)
* --CADDISFLY--|Free Chlorine (ppm)
* --CADDISFLY--|image-URL

The first column would then contain the test type, and brand, for example:
Chlorine and Free Chlorine (HACH 27450, 50pcs)

mtwestra:
The above system is a bit too rigid in the case of backward compatibility. Therefore, we will change to an implementation where a measurement is spread over two columns: one which has the parameter, and one which has the value:

Say, we have the question text = "Chlorine measurement"
Taking the above example result, this would be exported as follows:

1) using questionId
28772937|Chlorine measurement|TEST_TYPE
--CADDISFLY--|PARAMETER
[...]

PR #1911, opened 2017-01-18
[#1643] Import and export of reports for forms containing caddisfly questions (connect #1643)

muloem:

* We introduce the import and export code for forms containing caddisfly questions.  
  * For the export, we include the results of caddisfly questions, with the possibility for each result on a single column
  * For the import we ignore all caddisfly related columns

## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/akvo/flow/domain/DataUtils.java
```diff
@@ -28,15 +31,17 @@
...
 
+    private static final Logger log = Logger.getLogger(DataUtils.class);
+
     public static final ObjectMapper JSON_OBJECT_MAPPER = new ObjectMapper();
 
     public static String[] optionResponsesTextArray(String optionResponse) {
         String[] responseArray = null;
 
-        if (optionResponse == null || optionResponse.trim() == "") {
+        if (optionResponse == null || optionResponse.trim().equals("")) {
```
valllllll2000: For this case, it is irrelevant but I would recommend as general practice to replace somestring.equals("") by "".equals(somestring) that way you avoid possible NPE, obviously in this case it will not happen.

GAE/src/org/akvo/flow/domain/DataUtils.java
```diff
@@ -52,14 +57,14 @@
 
         return responseArray;
     }
-    
+
     public static String[] cascadeResponseValues(String data) {
         String[] values = null;
 
-        if (data == null || data.trim() == "") {
+        if (data == null || data.trim().equals("")) {
```
valllllll2000: same as line 41

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -1922,7 +1909,7 @@ public static void main(String[] args) {
         options.put(LOCALE_OPT, "en");
         // options.put(TYPE_OPT, RAW_ONLY_TYPE);
         options.put(LAST_COLLECTION_OPT, "false");
-        options.put("useQuestionId", "true");
+        options.put("useQuestionId", "false");
```
stellanl: Is this intentional? I have sometimes changed this here during testing...

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -817,138 +819,96 @@ private static String dateCellValue(String value) {
...
         return true;
     }
 
     /*
-     * Creates the cell values for a caddisfly question.
+     * Creates the cells containing responses to caddisfly questions
      */
     @SuppressWarnings("unchecked")
     private List<String> caddisflyCellValues(Long questionId, String value,
             Integer numResults, Boolean hasImage, String imagePrefix) {
```
valllllll2000: numResults seems to be unused

GAE/src/org/akvo/flow/domain/DataUtils.java
```diff
@@ -135,4 +140,25 @@ public static String parseSignatory(String value) {
...
+
+    /**
+     * Parse a caddisfly response string and return a corresponding map
+     *
+     * @param caddisflyValue
+     * @return
+     */
+    @SuppressWarnings("unchecked")
+    public static Map<String, Object> parseCaddisflyResponseValue(String caddisflyValue) {
+        Map<String, Object> caddisflyResponseMap = new HashMap<String, Object>();
```
valllllll2000: new HashMap<>(), explicit type is no longer necessary

GAE/src/org/akvo/flow/domain/DataUtils.java
muloem: Good point. and perhaps won't require also the verbosity of two statements.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
muloem: This was for testing but has no influence in production since the export() method is where the actual execution starts.

GAE/src/org/akvo/flow/domain/DataUtils.java
valllllll2000: Yes but in this case calling trim() may still produce NPE so the null check is required

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
muloem: Good catch! I will remove the unused parameter.

Issue #1920, opened 2017-02-01
Dashboard footer design improvements


loicsans:
Specifically looking at better integration of language selector

janagombitova:
@loicsans  currently on UAT1 in the public page (not logged in state) the user sees the language dropdown in it's new location. The location is a much better improvement but the language dropdown is not supposed to be visible on the public page, only on the private one, as we do not translate the public page. 

So it is same as with the Download Flow app link - they both are only shown once you log in 

<img width="1075" alt="screen shot 2017-02-01 at 13 01 41" src="https://cloud.githubusercontent.com/assets/12456965/22506057/a78cca8e-e87e-11e6-9e98-25f38f6d9c0e.png">

janagombitova:
The language dropdown is not showing anymore in the public page - what is now resoled 👍 

@loicsans considering the location of the dropdown is the same as it was, before we talked about improving the design of the footer that triggered this issue, I am going to assume this is still a work in progress.

loicsans:
Reopening the issue to commit improvement to language selector placement

muloem:
Just a question.. in the public page where there is no language button I see the footer links are right-aligned. Is this intentional?

janagombitova:
Footer appears properly on public and private pages, in all browser types and links work fine as well as changing translations. Ready to go

PR #1941, opened 2017-02-14
[#1920] Separated language selector from core links in footer (connect #1920)

loicsans:


loicsans:
screen shot 2017-02-14 at 13 20 17
It should look like that

Issue #1942, opened 2017-02-14
Change `FormInstance.formId` to parse both String and Long data types


muloem:
With the changes introduced in https://github.com/akvo/akvo-flow-mobile/issues/614, the `FormInstance.surveyId` field is now a string when received from the app. This will cause an error if a corresponding change is not made on the dashboard backend. 

In addition we need to ensure that backwards compatibility is maintained so that the apps that aren't updated are also correctly processed.

muloem:
# Test Plan

1. This requires having a device with an APK older than 2.2.10.1 and _manually assigned_ with a survey that contained a form with a "/" in the form name.

2. Fill out responses, submit them and make sure they correctly appear on the dashboard

PR #1950, opened 2017-02-16
Parse faulty device zips (connect #1942)

muloem:

* When processing device zip files that may have erroneously added a string to the `formId` element, the file parsing encounters an error and does not complete.

#### The solution

* We now parse the `formId` as a string and later check that it is indeed a number.  If not, we fall back to using `questionId`, from the responses, to look up the `formId`.

## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


GAE/src/org/waterforpeople/mapping/serialization/SurveyInstanceHandler.java
```diff
@@ -86,27 +91,58 @@ public static SurveyInstance fromJSON(String data) {
...
+            }
+        }
+
+        return null;
+    }
+
+    /*
+     * Retrieve the formId from the corresponding Question entity in the datastore
+     */
+    private static Long retriveFormIdByQuestionId(String questionId) {
```
valllllll2000: A minor detail but retriveFormIdByQuestionId should be retri**e**veFormIdByQuestionId

GAE/src/org/waterforpeople/mapping/serialization/SurveyInstanceHandler.java
```diff
@@ -86,27 +91,58 @@ public static SurveyInstance fromJSON(String data) {
...
+        Question question = new QuestionDao().getByKey(Long.parseLong(questionId));
```
valllllll2000: is it possible that questionId is null or empty? or can it fail because this question has been deleted?

GAE/src/org/waterforpeople/mapping/serialization/SurveyInstanceHandler.java
muloem: Thats no minor detail!

GAE/src/org/waterforpeople/mapping/serialization/SurveyInstanceHandler.java
muloem: Also possible.  Let me put a catch in place for that.

Issue #1948, opened 2017-02-16
Visually illustrate disabled buttons


muloem:
After the changes in #1869, we are no longer able to distinguish between disabled and enabled buttons for survey and folder creation.  This is confusing as the user thinks the action is available but nothing happens when he/she clicks.

<img width="1240" alt="screen shot 2017-02-16 at 12 08 36" src="https://cloud.githubusercontent.com/assets/533336/23019132/1bd3b5e2-f441-11e6-9acb-236441681e4b.png">

janagombitova:
I have tested this issue out and can confirm, these two buttons seem enabled (full colour and also hovering over works, but clicking on them does not trigger action) even though the user cannot use them. Other elements that should be disabled based on permissions work as expected. 

If a user has no folder related permissions, so he should not be able to use Add folder and Create new survey (as it is currently set up).  

If a user had all folder and form related permissions, but connected to a particular folder Add folder and Create new survey are only enabled once in that particular folder. 

If a user has all folder and form related permissions to all folders on the instance, then Add folder is enabled on the main survey page, but Create new survey is not, only until in a folder. Only the Akvo super admins can use that function also on the main page. So this one needs to be visually disabled as well.

janagombitova:
## Test plan: 

1. Make sure create survey button disabled outside of folder and info window is shown 
1. Make sure button is enabled inside a folder 
1. Check with different browsers and user permissions

janagombitova:
👍  works like a charm

PR #1951, opened 2017-02-17
[#1948] disabled buttons should appear as so (connect #1948)

loicsans:


loicsans:
screen shot 2017-02-17 at 12 55 01

Issue #1952, opened 2017-02-17
Load d3 javascript library only when user navigates to reports tab


kymni:
D3 library currently loaded when dashboard is first loaded. It should only be loaded when needed in the reports tab

kymni:
## Test plan
*Load the dashboard
*Confirm that d3 library is missing from automatically loaded libraries
*Navigate to reports tab
*Confirm that d3 library is now loaded

janagombitova:
https://github.com/akvo/akvo-flow/pull/1953#issuecomment-291405973 see this comment that explains why the issues were not merged but only closed

PR #1953, opened 2017-02-17
#1952 load d3 library when user opens the reports tab (connect #1952)

kymni:
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


muloem:
@kymni I will close this one for now.  It addresses the issue but I believe we better handle it once and for all with the module packaging setup.

Issue #1959, opened 2017-02-22
Fix NPE that prevents users tab from loading


muloem:
In some cases, a user has been added directly to the dashboard through the GAE console and is missing the `superAdmin` property.  This prevents the dashboard users tab from loading because it causes a NPE.

muloem:
# Test Plan

1. Edit a user in the GAE console and delete the `superAdmin` property

2. Load the dashboard and verify that you are still able to see the list of users in the users tab

janagombitova:
👍  works like a charm

PR #1960, opened 2017-02-22
Check for empty `superAdmin` property on user entity (connect #1959)

muloem:
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/app/web/rest/UserRestService.java
+++ b/GAE/src/org/waterforpeople/mapping/app/web/rest/UserRestService.java
@@ -110,7 +110,7 @@ public class UserRestService {
     }
 
     private boolean isSuperAdminRole(User user) {
-        return (user.isSuperAdmin()
+        return user.isSuperAdmin()
                 || user.getPermissionList()
                         .equals(Integer.toString(AppRole.SUPER_ADMIN.getLevel()));
     }
--- a/GAE/src/org/waterforpeople/mapping/app/web/rest/UserRestService.java
+++ b/GAE/src/org/waterforpeople/mapping/app/web/rest/UserRestService.java
@@ -110,7 +110,7 @@ public class UserRestService {
     }
 
     private boolean isSuperAdminRole(User user) {
-        return user.isSuperAdmin()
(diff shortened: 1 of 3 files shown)


GAE/src/org/waterforpeople/mapping/app/web/rest/UserRestService.java
```diff
@@ -110,7 +110,7 @@ private boolean isUserAdminRole(User user) {
     }
 
     private boolean isSuperAdminRole(User user) {
-        return user.isSuperAdmin()
+        return (user.isSuperAdmin() != null && user.isSuperAdmin())
```
osg74: A better fix would be to isolate any null handling to the User.setSuperAdmin() setter so that any subsequent isSuperAdmin() calls never return null (and we don't have to fix other cases where the getter is called).

https://github.com/akvo/akvo-flow/blob/b9f08b63089dece914189ab00a146f969a12f985/GAE/src/com/gallatinsystems/user/domain/User.java#L84

muloem:
@osg74 Good observation. 👍  I made the changes.  I figured its enough to check on the side of the getter by using the Boolean constants

GAE/src/org/waterforpeople/mapping/app/web/rest/UserRestService.java
```diff
@@ -110,7 +110,7 @@ private boolean isUserAdminRole(User user) {
     }
 
     private boolean isSuperAdminRole(User user) {
-        return (user.isSuperAdmin() != null && user.isSuperAdmin())
+        return (user.isSuperAdmin()
                 || user.getPermissionList()
                         .equals(Integer.toString(AppRole.SUPER_ADMIN.getLevel()));
```
osg74: This exposure of the permissionList to determine the user role outside the User class is also unnecessary -- rather move this check back into the isSuperAdmin() getter as a refactoring.

osg74:
Added resulting issue for permissionList refactoring mentioned above:
https://github.com/akvo/akvo-flow/issues/1962

Issue #1968, opened 2017-03-02
Exports with rollup sheets break on JSON data


stellanl:
The logic for creating geographical rollup sheets does not handle JSON responses well.
They also expose a bug in GraphicalSurveySummaryExporter.
The try-again loop at L1426 is broken, since it tests for existence of a different sheet name than it then uses to create a new one with...

stellanl:
Test plan: From akvoflow-17, run a report on 
OXFAM WASH > Post Ebola Public Health Units > Post Ebola Public Health Units RNA -> form ID: 170300001
You should a: get a report, and b: see that the Spreadsheet sheets have JSON-less names

janagombitova:
Implementation passes test plan 👍

janagombitova:
Checked also on the original survey that cause the problem (made a copy of it in uat) and all works fine now 👍 👍

PR #1969, opened 2017-03-02
[#1968]Extract JSON content; test the same name that would be created (Connect #1968)

stellanl:
#### The solution
a. Strip the Json from the answers when they are inserted into the rollup list
b. Test with the same sheet name that we will use to create if it does not exist.

## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [ ] Code formatting
* [ ] Documentation


stellanl:
Actually, i think this is complete. DataUtils.jsonResponsesToPipeSeparated() returns an empty string if the argument is not JSON, and I test for that.

Issue #1975, opened 2017-03-03
Improve how form tabs look


janagombitova:
The current form design looks like folders in a binder. In the separate form tabs we show the name of the form. However, the design does not show the full name of the form and neither the full text "Add new form". 

@loicsans Can you do some of your magic here? 

<img width="374" alt="screen shot 2017-03-03 at 18 23 11" src="https://cloud.githubusercontent.com/assets/12456965/23561497/c9c7dbe8-003e-11e7-9546-f3b33540f4e5.png">

<img width="732" alt="screen shot 2017-03-03 at 18 25 49" src="https://cloud.githubusercontent.com/assets/12456965/23561512/d8bd8e22-003e-11e7-938a-e5c16263f052.png">

muloem:
@janagombitova @loicsans is this the preferred solution? i.e. scrolling?  Could we maybe go for something where we keep the first x characters and only scroll of the number of forms becomes too many? otherwise a single form with a long name will make it look weird....  and will of course make the _add new form_ button disappear

<img width="1227" alt="screen shot 2017-03-08 at 16 25 12" src="https://cloud.githubusercontent.com/assets/533336/23710276/70a3164a-041c-11e7-9e26-cc81db1b2d04.png">

janagombitova:
I agree with Mulo. We need a way for the user to see the number of forms he has, to see the form name (which you can see after selecting a form), to select a form and see the full **Add form** button  

Maybe the implementation before the fix was ok, if we solve the fact that the **Add form** button was not fully visible and thus hard for user to find.

janagombitova:
I was thinking about this one a bit more over the last days and realised my initial comment that the full form name is not visible in the tabs is not such a big issue, as the user sees the full name after selecting the particular form. 

The problem is that the `Add form` button is not fully shown and visible at all times. The button should be as the ones to add a new survey visible for the user at all times and fully regardless of the number of forms in the survey and their name length

janagombitova:
I cross checked on beta1 on Safari and Chrome and all appears well. The button does not float over and there is not space under the tabs. Appears good to me 👍

PR #1978, opened 2017-03-06
[#1975] Fixed the form name being cut inside the tab (connect #1975) 

loicsans:


loicsans:
screen shot 2017-03-21 at 18 01 11
Long form titles are now displayed with an ellipsis inside tabs which are constraint to a max width.

loicsans:
Taking this further: the add form button, in case of monitoring unabled is moved aligned to the tabs themselves on the right for a more intuitive location. Also removed the "Forms:" header.
screen shot 2017-03-21 at 18 34 14

loicsans:
The last commit should look like this: 
screen shot 2017-04-05 at 14 44 43

janagombitova:
@muloem just a note in case you started review already, @loicsans will also remove the white tab from the screen

screen shot 2017-04-05 at 15 16 10

loicsans:
it doesn't for me
screen shot 2017-04-06 at 7 19 53

loicsans:
@muloem I cannot reproduce the gap you see between the tabs and the content. Caching issue again?

janagombitova:
Hi @loicsans and @muloem, could the gap be because you have the page zoomed out? I had it last time with a different part of Flow that when i was looking at the page in 100% everything was fine, but when going to 90% things were out of position.

muloem:
@loicsans @janagombitova I am deploying this branch to beta1 to confirm.  On my env its neither cached nor at a special zoom.

janagombitova:
@mulo and @loicsans I cross checked on beta1 on Safari and Chrome and all appears well. The button does not float over and there is not space under the tabs. Appears good to me 👍

PR #1980, opened 2017-03-09
Issue/1203 remove copy suffix (connect #1203)

stellanl:
Everybody got questionIds appended with "_copy".
This was not checked for uniqueness, and was annoying overkill when copying a whole form.
#### The solution
On whole-form copies, append nothing. On within-form copies, tries appending _1, _2 etc until it is unique.
## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


GAE/src/com/gallatinsystems/survey/dao/SurveyUtils.java
```diff
@@ -101,12 +106,44 @@ public static Survey copySurvey(Survey source, SurveyDto dto) {
...
     public static QuestionGroup copyQuestionGroup(QuestionGroup sourceGroup,
             QuestionGroup copyGroup, Long newSurveyId, Map<Long, Long> qDependencyResolutionMap) {
 
         final QuestionDao qDao = new QuestionDao();
         final Long sourceGroupId = sourceGroup.getKey().getId();
         final Long copyGroupId = copyGroup.getKey().getId();
+        final boolean sameSurvey = sourceGroup.getSurveyId() == newSurveyId;
+        Set<String> idsInUse = null;
+        
+        if (sameSurvey) {
```
muloem: Having this block within the `copyGroup` function,means we'll make redundant queries for these entities when copying an entire survey and will definitely have a performance impact.  Can we instead update the method signature of the copy group to take in a `Set<String> questionIdsList` parameter that we provide from outside of the function?

It will also serve the purpose of us checking whether we need to even verify the question ids. i.e. in some cases we are copying surveys without question ids but if we leave it like this, this code will still be executed for those cases.

GAE/src/com/gallatinsystems/survey/dao/SurveyUtils.java
```diff
@@ -168,14 +215,28 @@ public static Question copyQuestion(Question source,
...
+            if (idsInUse != null) { //must avoid these
+                String newId = source.getQuestionId() + "_1";
+                int index = 2;
+                while  (idsInUse.contains(newId)) {
+                    newId = source.getQuestionId() + "_" + index++;
+                }
+                tmp.setQuestionId(newId);
+                //one more to avoid
+                idsInUse.add(newId);
+                log.log(Level.INFO, "Changing QuestionId from " + source.getQuestionId() + " to " + newId);
```
muloem: Perhaps we should only log this at debug level?

stellanl: OK, changing it to FINE, along with other messages about the question copying

GAE/src/org/waterforpeople/mapping/app/web/DataProcessorRestServlet.java
```diff
@@ -137,8 +139,28 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
...
+                final QuestionDao qDao = new QuestionDao();
+                
+                Long surveyId = originalQuestionGroup.getSurveyId();
+                Survey s0 = sDao.getById(surveyId);
+                final Long surveyGroupId = s0.getSurveyGroupId();
+                List<Survey>sList = sDao.listSurveysByGroup(surveyGroupId);
+                for (Survey s : sList) {
+                    List<QuestionGroup>qgList = qgDao.listQuestionGroupBySurvey(s.getKey().getId());
+                    for (QuestionGroup qg : qgList) {
+                        List<Question> qList = qDao.listQuestionsInOrderForGroup(qg.getKey().getId());
```
muloem: Given the ordering of questions is not necessarily needed for this particular step, I suggest we skip querying for the question groups and directly retrieve the list of questions for a particular survey (form) with `QuestionDao().listQuestionsBySurvey()`

muloem: Given the ordering of questions is not necessarily needed for this particular step, I suggest we skip querying for the question groups and directly retrieve the list of questions for a particular survey (form) with `QuestionDao().listQuestionsBySurvey()`

Issue #1985, opened 2017-03-10
Exposed /remote_api on a different endpoint to enable tracing


iperdomo:
The first implementation of the new Flow API will use [Remote API](https://cloud.google.com/appengine/docs/standard/java/tools/remoteapi) for accessing the Datastore. In order to isolate requests done as part of a _usual_ remote api call (e.g. Datascripts, daily statistics) we want to expose the same functionality on a different endpoint. Ideally this new endpoint should allow us to group/trace/correlate calls and potentially calculate relative costs.

To achieve the latter the Remote API client **must** configure the path with a unique identifier. e.g.

````java
        String traceId = UUID.randomUUID().toString();        
        RemoteApiOptions options = new RemoteApiOptions().server(instanceUrl, port);
        
        options.useServiceAccountCredential(serviceAccount, serviceAccountPvk);
        options.remoteApiPath("/traced_remote_api/" + traceId);
````

See: https://cloud.google.com/appengine/docs/standard/java/tools/remoteapi/javadoc/com/google/appengine/tools/remoteapi/RemoteApiOptions.html#remoteApiPath-java.lang.String- 

If the client makes several requests with the same _path_ we can group and analyze them.

janagombitova:
@iperdomo and @muloem there is no test plan for this issue and I am not sure if the successful review is sufficient here to confirm these changes are ready for a release. Can you provide some guidance? Thanks

iperdomo:
@janagombitova this change is a low level one, but i can provide some code that you can run and verify that works by checking the output and the Logs in GCP dashboard.

valllllll2000:
@janagombitova , I have tested the code sent to me by @iperdomo and it worked as expected

PR #1986, opened 2017-03-10
[#1985] Enable tracing of Remote API calls with a different endpoint (connect #1985)

iperdomo:
- The client must configure the `remoteApiPath` to be something unique
  in order to group calls for a given API request
- The Remote API functionality remains unchanged, we just expose it in
  in a different path: /traced_remote_api/*
- The client can use *anything* after the last slash, e.g.
  /traced_remote_api/<form-id>/<uuid>
  /traced_remote_api/<user>/<form-id>/<uuid>
  The only restriction is to use the same path within the set of requests

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


GAE/war/WEB-INF/web.xml
```diff
@@ -480,6 +480,14 @@
 		<servlet-name>RemoteApi</servlet-name>
 		<url-pattern>/remote_api</url-pattern>
```
muloem: Taking the deleted class in to account, should we then also only add an extra `<url-pattern></url-pattern>` here for tracing and remove the servlet definition below?

iperdomo:
@stellanl PR updated based on @muloem review

Issue #1736, opened 2016-07-15
Optimise retrieval of questions for report exports


muloem:
When generating reports to be used in external systems, Flow services retrieves a set of questions without options in a first call and in some cases goes on to retrieve the same set of questions with options, thus retrieving the same set of data twice.  Ideally we should merge these calls to save on the amount of reads from GAE and to keep the report generation time to a minimum.

stellanl:
In GraphicalDataSummaryExporter.export(), we always call loadAllQuestions() [l 337] and then conditionally loadFullQuestions() [l 353]

stellanl:
I cannot find a place where loadFullQuestions() refetches the question list.
Or did you mean somewhere else, @muloem ?

stellanl:
Ok, apparently we parse lots of redundant info in loadFUllQuestions(); we could use BulkDataService.fetchOptionNodes() to avoid any duplication.

janagombitova:
Test plan is missing, please make sure in the next changes there is one. I assume here the expectation is that exports are retrieved and built as they should. All export types work perfectly 👍

PR #2025, opened 2017-04-05
[#1736]Optimise fetching questions and options. (Connect #1736)

stellanl:
Questions were often refetched with full details just to get the option list
#### The solution
Use the minimal fetch of options from BulkDataServiceClient
## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


GAE/src/org/waterforpeople/mapping/dataexport/SurveySummaryExporter.java
```diff
@@ -213,17 +213,27 @@ protected SummaryModel buildDataModel(String surveyId, String serverBase, String
...
+     * @throws Exception
+     */
     protected Map<QuestionGroupDto, List<QuestionDto>> loadAllQuestions(
             String surveyId, boolean performRollups, String serverBase, String apiKey)
             throws Exception {
         Map<QuestionGroupDto, List<QuestionDto>> questionMap = new HashMap<QuestionGroupDto, List<QuestionDto>>();
+        //TODO: fetching in nested loops is inefficient, but we need the ordering of groups and questions in them 
         orderedGroupList = fetchQuestionGroups(serverBase, surveyId, apiKey);
         rollupOrder = new ArrayList<QuestionDto>();
         for (QuestionGroupDto group : orderedGroupList) {
```
muloem: Even though we need the question order, its still probably better to fetch the questions themselves in one go and then do the local processing to reorder them based on their question group.

muloem:
Oh and resolving the conflicts 👍

GAE/src/org/waterforpeople/mapping/app/web/SurveyRestServlet.java
```diff
@@ -393,6 +395,32 @@ private SurveyInstanceDto findSurveyInstance(Long surveyInstanceId) {
...
+     * @param surveyId
+     * @return
+     */
+    private List<QuestionDto> listSurveyQuestions(Long surveyId) {
+        List<Question> questions = qDao.listQuestionsBySurvey(surveyId); //useless ordering
+        List<QuestionDto> dtoList = new ArrayList<QuestionDto>();
+        if (questions != null) {
+            for (Question q : questions) {
+                QuestionDto dto = new QuestionDto();
+                DtoMarshaller.copyToDto(q, dto);
```
muloem: As @valllllll2000 mentioned in one comment, we should already replace this with springs `BeanUtils.copyProperties()`. Be sure to convert the `key().Id()` to `keyId()`

GAE/src/org/waterforpeople/mapping/app/web/SurveyRestServlet.java
```diff
@@ -393,6 +395,32 @@ private SurveyInstanceDto findSurveyInstance(Long surveyInstanceId) {
...
+                if (q.getType().equals(Question.Type.CASCADE) && q.getCascadeResourceId() != null) {
```
muloem: Maybe we can take this part out of here as its not to do with listing questions.  Maybe create a separate function and pass it the list of questions and then do this.

GAE/src/org/waterforpeople/mapping/app/web/SurveyRestServlet.java
```diff
@@ -409,6 +437,28 @@ private SurveyInstanceDto findSurveyInstance(Long surveyInstanceId) {
...
+    private List<QuestionOptionDto> listSurveyQuestionOptions(Long surveyId) {
+
+        List<QuestionOptionDto> dtoList = new ArrayList<>();
+        List<Question> questions = qDao.listQuestionByType(surveyId, Type.OPTION);
+        for (Question question : questions) {
+            List<QuestionOption> options = qoDao.listByQuestionId(question.getKey().getId());
+            if (options != null) {
+                for (QuestionOption option : options) {
+                    QuestionOptionDto dto = new QuestionOptionDto();
+                    DtoMarshaller.copyToDto(option, dto);
```
muloem: Same comment regarding the use of `BeanUtils.copyProperties()`

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -330,38 +330,40 @@
...
         boolean useQuestionId = "true".equals(options.get("useQuestionId"));
         String from = options.get("from");
         String to = options.get("to");
         String limit = options.get("maxDataReportRows");
         try {
-            Map<QuestionGroupDto, List<QuestionDto>> questionMap = loadAllQuestions(surveyId,
-                    performGeoRollup, serverBase, criteria.get("apiKey"));
-            if (questionMap != null) {
+            Map<QuestionGroupDto, List<QuestionDto>> questionMap = 
+                    loadAllQuestions(surveyId, performGeoRollup, serverBase, apiKey);
```
muloem: Could we already pass the `useQuestionId` parameter here and if its `true` then we already have the question options loaded in this call?

GAE/src/org/waterforpeople/mapping/dataexport/SurveySummaryExporter.java
```diff
@@ -290,8 +291,8 @@ private void writeHeader(PrintWriter pw, boolean isRolledUp) {
         return dtoList;
     }
 
-    protected List<QuestionDto> parseQuestions(String response)
-            throws Exception {
+    
+    protected List<QuestionDto> parseQuestions(String response) throws Exception {
```
muloem: I think we should replace this parsing with the simpler Jackson deserialiser, like in https://github.com/akvo/akvo-flow/blob/fb7a5b7928047033dc8780a31abd025441076f69/GAE/src/org/waterforpeople/mapping/dataexport/service/BulkDataServiceClient.java#L1396-L1401

That way we don't have to keep adding properties later on if new ones appear.

GAE/src/org/waterforpeople/mapping/dataexport/SurveySummaryExporter.java
```diff
@@ -352,50 +362,6 @@ protected JSONArray getJsonArray(String response) throws Exception {
         return null;
     }
 
-    protected class InputDialog extends JDialog implements ActionListener {
```
muloem: Snip snip! 👍

GAE/src/org/waterforpeople/mapping/app/web/SurveyRestServlet.java
valllllll2000: I propose I replace it once I finish the Mappers if that is OK with you all.

GAE/src/org/waterforpeople/mapping/dataexport/SurveySummaryExporter.java
```diff
@@ -234,17 +230,58 @@ protected SummaryModel buildDataModel(String surveyId, String serverBase, String
             }
             questionMap.put(group, questions);
         }
+        
+        if (loadOptionNames) {
```
muloem: Given that this function doesn't take any extra parameters that are not present at the level of `loadAllQuestions()` I would suggest we extract it from here and avoid having to add an extra parameter for `loadAllQuestions()` i.e. 

```
export () {
   loadAllQuestions();
   if(...) { loadQuestionOptions(); }
}
```

GAE/src/org/waterforpeople/mapping/app/web/SurveyRestServlet.java
```diff
@@ -390,6 +390,27 @@ private SurveyInstanceDto findSurveyInstance(Long surveyInstanceId) {
...
+    }
+    
+    /**
+     * lists all questions for a given questionGroup
+     *
+     * @param groupId
+     * @return
+     */
+    private List<QuestionDto> listQuestions(Long groupId) {
+        return listQuestionsWithLevelNames(qDao.listQuestionsByQuestionGroup(groupId, false).values());
```
muloem: When I looked at  this, I felt that the nesting intuitively would be the other way around i.e. 

```
listGroupQuestionsWithLevelName() {
 listQuestions();
addLevelNames();
}
```

GAE/src/org/waterforpeople/mapping/app/web/SurveyRestServlet.java
```diff
@@ -390,6 +390,27 @@ private SurveyInstanceDto findSurveyInstance(Long surveyInstanceId) {
...
+    }
+
+    /**
+     * lists all questions for a given survey
+     *
+     * @param surveyId
+     * @return
+     */
+    private List<QuestionDto> listSurveyQuestions(Long surveyId) {
+        return listQuestionsWithLevelNames(qDao.listQuestionsBySurvey(surveyId)); //useless ordering
```
muloem: Same as above with the nesting of `listQuestions()`

GAE/src/org/waterforpeople/mapping/dataexport/SurveySummaryExporter.java
```diff
@@ -290,54 +337,27 @@ private void writeHeader(PrintWriter pw, boolean isRolledUp) {
...
-                }
-            }
-        }
+
+        final JsonNode questionListNode =
+                JSON_RESPONSE_PARSER.readTree(response).get("dtoList");
+        final List<QuestionDto> qList = JSON_RESPONSE_PARSER.readValue(
+                questionListNode, new TypeReference<List<QuestionDto>>() {
+                });
+        dtoList.addAll(qList); //Why?
```
muloem: Returning `qList`here should be enough, unless `qList` is null

Issue #2017, opened 2017-03-24
`Next` and `Previous` links are strangely aligned


muloem:
Under the inspect data tab

<img width="271" alt="screen shot 2017-03-24 at 12 46 22" src="https://cloud.githubusercontent.com/assets/533336/24292984/e36d329e-108f-11e7-9f16-403e404eb538.png">

janagombitova:
👍 solved and ready to be released

PR #2026, opened 2017-04-06
[#2017] Fixed previous / next alignment (connect #2017)

loicsans:


loicsans:
This should look like this: 
screen shot 2017-04-06 at 12 31 58

Issue #1954, opened 2017-02-21
Wrapping of survey/folder names when moving or copying


janagombitova:
When the user moves or copies a survey or a folder, the action is indicated in the top right corner of the screen. Surveys and folders that have a longer name expand the area where the moving is indicated but this creates a lot of white space on overlays over the list of surveys, leaving the user with not a fully visible list to work with.

<img width="1261" alt="screen shot 2017-02-21 at 13 57 03" src="https://cloud.githubusercontent.com/assets/12456965/23165702/c0f6c0be-f83d-11e6-9f28-f0028b47566a.png">

@loicsans can you tackle this one as part of the layout work as well?

janagombitova:
This is what happens if you also have the long name in the breadcrumb 

<img width="1259" alt="screen shot 2017-02-21 at 13 59 47" src="https://cloud.githubusercontent.com/assets/12456965/23165787/14a112be-f83e-11e6-8319-83469f92f436.png">

janagombitova:
@loicsans this is how it looks like on UAT1 currently. Can you please fix this? Can you also make it that the entire survey + title + move and cancel buttons are all in one line? Thanks 
<img width="1269" alt="screen shot 2017-04-10 at 13 47 40" src="https://cloud.githubusercontent.com/assets/12456965/24860388/a57670f0-1df4-11e7-98d9-ac43463d695d.png">

PR #2027, opened 2017-04-06
[#1954] wrapping name (connect #1954)

loicsans:


loicsans:
This is the fix:
screen shot 2017-04-06 at 12 55 37

Issue #1972, opened 2017-03-03
Fix the header in Messages tab table


janagombitova:
Currently the entire Messages page is scrollable and if the user looks at things that happen in the bottom of the table she looses sight of the headers and then it is hard to orientate what she is looking at. Would it be possible to fix the header of the table? 

![5](https://cloud.githubusercontent.com/assets/12456965/23557705/338a8502-0031-11e7-9121-48ac30bbe70a.gif)

janagombitova:
@loicsans on the changes to the header, I think they are great. I have a few questions though? Is the column width now fixed? If yes, can we make it that the Form ID column is not on two lines? And can you enlarge the Survey column (perhaps to even up with the Message column)? The reason is that sometimes the survey name can be long and it needs some space 

<img width="1260" alt="screen shot 2017-04-07 at 12 40 16" src="https://cloud.githubusercontent.com/assets/12456965/24796992/b0ed04d8-1b8f-11e7-8b80-34e728f54c99.png">

iperdomo:
Make sure you  test the changes on lower resolutions on Windows (and/or Linux), since Mac hides scrollbars _automagically_.
![multiple-scrollbars](https://cloud.githubusercontent.com/assets/178474/24797777/19bbcce4-1b93-11e7-9175-b5bd3fd6a69f.png)

janagombitova:
@iperdomo yes will do, and we have a separate issue to handle the multiple scroll bars. Thanks for pointing it out

janagombitova:
@loicsans here is an example of how the items in the Survey column can get long and currently the column width adjusts, but perhaps we could fix the column width so you do not end up with this type of view of the Messages page

The longest items are in the Survey and Message columns

<img width="1263" alt="screen shot 2017-04-11 at 14 18 38" src="https://cloud.githubusercontent.com/assets/12456965/24908847/090f27e2-1ec2-11e7-9f11-80127ad4411d.png">

loicsans:
![apr-26-2017 14-48-05](https://cloud.githubusercontent.com/assets/1717219/25433010/84449db0-2a8f-11e7-8ed6-7f14572c1f13.gif)

stellanl:
I probably should not mention this, but in Firefox, the header still scrolls away (like I want it to)...

janagombitova:
And besides Stellan's point, now there is this odd white filed on the bottom of the page 
[...]

PR #2028, opened 2017-04-06
[#1972] table header in messages tab gest fixed while scrolling (connect #1972)

loicsans:


loicsans:
It behaves like this: 
apr-06-2017 14-13-10

loicsans:
The cells are now displaying the content properly.
apr-06-2017 15-55-32

janagombitova:
@loicsans would it be possible to make the Date and Form ID columns a bit narrower to create more space for the Survey column, as this one can hold quite a long string of texts?

loicsans:
apr-26-2017 14-48-05

Issue #1442, opened 2015-10-15
Performance improvements loading survey questions and question options (estimated: 8)


muloem:
While loading surveys we make a request to the `GET /questions?surveyId=` end point. Loading surveys that contain lots of `OPTION` type questions results in a timeout due to the deadline exceeded exceptions from GAE.  There are some straight forward improvements that can be made in the retrieval of questions and their associated components to improve on the current performance.

* [x] Retrieve the list of questions by querying based on the `surveyId` property of the question rather than querying the list of questions per question group (we have to ensure all questions have the `surveyId` property set)
* [x] Skip querying for translation objects of `QuestionOption` entities when making `GET /questions?surveyId=` requests
* [x] Eliminate the redundant `DtoMarshaller.copyToDto()` and `BeanUtils.copyProperties()` calls.  We can instead wrap the domain object with the DTO

janagombitova:
**Decision from Planning poker session:**

For this particular issue we will only invest one day of work to tackle small fixes that will improve the loading. Larger tasks as e.x. tackling translations, will be handled in separate issues.

janagombitova:
Loading of questions for larger surveys has became a growing issue for more dashboards in the last months. We have solved this issue temporarily by re-deploying the instances with a higher instance class. A more long term solution is needed.

@muloem will this issue already cover one of the problems or do we need to create more issues to tackle this problem? 
- issue that covers loading of translations need to be created

stellanl:
Re the first checkbox, I have looked in the datastore of a dozen of our biggest instances and not found a single question without a surveyId.

valllllll2000:
for the first checkbox, it is possible to query by surveyId and then order by questionGroupId column as long as the question group are in correct order. What would happen if the survey group is moved to another position?

valllllll2000:
As for the point 3, both DtoMarshaller.copyToDto() and BeanUtils.copyProperties() actually use BeanUtils.copyProperties(). According to this article:
http://web.archive.org/web/20160404130436/http://rodrigojramirez.com/2013/10/21/beware-of-copyproperties-or-the-30-milliseconds-tale/
it does seem like BeanUtils is very slow especialy when we have large collections. We can use BeanUtils from Spring framework which is faster or write our own Mapper to map the objects. There is also this tool http://orika-mapper.github.io/orika-docs/intro.html, not sure how well it would preform.

muloem:
[...]

PR #2039, opened 2017-04-13
Preformance improvements questions (connect #1442)

valllllll2000:
Slow loading of questions
#### The solution
This fix concerns only the part of Bean mapping between Question and QuestionDto and also mapping of QuestionOption
NA
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/com/gallatinsystems/survey/domain/Question.java
```diff
@@ -16,16 +16,11 @@
...
-import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
+import com.gallatinsystems.framework.domain.BaseDomain;
 
 import javax.jdo.annotations.NotPersistent;
 import javax.jdo.annotations.PersistenceCapable;
-
-import com.gallatinsystems.framework.domain.BaseDomain;
+import java.util.*;
```
iperdomo: This is a default setting on IntelliJ IDEA, after 5 imports of the same package, you get `.*` which not desirable.

valllllll2000: Fixed!

valllllll2000:
@stellanl before the rename, spring framework BeanUtils.copyProperties() just would not work at all even excluding the field "isName" it said the getters and setters did not match. I used intellij to generate getters and setters for isName and it created me those even though the previous ones seem to make more sense for human reading. The ideal solution would have been to rename the field to something less ambiguous but unfortunately due to GAE datastore it has to stay the same as the column name.

Issue #2002, opened 2017-03-15
Remove Export raw text file


janagombitova:
We currently support exporting data in different ways. The most used ones are raw data export and comprehensive export. We also support the raw text file export. This export type allows the users to get data out in a tsv file. 

Our users work with Flow collected data in Excel mostly, or in tools that allow import of xlsx or xls forms. The question is: Do we need to support raw text file export? Can we remove this functionality fully? 

@stellanl Is there a way to see if we have generated these raw text files recently and how many so we can evaluate the use of the feature and consider removing it.

stellanl:
Searching the services log files for RAW_DATA_TEXT should find any such reports generated. It will not show fetching of cached reports.
Last month, on test there were about a dozen generated (by me, jana, nischal*10)
On live, there were 9 (by lexy.rateringarntz@gmail.com and carlos.medina@charitywater.org *8)

janagombitova:
@stellanl thank you very much for that info. Would it be possible to get an overview of the last 6 months (per month) on live? 

I have reached out to both Lexy and Carlos to understand their use of the text file export to help us decide on the future of this feature

stellanl:
We only keep the last 30 days of log files for flowservices. GAE even less AFAIK

janagombitova:
@stellan noted. Will wait for the reply from our partners before we see how to move forward

janagombitova:
Considering these reasons: 
1. only two partners used this functionality in the last 30 days (Carlos replied that his reasons for using the text export was pure curiosity as he was playing around with the dashboard and Lexy did not reply) 
1. the feature is "broken" we only export the header structure and no data
1. users can export raw data and then save it as a txt file 

We **can remove** this feature from Flow.

stellanl:
Test plan: In the dashboard, go to the Reports->Export Reports tab. There should be no section or button titled "Export raw text file".

janagombitova:
All good and documentation updated 👍

stellan:
Hi!
Just as you know, you´ve been tagging the wrong Stellan in some occasions in this repo. :-)

BR //Stellan R

janagombitova:
Hi @stellan 

My apologies. Hope we have not been bothering you too much. Will do our best to not have that happening again

Cheers 
Jana

PR #2050, opened 2017-04-20
Issue/2002 remove raw textfile report (Connect #2002)

stellanl:
Raw text report was almost never used
#### The solution
Removed it from dashboard and flow-services factory class.
We should make separate issue for cleaning out the legacy browser applet code, which should make the raw data exporter itself unreferenced and deletable.
## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


muloem:
:+1: code reviewed!

Issue #2088, opened 2017-05-05
Performance improvement for question creation


muloem:
When creating a new question, the saving spinner takes a while before it disappears. We need to improve on the performance of the question creation.

* [x] Replace the `DtoMarshaller` for questions and the returned question options

PR #2094, opened 2017-05-08
[#2088]Use beanutils copying everywhere. (Connect #2088)

stellanl:
Marshalled Question data the old, slow way
#### The solution
Switch to the faster way from beanutils
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
+++ b/GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
@@ -51,6 +51,7 @@ import org.waterforpeople.mapping.dao.QuestionAnswerStoreDao;
 import javax.inject.Inject;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -422,8 +423,8 @@ public class QuestionRestService {
     }
 
     private List<QuestionOptionDto> attachAnyOptions(QuestionDto dto) {
-        ArrayList<QuestionOptionDto> qoResults = new ArrayList<QuestionOptionDto>();
         if (dto.getType() == QuestionDto.QuestionType.OPTION) {
+            ArrayList<QuestionOptionDto> qoResults = new ArrayList<QuestionOptionDto>();
             // since we do not need translations:
             List<QuestionOption> qoList = questionOptionDao.listByQuestionId(dto.getKeyId());
             List<Long> qoIdList = new ArrayList<Long>();
@@ -434,8 +435,9 @@ public class QuestionRestService {
                 qoResults.add(qoDto);
             }
             dto.setQuestionOptions(qoIdList);
+            return qoResults;
         }
-        return qoResults;
+        return Collections.emptyList();
     }
 
 }
--- a/GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
+++ b/GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
@@ -205,7 +205,7 @@ public class QuestionRestService {
         if (q != null) {
             QuestionDtoMapper qdm = new QuestionDtoMapper();
             dto = qdm.transform(q);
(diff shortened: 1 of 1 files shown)


GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
```diff
@@ -204,8 +203,8 @@ private static void setQuestionOptionsIdList(List<QuestionDto> questionDtoList,
         Question q = questionDao.getByKey(id);
         QuestionDto dto = null;
         if (q != null) {
-            dto = new QuestionDto();
-            DtoMarshaller.copyToDto(q, dto);
+            QuestionDtoMapper qdm = new QuestionDtoMapper();
+            dto = qdm.transform(q);
             if (q.getType() == Question.Type.OPTION) {
```
muloem: In this `if () {}` block we copy question options as well.. maybe you can replace the direct use of the spring library with our `QuestionOptionDtoMapper` as well?

GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
```diff
@@ -204,8 +203,8 @@ private static void setQuestionOptionsIdList(List<QuestionDto> questionDtoList,
...
                 Map<Integer, QuestionOption> qoMap = questionOptionDao.listOptionByQuestion(dto
```
muloem: I think here we should also consider using the `questionOptionDao.listByQuestionId()` so that we do not even bother with querying for translations.  We do not need for this case

GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
```diff
@@ -350,8 +349,8 @@ private static void setQuestionOptionsIdList(List<QuestionDto> questionDtoList,
             } else {
                 q = copyQuestion(questionDto);
             }
-            dto = new QuestionDto();
-            DtoMarshaller.copyToDto(q, dto);
+            QuestionDtoMapper qdm = new QuestionDtoMapper();
+            dto = qdm.transform(q);
             statusDto.setStatus("ok");
             statusDto.setMessage("");
```
muloem: infact line 357 - 371 is copied and pasted in all the functions. I suggest extracting into a single function and reuse

GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
```diff
@@ -448,4 +420,21 @@ private Question newQuestion(QuestionDto dto) {
         final Question result = questionDao.save(q);
         return result;
     }
+
+    private void attachAnyOptions(QuestionDto dto, List<QuestionOptionDto> qoResults) {
```
muloem: As a proposal, since the `qoResults` is never used locally besides adding elements to the list, perhaps we could eliminate it from the parameter list and instead have a return type `List<QuestionOptionDto>`

GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
```diff
@@ -448,4 +420,21 @@ private Question newQuestion(QuestionDto dto) {
...
+        if (dto.getType() == QuestionDto.QuestionType.OPTION) {
```
muloem: Remember to use `.equals()` for these comparisons unless they are primitives.

stellanl: I did. It's an enum, not a String.

GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionRestService.java
stellanl: Slightly less efficient, since we create & discard a List for every type except option, but ok.

Issue #2089, opened 2017-05-05
Performance improvement for question group creation


muloem:
Issue to tackle to improve on the question group creation performance
* [x] Replace the serialisation / deserialisation utility `DtoMarshaller`

PR #2095, opened 2017-05-09
Issue/2089 speed up questiongroup creation (Connect #2089)

stellanl:
Uses slow DtoMarshaller code.
desc/description fields are not used.
#### The solution
Switch to beanutils copying.
Remove the field.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionGroupRestService.java
+++ b/GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionGroupRestService.java
@@ -23,6 +23,7 @@ import java.util.Map;
 
 import javax.inject.Inject;
 
+import org.akvo.flow.domain.mapper.QuestionGroupDtoMapper;
 import org.apache.commons.lang.ArrayUtils;
 import org.springframework.beans.BeanUtils;
 import org.springframework.stereotype.Controller;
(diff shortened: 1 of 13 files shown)


GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionGroupRestService.java
```diff
@@ -33,7 +33,7 @@
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseBody;
 import org.waterforpeople.mapping.app.gwt.client.survey.QuestionGroupDto;
-import org.waterforpeople.mapping.app.util.DtoMarshaller;
+import org.waterforpeople.mapping.app.gwt.client.survey.QuestionGroupDtoMapper;
```
muloem: I just noticed the package of this as well as the `QuestionDtoMapper` and I think we should put it somewhere else.... e.g. `org.akvo.flow.domain.mapper.`  please propose some other one but I think it should not be in the `...gwt.client..`  because these are packages we have been slowly phasing out as they were part of the google web toolkit

GAE/src/org/waterforpeople/mapping/app/web/rest/QuestionGroupRestService.java
```diff
@@ -173,11 +156,7 @@
         QuestionGroup s = questionGroupDao.getByKey(id);
         QuestionGroupDto dto = null;
         if (s != null) {
-            dto = new QuestionGroupDto();
-            DtoMarshaller.copyToDto(s, dto);
-            // needed because of different names for description in
-            // questionGroup and questionGroupDto
-            dto.setDescription(s.getDesc());
+            dto = QuestionGroupDtoMapper.transform(s);
```
muloem: I was a little thrown off when I saw the variable name `s`. I thought you were using the function to transform a survey.  May be an idea to rename it appropriately.

Issue #2116, opened 2017-05-30
Deletions fail because of 'transient instances'


stellanl:
The problem in #857 is still around. In the log for akvoflow-149:

```2017-05-27 BST
14:03:56.877
org.waterforpeople.mapping.app.web.SurveyTaskServlet handleRequest: action: deleteQuestion id: 2440145 (SurveyTaskServlet.java:73)
14:03:57.515
com.gallatinsystems.framework.rest.AbstractRestApiServlet executeRequest: Could not execute rest request (AbstractRestApiServlet.java:95)
javax.jdo.JDOUserException: Transient instances cant be deleted.
	at org.datanucleus.jdo.NucleusJDOHelper.getJDOExceptionForNucleusException(NucleusJDOHelper.java:375)
	at org.datanucleus.jdo.JDOPersistenceManager.jdoDeletePersistent(JDOPersistenceManager.java:759)
	at org.datanucleus.jdo.JDOPersistenceManager.deletePersistent(JDOPersistenceManager.java:772)
	at com.gallatinsystems.framework.dao.BaseDAO.delete(BaseDAO.java:566)
	at com.gallatinsystems.survey.dao.QuestionDao.delete(QuestionDao.java:165)
	at com.gallatinsystems.survey.dao.QuestionDao.delete(QuestionDao.java:127)
	at org.waterforpeople.mapping.app.web.SurveyTaskServlet.handleRequest(SurveyTaskServlet.java:92)
	at com.gallatinsystems.framework.rest.AbstractRestApiServlet.executeRequest(AbstractRestApiServlet.java:85)
	at com.gallatinsystems.framework.rest.AbstractRestApiServlet.doPost(AbstractRestApiServlet.java:61)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:637)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:717)
	at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:511)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1166)
	at com.gallatinsystems.framework.servlet.PersistenceFilter.doFilter(PersistenceFilter.java:71)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1157)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:330)
[...]

PR #2145, opened 2017-06-07
Issue/2116 questions are never deleted (Connect #2116)

stellanl:
Deletion of questions fails silently, they remain in the datastore.
#### The solution
Bypass cache when fetching the question to delete. To avoid a problem when the just-deleted q would be fetched during the resorting, move the deletion last.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


muloem:
Closing this one.  The changes are too close to the database level.  We need to address the issue of how caching happens that avoids the deletion.

stellanl:
Not quite sure what you mean. Since we do the caching "by hand", we need to consider the cache when deleting.

Issue #1779, opened 2016-10-05
Export crashes w NPE if option question has no options


stellanl:
Related to #362.
If an OptionQuestion has no defined options at the time of a data export (with QuestionIds enabled) this is not handled and an NPE is thrown at GraphicalSummaryExporter.createRawDataHeader (L1143).

I suggest outputting a header for the (empty) column that indicates the problem (it could be happening in several places), but we might return a helpful error message instead.

stellanl:
Test plan:
Create a question without options. Order a raw data report (importable format).
Report should be generated & notification email sent.

janagombitova:
After testing this seems to still not work, as I am not getting an export for a form where I created an option-less question

janagombitova:
👍  works

janagombitova:
Created via Reamaze:

Link: /admin/conversations/akvoflow-support-request-number-145364-from-geert-at-akvo-dot-org

PR #2148, opened 2017-06-07
[#1779]Survive optionless questions. (connect #1779)

stellanl:
Questionless questions made report generator crash.
#### The solution
Test for it.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
@@ -1377,37 +1377,37 @@ public class GraphicalSurveySummaryExporter extends SurveySummaryExporter {
 
                             createCell(row, offset++, header, headerStyle);
 
-                            // check if we need to create columns for all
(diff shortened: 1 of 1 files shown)


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -1377,37 +1377,37 @@ private String sanitize(String s) {
...
-                                            + header + qo.getText(),
-                                            headerStyle);
-                                }
-
-                                // add 'other' column if needed
-                                if (q.getAllowOtherFlag()) {
-                                    createCell(row, offset++, "--OTHER--",
-                                            headerStyle);
+                                OptionContainerDto ocDto = q.getOptionContainerDto();
+                                if (ocDto != null) { //used to be legal
```
muloem: Just as an extra check I would suggest we check that the list has items and not just that it is `null`

Issue #1688, opened 2016-06-06
Serving static files from the backend - estimated: 6


muloem:
In a number of cases, we have images and some JS library files that do not change frequently.  As a way to improve the performance of app loading, we should modify the deployment to take into account some of these files and take advantage of the current (and possibly future) platform's capabilities around serving static files.

Where possible  we can serve the JS libraries from an existing CDN.

janagombitova:
**From planning poker meeting:** This issue means identifying the files. Questions as If you have changed the file how do you keep track of it and how to test this? came up in the discussion

kymni:
## Test plan
* Open browser's console to inspect loaded libraries
* Confirm that the following libraries now load from CDNs instead of from Flow
  * jQuery
  * jQuery-UI
  * handlebars.js
  * ember
  * jQuery.dataTables
  * d3
  * resumable
  * mapbox/cartodb
  * iframeResizer
* Same applies for public interface

janagombitova:
@stellanl could you check  this fix on uat1 for me? (so we get an unbiased test run)

stellanl:
Admin: check.
Public: check.

PR #2157, opened 2017-06-19
Serve js files from available CDNs (connect #1688)

kymni:
Some JavaScript files were served from Flow
#### The solution
Changed the source of files to available CDNs

## Checklist
* [x] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation

### DIFF
--- a/Dashboard/app/dashboard.html
+++ b/Dashboard/app/dashboard.html
@@ -24,13 +24,13 @@
     <script>
         window.jQuery || document.write("<script src='/vendorjs/js/vendor/jquery-1.8.2.min.js'><\/script>")
     </script>
-    <script src="/vendorjs/js/vendor/handlebars-1.0.rc.1.js"></script>
-    <script src="/vendorjs/js/vendor/ember-1.0.0.pre-2-36.min.js"></script>
+    <script src="//cdnjs.cloudflare.com/ajax/libs/handlebars.js/1.0.rc.1/handlebars.min.js"></script>
+    <script src="//cdnjs.cloudflare.com/ajax/libs/ember.js/1.0.0-pre.2/ember-1.0.0-pre.2.min.js"></script>
     <script src="/vendorjs/js/vendor/ember-data-rev10.min.js"></script>
     <script src="/vendorjs/js/vendor/jquery-ui-1.8.21.custom.js"></script>
-    <script src="/vendorjs/js/vendor/jquery.dataTables.js"></script>
-    <script src="/vendorjs/js/vendor/d3.v2.min.js"></script>
-    <script src="/vendorjs/js/vendor/resumable.min.js"></script>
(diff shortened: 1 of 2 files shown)


Dashboard/app/dashboard.html
```diff
@@ -24,13 +24,13 @@
     <script>
         window.jQuery || document.write("<script src='/vendorjs/js/vendor/jquery-1.8.2.min.js'><\/script>")
     </script>
```
muloem: I think we can remove the above lines and also load jquery from cloudflare.

Dashboard/app/dashboard.html
```diff
@@ -24,13 +24,13 @@
...
-    <script src="/vendorjs/js/vendor/handlebars-1.0.rc.1.js"></script>
-    <script src="/vendorjs/js/vendor/ember-1.0.0.pre-2-36.min.js"></script>
+    <script src="//cdnjs.cloudflare.com/ajax/libs/handlebars.js/1.0.rc.1/handlebars.min.js"></script>
+    <script src="//cdnjs.cloudflare.com/ajax/libs/ember.js/1.0.0-pre.2/ember-1.0.0-pre.2.min.js"></script>
     <script src="/vendorjs/js/vendor/ember-data-rev10.min.js"></script>
     <script src="/vendorjs/js/vendor/jquery-ui-1.8.21.custom.js"></script>
```
muloem: Isn't there a CDN for this one?

kymni: This version of ember-data seems to precede even the earliest tagged version so couldn't find it on any CDN.

iperdomo:
If the resource is available over https, we should always use the https version. 

* https://www.paulirish.com/2010/the-protocol-relative-url/

We can also add `integrity` to those files.

* https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
* http://caniuse.com/#feat=subresource-integrity

Dashboard/app/dashboard.html
```diff
@@ -20,14 +20,11 @@
 <body class="appPage">
 
     <!-- Grab Google CDN's jQuery. fall back to local if necessary -->
```
muloem: We don't need this comment either.

Dashboard/app/public.html
```diff
@@ -28,14 +28,13 @@
   <body class="publicPage">
 
   <!-- Grab from a CDN, fall back to local if necessary -->
```
muloem: Ditto for this comment

kymni:
Regarding loading libraries via HTTPS, I've not specified the protocol in the src so the libraries will always be pulled using the same protocol as the dashboard.

As for d3, suggested fix is tied to PR #1953

muloem:
> Regarding loading libraries via HTTPS, I've not specified the protocol in the src so the libraries will always be pulled using the same protocol as the dashboard.

@kymni true. I think the idea is to explicitly specify HTTPS.

Issue #1762, opened 2016-09-21
Data point name not shown in monitored form export


janagombitova:
There is a case where a few data points do not show the data point name in the raw data export for a monitored form. 

Here are some more details: 
- all point names are properly shown in the registration form export
- all names are shown after syncing on the app 
- after opening a point on the app the monitored form is properly linked via the point ID
- after re-downloading the export again, the names are still missing in the column

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/first-mile-plot-datas-2

janagombitova:
Created via Reamaze:

Link: https://akvoo.reamaze.com/admin/conversations/display-name-empty

janagombitova:
Hi both, 

I am missing a test plan here, can you please add it? 

On the current UAT1 build I am getting the point names for new surveys, but still not getting the point names on older surveys after a quick check.

janagombitova:
So, in the raw exports the name is now showing. 

I tried one more thing, changing the "setting" for the name, I have added another question (numeric) to build the point name and made sure all the submissions have the numbers. After syncing in the points on the app the name was built properly - with the extended values. But if I export the data for both the reg form or for the monitoring form, the changes are not shown in the data point name column. 

See the screen shot from the app and the example export: 

[Example - point name.xlsx](https://github.com/akvo/akvo-flow/files/1084754/Example.-.point.name.xlsx)

![screenshot_2017-06-19-13-31-46](https://user-images.githubusercontent.com/12456965/27283477-ca9c0804-54f4-11e7-95f6-064eb9b55553.png)

stellanl:
AFAIK the changes are only applied on data submission/import. Nothing loops over the existing entities in the db just because you publish a new version of the survey.

janagombitova:
Before when I made changes to the survey and synced the points to the phone and then exported the data, the names were updated as well in the exports. This is different now.

stellanl:
Ok. Digging deeper. Which instance is this on? uat1?

janagombitova:
UAT1 > 1.9.19 folder, there is one survey there. I am testing the same on pilots.akvoflow.org to see if it still works there. Will let you know

stellanl:
Just looked in the db. Same as in the report.

janagombitova:
Hmm.. So in the app and in Monitoring tab we have the "updated" data point name, but in the db and reports not...

stellanl:
[...]

PR #2158, opened 2017-06-19
[#1762]Make script to fix up existing names

stellanl:
Lots of existing SurveyInstance entities with no or wrong surveyedLocaleDisplayName.
#### The solution
Reset any that are empty or wrong from the pointed-to SurveyedLocale
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


stellanl:
This fix and the one for #2160 should keep it from cropping up again.

Issue #2160, opened 2017-06-20
Raw data exports do not reflect up-to-date data point names 


janagombitova:
The user defines what builds the data point name. However many times the users realise after a while that other information is more helpful to have in the name to identify data points. So they change the data point name definition. They make another question to build the name,  data clean the answers, remove an existing one or fully add a new question (can do anything). After syncing the data points in the app the changes are shown and before today (approximate day) this change was also reflected in the data exports. Not anymore. 

Here are the steps the user takes and what she sees: 

1. User creates a form and defines what builds the data point name (free text question) 
1. User captures data 
1. User exports data (registration form) and the name is shown correctly 
1. User makes a change to what builds the data point name (added the number question as well) 
1. User collects new data on the app and the name is shown and built correctly
1. User syncs the data on the app and **all** the data points now show with the updated version of the name (text+number)
1. User goes into the browser, on the monitoring tab the data point names show updated 
1. User exports data (registration form) and sees the names show their first version, so not the updated one after syncing (old points that were created with the first definition of the name have the first version of the name, not the 2nd updated one even though on the app in the list of points they show updated.  

Example: 

Screen shot of the updated data point name in the app and from the export. 

![screenshot_2017-06-20-13-38-23](https://user-images.githubusercontent.com/12456965/27331468-dd300fb4-55bd-11e7-8ac6-6dbff8a7c4a9.png)

<img width="637" alt="screen shot 2017-06-20 at 13 37 13" src="https://user-images.githubusercontent.com/12456965/27331416-9f608cb8-55bd-11e7-8a3a-470b8c2d2fdd.png">

janagombitova:
Here is an issue and a fix where we know for sure that it was still working: https://github.com/akvo/akvo-flow/issues/1934

stellanl:
In the database, we see that the new names are set for the SurveyedLocation entities but not for the old surveyInstance entities.

stellanl:
One solution is to make the exporter get the name from the SurveyedLocation.

stellanl:
Test plan:
* Change the name-generating questions for a form with data.
* Publish the form.
* Wait a few minutes for the background point renaming to 
complete.
* Make a raw data report; the name column should have updated.

janagombitova:
@stellanl will this issue https://github.com/akvo/akvo-flow/issues/1743 we also resolved with the fixes here?

stellanl:
[...]

PR #2161, opened 2017-06-21
[#2160]Change names in SurveyInstance entities, too. (Connect #2160)

stellanl:
Only the names for SurveyedLocale entities were updated when a a survey was published, not the SurveyInstance names.
#### The solution
Change them too.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


muloem:
We will close this PR as we have decided to go for an alternative solution that does not add to the reads and writes to and from the datastore.

Issue #2022, opened 2017-04-04
Tweak spring framework configuration to improve on performance of backend


muloem:
* [x] Define controller beans in rest-servlet.xml beans definition
* [ ] ~Set controller beans as singletons~
* [ ] Define controller beans dependencies (avoid auto wiring)
* [ ] ~Define DAO beans as singletons~
* [ ] what to do with static resources like `/favicon.ico`?

muloem:
# Test Plan

Regression testing should catch any issues caused by these changes.

janagombitova:
Nothing found in the regression round. We are good to go 👍

PR #2181, opened 2017-07-07
Configure spring to initialize beans lazily (connect #2022)

muloem:
* Spring beans are created at application startup time with the consequence that the application startup time is longer. 

#### The solution
* We configure the beans to only be initialised when they are needed.

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


muloem:
Wrong commit issue number. I will close this and create one with a correct issue id!

PR #2189, opened 2017-07-12
Issue/2160 compose display name during report generation (Connect #2160)

stellanl:
Display names in the raw data report don't change when the selection of questions change.
#### The solution
Build the display name during report generation and ignore any stored value.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
@@ -731,7 +731,7 @@ public class GraphicalSurveySummaryExporter extends SurveySummaryExporter {
         StringBuilder dname = new StringBuilder();
         
         for (QuestionDto q : questionList) {
-            if (q.getKeyId() != null && q.getType() != null) {
+            if (q != null && q.getKeyId() != null && q.getType() != null) {
                 Map<Long, String> allAnswersForQuestion = instanceData.responseMap.get(q.getKeyId());
                 if (allAnswersForQuestion != null && allAnswersForQuestion.size() > 0) {
                     String qtype = q.getType().toString();
--- a/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
@@ -72,11 +72,13 @@ import org.waterforpeople.mapping.app.web.dto.SurveyRestRequest;
 import org.waterforpeople.mapping.dataexport.service.BulkDataServiceClient;
 import org.waterforpeople.mapping.domain.CaddisflyResource;
 import org.waterforpeople.mapping.domain.CaddisflyResult;
+import org.waterforpeople.mapping.domain.QuestionAnswerStore;
 import org.waterforpeople.mapping.domain.response.value.Media;
 import org.waterforpeople.mapping.serialization.response.MediaResponse;
 
 import com.gallatinsystems.common.util.JFreechartChartUtil;
 import com.gallatinsystems.survey.dao.CaddisflyResourceDao;
+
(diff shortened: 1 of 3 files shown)


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -716,6 +723,32 @@ private synchronized int writeInstanceData(Sheet sheet, final int startRow,
...
     }
+    
+    private String constructDisplayName(
+            List<QuestionDto> questionList,
+            InstanceData instanceData
+            ) {
+        StringBuilder dname = new StringBuilder();
+        
+        for (QuestionDto q : questionList) {
+            if (q != null && q.getKeyId() != null && q.getType() != null) {
```
muloem: Perhaps we can make these checks (`null`, `keyId`, `type`) at the point in time when we add the display name question to `displayNamePartList `.  It is something that only happens at the beginning of the report generation. That way we don't have an extra `if` block here

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -572,7 +578,8 @@ private synchronized int writeInstanceData(Sheet sheet, final int startRow,
             InstanceData instanceData, boolean generateSummary,
             List<String> questionIdList, List<String> unsummarizable,
             Map<String, String> nameToIdMap, Map<String, String> collapseIdMap,
-            SummaryModel model, boolean useQuestionId)
+            SummaryModel model, boolean useQuestionId, 
+            String displayName)
```
muloem: Rather than add an extra parameter here, I would be in favour of having `displayNamePartList` as a global since it does not change over the entire survey generation..  Otherwise it feels strange on line 554 to be passing in the `instanceData` twice. i.e. as one of parameters and in addition nested in the `constructDisplayName` call

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -716,6 +723,32 @@ private synchronized int writeInstanceData(Sheet sheet, final int startRow,
 
         return maxRow + 1;
     }
+    
+    private String constructDisplayName(
```
muloem: This function in essence does what `SurveyedLocale.assembleDisplayName()` does. Perhaps its worth making that one static and having the display name generation code in one place?

stellanl: But then we need a lot of conversion as it uses Question and QAS object lists, and here we have QuestionDto and String lists.

muloem: that's true. ok lets leave that as is. 👍

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -716,6 +721,25 @@ private synchronized int writeInstanceData(Sheet sheet, final int startRow,
...
     }
+    
+    private String constructDisplayName(InstanceData instanceData) {
+        StringBuilder dname = new StringBuilder();
+        
+        for (QuestionDto q : displayNamePartList) {
+            Map<Long, String> allAnswersForQuestion = instanceData.responseMap.get(q.getKeyId());
+            if (allAnswersForQuestion != null && allAnswersForQuestion.size() > 0) {
+                String qtype = q.getType().toString();
+                String answer = allAnswersForQuestion.get(Long.valueOf(0));
```
muloem: Maybe you can just use the literal `0L` instead of `Long.valueOf(0)`

Issue #2035, opened 2017-04-11
Improve loading time of device list


janagombitova:
I noticed on a partner dashboard that the Device list is loading all the devices ever connected to the dashboard (with some being last connected in 2013) 

<img width="1184" alt="screen shot 2017-04-11 at 16 20 56" src="https://cloud.githubusercontent.com/assets/12456965/24913898/e50c0ab6-1ed2-11e7-878d-5e172c466dca.png">

Considering the user can organise the Device list based on all the header items, I think we need to reconsider how many entries we show in the list. Secondly, are we showing only the last time device connected to the instance or are we showing every time a device connected to the instance?

stellanl:
We show only the last time it connected, but it can still be a long list with all the phones that were ever used with that instance.
I think there are a few ways to implement this:

1 By default show just the devices seen in the last week/month/year, Then there is a button to push to "Show all".
2 Show a first page of, say, 30 devices, and then have the classic "Next page", "Prev page" controls.
3 Show a first page of 30 devices, and add to the bottom of the list as the user scrolls down.

I don't (yet) know how hard it is to implement any of them.

loicsans:
There are several considerations on the subject (randomly in no particular order)
Should devices have a lapsing date? (ie. If a device has not connected to the instance for more than 6 months, it should automatically disappear from the list or get archived (somewhere) ).
This brings the question of device status: Active, inactive, deleted...
Maybe each devices should have a log of every time they connected but we should only show the last time it has connected in the table.

janagombitova:
@loicsans those are interesting points. But they complicate things. Currently once an app is installed on a device the device is in the list (user facing - not sure how we handle things internally). We cannot delete a device from the dashboard, or enforce it to not collect data. The interaction is super simplistic - these are the devices connected to this instance > put them in an assignment.  These (and more) questions should be considered once we get to improving device management, groups, etc. 

With this issue we are trying to solve the problem that the list takes up to 45s to load what is too long and once it loads it shows all the devices ever connected to the instance, organising the list by the most recently connected device. The user indeed does not need to see the entire thing.

Kiarii:
do we need the phone number col, how are these used? most appear 'encrypted'..

janagombitova:
[...]

PR #2197, opened 2017-07-18
Issue/2035 speed up device tab (Connect #2035)

stellanl:
Old, slow marshaling code
#### The solution
Beanutils
## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


GAE/src/org/waterforpeople/mapping/app/web/rest/SurveyAssignmentRestService.java
```diff
@@ -164,17 +162,20 @@
...
         return dto;
     }
 
     private SurveyAssignment marshallToDomain(SurveyAssignmentDto dto) {
+        //String[] handsOff = {"keyId", "key"};
         final SurveyAssignment sa = new SurveyAssignment();
 
-        DtoMarshaller.copyToCanonical(sa, dto);
+        BeanUtils.copyProperties(dto, sa);
+        sa.setKey(KeyFactory.createKey("SurveyAssignment", dto.getKeyId()));
```
muloem: Perhaps only set the key when `keyId != null`.

Issue #2199, opened 2017-07-18
Speed up device group list fetch


stellanl:
The third part of #2035.
/rest/device_groups is still using the DtoMarshaller method.
Switching to beanutils should speed it up.

stellanl:
Marshaling to dto is straightforward, but the beanutils-optimised to-domain code looks like this:
```
  // copy the properties, except the createdDateTime property,
  // because it is set in the Dao.
  BeanUtils.copyProperties(deviceGroupDto, s,
        new String[] {
         "createdDateTime"
        });
```
which makes me wonder if that should be the pattern for all entity creation.
It is not for (at least) device entites.

janagombitova:
Is this issue also merged to UAT1 and will be released with 1.9.21? If yes, I can confirm everything around the devices works as expected. 

This bug is still there https://github.com/akvo/akvo-flow/issues/402 but that is a separate issue

PR #2200, opened 2017-07-18
[#2199]Beanutils marshaling to dto. (Connect #2199)

stellanl:
DeviceGroup list fetch was slow
#### The solution
Use beanutils marshaling
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/waterforpeople/mapping/app/web/rest/DeviceGroupRestService.java
```diff
@@ -74,8 +74,7 @@
         DeviceGroup s = deviceGroupDao.getByKey(id);
```
muloem: Its a small detail but referring to DeviceGroup as `s`  (in the `POST` and `PUT` methods) threw me off a little. I was expecting a survey when I saw it in line `marshallToDto(s)`

Issue #2033, opened 2017-04-10
Multiple colour background in cascades 


janagombitova:
<img width="1253" alt="screen shot 2017-04-10 at 14 23 03" src="https://cloud.githubusercontent.com/assets/12456965/24861513/48333b1c-1df9-11e7-9c0a-58d355a096ba.png">

If you look at the Level part of the screen you can see that there are multiple layers of different shades of grey there. Can we simplify this?

janagombitova:
Hey @loicsans 

I have checked the changes on https://flowaglimmerofhope.appspot.com/admin/ and like that you chose for "less is more" strategy. It looks much better and more in line with the other changes introduced this year. 

This cleaner design however highlights some issues we have in the overall design (already) that I think we then need to address as well. I have created them in separate issues so we can tackle them independently in the next weeks. They are tagged as 1-Cascades. 

However, there are a few issues, that need to be resolved in this fix in order to release these changes. All relate to when importing a file to create a cascade
 
## Empty space under the import selection part and the progress bar

<img width="1280" alt="screen shot 2017-09-26 at 12 24 43" src="https://user-images.githubusercontent.com/12456965/30858363-a6bc6bd6-a2bf-11e7-904a-172c924ddc4d.png">

## Importing does not show you the file anymore
After the user selects the file to import, you do not see the file name anywhere anymore

![1](https://user-images.githubusercontent.com/12456965/30858282-4fe7a8a2-a2bf-11e7-9a7c-97dc2b279bfd.gif)

![screen shot 2017-09-26 at 13 38 26](https://user-images.githubusercontent.com/12456965/30858468-1cb02ac6-a2c0-11e7-84b1-cfe753fc3782.png)

## Text field not clear for input of number of level
When importing the cascade, you need to specify the number of levels the file contains to build the cascade. That field is not so visible anymore in the new design

<img width="1258" alt="screen shot 2017-09-26 at 13 40 44" src="https://user-images.githubusercontent.com/12456965/30858532-5d264fae-a2c0-11e7-8f02-76b87f745b3e.png">

## Align checkbox for codes with text
The checkbox the user has to select if the cascade also has codes, is in a second row in comparison to the text 
<img width="815" alt="screen shot 2017-09-26 at 13 44 10" src="https://user-images.githubusercontent.com/12456965/30858648-d47f5a96-a2c0-11e7-8fe7-7dd465906214.png">

## Description missing
[...]

PR #2227, opened 2017-08-24
Issue/2033 cascade (connect #2033)

loicsans:


loicsans:
screen shot 2017-08-25 at 15 03 40
The cascade looks cleaner

Issue #2144, opened 2017-06-07
Maps code refactor


kymni:
The current code that controls the maps needs to be refactored to align it with the rest of the framework. This means that parts of the code such as calls for survey groups, surveys, questions, and answers which are currently being made need to be replaced with standardized and reusable versions used in other parts of the dashboard

kymni:
Creating new branch to better structure the changes

kymni:
## Test plan
* Mapbox maps function as before
* Changes in Cartodb maps
  * Folder/Survey filters now load in same style and manner as rest of dashboard
  * Survey groups end point no longer triggers each time you open the maps tab
  * Answers endpoint now pulls responses from Flow storage instead of Carto

janagombitova:
Here are a few issues from the 1st test round: 

## 1) Loading spinner kept on spinning
![screen shot 2017-10-31 at 10 13 14](https://user-images.githubusercontent.com/12456965/32218229-918b2140-be2a-11e7-955d-2e196fc66599.png)

## 2) Sometimes data points did not render
![screen shot 2017-10-31 at 10 56 44](https://user-images.githubusercontent.com/12456965/32218246-9e8604fa-be2a-11e7-90a3-468ab0aa2ee4.png)

But once I reloaded the entire instance maps showed to points for the same selected form 

## 3) Forms remain on map also after selecting a new folder 
I selected a folder > survey > form and then the points loaded on the map. Then I selected another folder, but the initial points remained on the map until the point I selected the new form. I would expect that once I select a different survey or folder, that the points will disappear from the map

## 4) Missing data point name, Submitter 
When opening a point I only see Collected on. Previously we also showed the data point name, ID (I think) and submitter besides the questions and answers. These should be brought back

kymni:
Unable to replicate issues 1 & 2

> 3) Forms remain on map also after selecting a new folder 

We've reused the folder/survey selectors used in other parts of the dashboard so unable to bind the event listeners, which would allow us to clear the map as selection changes

> 4) Missing data point name, Submitter

The refactor reuses the end point that pulls point data from Flow instead of Cartodb. This end point does not currently provide these two properties. We've therefore replicated the details as displayed on Mapbox maps

janagombitova:
> 1) Loading spinner kept on spinning
> 2) Sometimes data points did not render

Could not reproduce either. Will test out more later to see if I can get it happen again.

> 3. Forms remain on map also after selecting a new folder
[...]

PR #2249, opened 2017-09-13
Maps code refactor (connect #2144)

kymni:
There was a lot of superfluous that could be replaced by already existing code
#### The solution
Replaced this code with reusable existing solutions
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/controllers/maps-controllers-common.js
+++ b/Dashboard/app/js/lib/controllers/maps-controllers-common.js
@@ -169,15 +169,10 @@ FLOW.mapsController = Ember.ArrayController.create({
         var namedMapObject = {};
         namedMapObject['mapName'] = 'raw_data_'+formId;
         namedMapObject['tableName'] = 'raw_data_'+formId;
-        namedMapObject['interactivity'] = ['lat','lon','data_point_id'];
+        namedMapObject['interactivity'] = ['lat','lon','id'];
         namedMapObject['query'] = 'SELECT * FROM raw_data_'+formId;
 
-        if (!Ember.none(this.questions)) {
-            this.questions.forEach(function (qItem) {
-                namedMapObject['interactivity'].push("q"+qItem.get('keyId'));
-            });
-            this.namedMapCheck(namedMapObject, formId);
-        }
+        this.namedMapCheck(namedMapObject, formId);
     },
 
     /*Check if a named map exists. If one exists, call function to overlay it
--- a/Dashboard/app/js/lib/controllers/maps-controllers-common.js
+++ b/Dashboard/app/js/lib/controllers/maps-controllers-common.js
@@ -277,150 +277,14 @@ FLOW.mapsController = Ember.ArrayController.create({
                 FLOW.mapsController.set('markerCoordinates', [data.lat, data.lon]);
                 
                 //get survey instance
-                self.set( 'selectedSI', FLOW.store.find(FLOW.SurveyInstance, data.id));
+                FLOW.placemarkDetailController.set( 'si', FLOW.store.find(FLOW.SurveyInstance, data.id));
 
                 //get questions answers for clicked survey instance
-                self.set('questionAnswers', FLOW.store.findQuery(FLOW.QuestionAnswer, {
+                FLOW.placemarkDetailController.set('content', FLOW.store.findQuery(FLOW.QuestionAnswer, {
                     'surveyInstanceId' : data.id
                 }));
             });
         });
-    },
-
-    loadQuestionAnswers: function(){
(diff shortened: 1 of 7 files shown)


Dashboard/app/js/lib/views/maps/map-views-common.js
```diff
@@ -180,307 +164,132 @@ FLOW.NavMapsView = FLOW.View.extend({
...
     If a placemark is selected and the details pane is hidden make sure to
     slide out
   */
   handlePlacemarkDetails: function () {
-    var details;
-
-    details = FLOW.placemarkDetailController.get('content');
+    var details = FLOW.placemarkDetailController.get('content');
 
     if (!this.detailsPaneVisible) {
```
muloem: you have this block `if (!this.detailsPane) {.... set to true}` which is repeated in a few places. I suggest to collapse them all under a single function `showDetailsPane()` attached to this view. That can make it more explicit what you are attempting to do with that block.

But It will be cleaner this time around because you are only using properties on the view rather than all the jQuery and HTML that was in the previous version.

Dashboard/app/js/lib/views/maps/map-views-common.js
```diff
@@ -130,23 +123,14 @@ FLOW.NavMapsView = FLOW.View.extend({
         }
```
muloem: Is the block above here `$.ajaxSetup()` still needed?

Dashboard/app/js/lib/views/maps/map-views-common.js
```diff
@@ -88,8 +80,9 @@ FLOW.NavMapsView = FLOW.View.extend({
       self.handleShowHideDetails();
     });
 
-    // Slide in detailspane after 1 sec
-    this.hideDetailsPane(1000);
+    self.set('detailsPaneVisible', false);
```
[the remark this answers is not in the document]
kymni: needed to trigger the property observer. The details pane is open by default when the view is inserted so needs to be closed

Dashboard/app/js/lib/views/maps/map-views-common.js
kymni: Carto maps still uses ajax calls to get and create named maps so this block is still needed to display the loading icon when the calls are active

Dashboard/app/js/lib/views/maps/map-views-common.js
```diff
@@ -887,80 +335,21 @@ FLOW.NavMapsView = FLOW.View.extend({
...
-  clearCartodbLayer: function(){
-    //check to confirm that there are no layers displayed on the map
-    if(this.layerExistsCheck){
-      this.map.removeLayer(this.cartodbLayer);
-      this.layerExistsCheck = false;
-    }
-  },
-
   clearMap: function() {
     var self = this;
```
muloem: Given we are not using any closures, we can probably drop the use of `self` variable in this function.

Dashboard/app/js/lib/views/maps/map-views-common.js
```diff
@@ -887,80 +335,21 @@ FLOW.NavMapsView = FLOW.View.extend({
...
+    self.set('detailsPaneVisible', false);
+    if (self.marker) {
       self.map.removeLayer(self.marker);
-      self.hideDetailsPane();
-      $('#pointDetails').html('<p class="noDetails">'+Ember.String.loc('_no_details') +'</p>');
     }
 
-    if(!$.isEmptyObject(self.mediaMarkers)) {
-      for(mediaMarker in self.mediaMarkers) {
+    if (!$.isEmptyObject(self.mediaMarkers)) {
```
muloem: In the same spirit of trying to use our framework of choice I would suggest using `Ember.empty(...)`

Dashboard/app/js/lib/controllers/maps-controllers-common.js
```diff
@@ -6,158 +6,335 @@
...
-
-      if (!Ember.none(this.allPlacemarks)){
-        this.allPlacemarks.clearLayers();
-      }
+FLOW.mapsController = Ember.ArrayController.create({
+    content: null,
+    map: null,
+    marker: null,
+    markerCoordinates: null,
+    cartodbLayer: null,
```
muloem: This feels like a view layer concern

Dashboard/app/js/lib/controllers/maps-controllers-common.js
```diff
@@ -6,158 +6,335 @@
...
+    layerExistsCheck: false,
```
muloem: also feels like view layer concern

Issue #2247, opened 2017-09-12
Add meta data to all rows in data exports 


janagombitova:
# Context and problem

This year we have decided to improve how data is exported from Flow and solving the issues around repeated questions groups in the first problem we will try to address under this initiative. This issue holds a first small improvement. More is to come. 

If a user exports data where question groups are repeated, the repetitions are in separate rows. Current export puts each repetition of the questions in a new row, with a column indicating the number of repetitions for that submission. The repeated rows hold the answer and number of repetition but no further links to the submission (as data point ID or more). Users are not able to filter or sort the submissions, as they then loose part of the data (repeated row 2 and more end up at the end of the sheet). It is hard for them to see which group(s) is/are repeated. 

For more details, how users analyse data, results from testing, etc check this issue: https://github.com/akvo/akvo-product-design/issues/227

# Next steps 

With this issue we will not change the structure of exports. We will only add all meta data to all rows. Thus if a submission has 3 repeated answers in all three  #1496  of this submission the: Identifier, Display name, Device identifier, Instance Submission date, Submitter and Duration will be copied into all the three rows. 

See example from A/B testing (dummy form with dummy data). 
[Option B.xlsx](https://github.com/akvo/akvo-flow/files/1294918/Option.B.xlsx)

Questions: 
- currently users can add new repetitions to a submission via data cleaning import simply by adding a new row to the submission and properly adding the Repeat no.. Will this change with these changes? 
- currently the user can also remove a full repetition by simply deleting the row from the sheet and fixing the order of repetitions in the Repeat no. column with data cleaning. Will this rule change with these changes? 
- it still has to remain impossible to delete a form submission via data cleaning

@stellanl and @muloem any thoughts for this particular change? Let's discuss the further changes in the larger issue on product-design repo

janagombitova:
All looks good from my point of view. Will stress test in regression round and comment if needed

PR #2259, opened 2017-09-20
[#2247]Loop over more items. (Connect #2247)

stellanl:
Too weird report structure for RQGs for most consumers.
#### The solution
Add the survey metadata to all the repeat rows it applies to.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


stellanl:
RawDataSpreadsheetImporter.parseInstance() reads per-instance data from the first row och each instance. It ignores them and only looks at question data in rows where the "Repeat no." column is > 1.

Issue #2266, opened 2017-09-26
Long names break cascade layout - in levels 


janagombitova:
Since the design improvements introduced in this issue: https://github.com/akvo/akvo-flow/issues/2033 some already existing cascade UI issues resurfaced. Here is one of them: 

# Context 

The use of cascades has expanded from just using it to select administrative areas to using it for scoring. This shift in use results also in the level names not holding short one word text, as District or County, but actual questions and sentences. 

The current design does not support long level names. 

# Problem 
If the cascade level name is too long is breaks the layout in multiple places: 
- in the top of the cascade editor, if the level names are too long, or there are too many levels they overlap with the Add level button which is then hard to find in the screen. 
- In the level itself you loose the button to edit the name 
- This then results in the add, edit and delete clickable links for the options of the level to be out of range as well (see 2nd image) 
 
<img width="1261" alt="screen shot 2017-09-26 at 12 20 27" src="https://user-images.githubusercontent.com/12456965/30858060-4b9d6cc4-a2be-11e7-8f0f-9ae6c2b5e552.png">

<img width="1262" alt="screen shot 2017-09-26 at 12 17 36" src="https://user-images.githubusercontent.com/12456965/30856157-ce331f14-a2b7-11e7-8873-e2b757b72f35.png">

<img width="1264" alt="screen shot 2017-09-26 at 12 18 27" src="https://user-images.githubusercontent.com/12456965/30857960-01f257e2-a2be-11e7-8953-45b47717bfcc.png">

# Next steps 

With this issue we do not need to introduce a large redesign of how cascades are created a that is a large endeavour, but we should aim to resolve (if possible) the above mentioned issue in the current structure

janagombitova:
@loicsans Hey Loic, I just checked the changes on glimmerofhope and if there is a long name in the level of the cascade, it looks ok on the level arrows where the name does not show fully, but there are a few issues still pending: 
- on the level arrows the "add level" button still overlaps (circled in the 1st image)
- when editing the levels, in the level, you get that odd long scrolling to the side (see gif)

<img width="1257" alt="screen shot 2017-10-06 at 12 08 21" src="https://user-images.githubusercontent.com/12456965/31273508-4cf103be-aa8f-11e7-8572-d1a5d92d17f9.png">

![2](https://user-images.githubusercontent.com/12456965/31273586-9f54c424-aa8f-11e7-9c96-7f817c7889a7.gif)

[...]

PR #2269, opened 2017-09-27
Fixed long names in breadcrumb (connect #2266)

loicsans:
As well as level name editing missing actions.


janagombitova:
@muloem I agree. Let's handle the rest in a separate issue.

Issue #2224, opened 2017-08-23
Deleting survey with data seems to succeed, then silently fails


stellanl:
When the UI calls DELETE /rest/surveys/8200001, 200 is returned (if the survey exists), and the actual deletion is queued. So the UI removes the survey. 
Then the queue does a POST to /app_worker/surveytask with action: deleteSurvey which deletes the survey groups, which deletes the questions, which fails if there is data, an exception is thrown and 500 is returned. 
Two things are bad about this:
* The UI shows something that is only sometimes true. The survey reappears on a UI reload.
* The deletion is only halted cleanly is there is data for the *first* question. Otherwise some questions may be deleted, while others remain.

janagombitova:
From the users perspective, you should not be able to delete the form (how we call it in the users eyes) in the first place even if only one question has data collected to it regardless of the position of the question in the form. 

For now, we could reuse the same pattern - hit delete > ask if you are sure > show message it is not possible because there is data collected. (from the user's perspective) and the form does not get deleted. 

<img width="629" alt="screen shot 2017-08-23 at 09 28 25" src="https://user-images.githubusercontent.com/12456965/29608299-07d57a48-87f4-11e7-9af6-e69d6b25395c.png">

stellanl:
To make the UI behave, the SurveyRestService needs to check for existing data before it decides whether to delete or not. The existence of SurveyInstance entities for the survey seems to be the correct criterion, and should be inexpensive enough given that survey deletion is done manually one at a time.

stellanl:
Unfortunately, returning "failed" status to the UI only shows an error dialog with the supplied message.
The form is still deleted from the UI, and will return on a UI reload.

janagombitova:
Note to myself: Once this is ready for testing cross check if the issue below are still relevant: 
- https://github.com/akvo/akvo-flow/issues/1656
- https://github.com/akvo/akvo-flow/issues/1781

janagombitova:
What is the status of this change?

PR #2271, opened 2017-09-28
Issue/2224 deleting surveys (Connect #2224)

stellanl:
Deleting a form seems to succeed, but if there is data, it fails quietly, and the form reappears later.
#### The solution
Check with the backend using a /rest/surveys/?preflight=delete&surveyId=nnnn API call.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/controllers/permissions.js
```diff
@@ -446,7 +446,7 @@ FLOW.dialogControl = Ember.Object.create({
 
     case "delForm":
```
muloem: After reviewing, I believe that actually there is a simpler fix for this.  in this same switch statement, the case `delS` seems to have been previously applying to the deletion of a form.  This case has code that actually performs the preflight check and then only accepts to delete if the preflight check returns correct. take a look there and see what you can figure out

muloem:
We're going to redo this section so closing this one.

PR #2272, opened 2017-09-28
[#1367]Do not open folder being moved. Fix spelling. (Connect #1367)

stellanl:
User could move a folder into itself.
#### The solution
Disable selecting the folder being moved.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


stellanl:
Well, it's already marked special for being moved. Might not be clearer to mark it two ways at once...

muloem:
Exactly. I was thinking instead of the green highlight we should grey it out like its half way moved already

janagombitova:
I would say, let's test it out and see. Some colour improvements can be done along the road as well

Issue #2119, opened 2017-06-06
Simplify assignment modal dialogue - deleting


janagombitova:
<img width="694" alt="screen shot 2017-06-06 at 13 42 40" src="https://cloud.githubusercontent.com/assets/12456965/26827570/38b812a6-4abe-11e7-885c-d2bb5d83d5c7.png">

Current modal dialogue that shows after a user wants to delete an assignment can be improved and simplified. The buttons need to relate directly to the actions. The outcome of the users decision needs to be clearly stated. 

## Suggestions 

Change text to: "Delete assignment" 
Buttons: instead of OK > "Delete" ; leave Cancel as it is 
Colours: Keep only the white background. No need fro the grey area around the text field. 

Do we need to explain that "this cannot be undone" or is that clear from the delete action itself. I personally am in favour of removing this sentence, but would like to hear your thoughts.

Maseline:
I think it's nice to have a warning statement if the information being deleted is crucial

janagombitova:
Hi @Maseline after consulting with the design team a while ago we agreed that keeping the dialogue as simple as possible is the best way to go. So the expectation is to have: 
- Delete assignment? 
- buttons: Delete and Cancel

janagombitova:
Hi @Maseline the final change is: 
- Title: Delete assignment? 
- Buttons: Delete, Cancel 
- remove the text and background colour (grey) 
- Delete button has the same design as the current OK button 
- Cancel button has the same design as now

Maseline:
@janagombitova , one more commit and all changes are complete, however, wanted to let you know that changing the button text to Delete, that change is applied across all similar dialog boxes, meaning any other delete option that opened a dialog box now has delete and cancel as the button options, this is because of the way the dialog box was set up in general. Is this okay?

janagombitova:
@Maseline if you can cross check and confirm that it only affects the Delete dialogues, then it is actually perfect! Please add a note later to the test plan to cross check once more that only delete dialogues have the button change

Maseline:
Test plan
Select Devices from the navigation bar
Under devices select the Assignments list
Select the option to delete one of the assignments in the list, if none exists, create a new assignment and try deleting it.
The new dialog box should have Delete Assignment? as the heading, no warning message and the buttons should have the text Delete, Cancel

Maseline:
[...]

PR #2301, opened 2017-11-07
Simplify assignment dialogue deleting (connect #2119)

Maseline:
Assignments delete dialog showed a warning message : this cannot be undone, the button to delete was labeled OK and the heading of the dialog was long
#### The solution
Changed the placeholder text for the dialog header and Ok button in ui-strings.properties file
Removed the line of code that set the message for the delete assignment dialog in permissions.js file
Removed the background color for dialog  message in main.scss
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/locale/ui-strings.properties
```diff
@@ -333,7 +333,7 @@ _number = Number
 _number_of_question_groups = Number of question groups
 _number_of_questions = Number of questions
 _number_of_users = Number of users
-_ok = OK
+_ok = Delete
```
muloem: Rather than changing the text here, (this placeholder may be used by other elements in the dashboard) use a `_delete` placeholder in the dialog control button

Maseline: the challenge is that dialog control button is set in one file, application.handlebars from what i have seen so far, in order to change the text all i had to do was change the placeholder there, say from _ok to _delete, but that still changes it to delete everywhere.

muloem: > but that still changes it to delete everywhere.

yes. and thats ok.  All I meant was that if we edit the _ok placeholder, it may be used in other places where its inappropriate to have the text "Delete".  So rather use the _delete placeholder.

Maseline: i've replaced _ok with _delete, changed _ok back to OK, but it still doesn't deal with the issue since its one html file for the dialogues

muloem:
Closing this as we are redoing the assignments. Will incorporate there.

Issue #1839, opened 2016-11-23
After using data cleaning to modify response data the API still returns old data to Android device


valllllll2000:
There was an issue for akvo-flow-mobile:
https://github.com/akvo/akvo-flow-mobile/issues/451

It turns out the api is returning the old value. The dashboard has the correct value.

Steps to reproduce:
On dashboard create a monitored survey with just one text question for example "why?"
Once assigned the survey to your device create an answer for example "bc" and send it.
Using dashboard data cleaning export the excel, modify the answer to "because" and reimport it.
The dashboard now shows the correct answer.
Sync the datapoints on your device (using menu option "sync")
Open that entry you will see the answer is the old one: "bc".
In the logs you will see that the api returns something like this:

`{"surveyedLocaleData":[{"displayName":"","id":"7975-8kjs-e90d","identifier":null,"lon":null,"surveyGroupId":145192019,"lastUpdateDateTime":1479827303508,"lat":null,"surveyInstances":[{"qasList":[{"q":"147092013","a":"bc","t":"FREE_TEXT"}],"collectionDate":1479827149202,"uuid":"a9b51b74-31fa-4ee1-8ce9-bb419a40f668","surveyId":148092013,"submitter":"valeria_api19_tablet"}],"keyId":null}],"message":null,"offset":0,"code":"200","resultCount":1}`

I searched for the QuestionAnswerStore and it is the new value that shows. I also tried deleting the memcache but still the api returns the old value.

stellanl:
AFAIK, the API does not read from the QuestionAnswerStore table, but the surveyalValue table.

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/akvoflow-support-request-number-71638-from-nikki-at-akvo-dot-org

muloem:
@janagombitova So we are planning to sort this out for the next release. the issue is #2229

janagombitova:
@muloem  Ah nice. The partner will be so happy how quick we are in resolving issues. The machine is rolling well

janagombitova:
[...]

PR #2313, opened 2017-11-13
Update the datapoint geolocation based on updated responses (connect #1839)

muloem:

* The data point location was not being updated on the map when data was changed via data cleaning

#### The solution

* When saving data from data cleaning, we check for any `GEO` question answers and use those to update the datapoint location.  Note: We do this only when updating registration forms responses for existing datapoints.

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/com/gallatinsystems/surveyal/domain/SurveyedLocale.java
+++ b/GAE/src/com/gallatinsystems/surveyal/domain/SurveyedLocale.java
@@ -350,5 +350,14 @@ public class SurveyedLocale extends BaseDomain {
                 oldStyleIdentifier.substring(4, 8),
                 base32Uuid.substring(10));
     }
-    
+
+    public void setGeoLocation(String geoLocationString) {
+        String[] geoParts = geoLocationString.split("\\|");
+        if (geoParts == null || geoParts.length < 2 || geoParts[0] == null || geoParts[1] == null)
+            return;
+
+        this.latitude = Double.parseDouble(geoParts[0]);
+        this.longitude = Double.parseDouble(geoParts[1]);
+    }
+
 }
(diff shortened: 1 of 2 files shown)


GAE/src/com/gallatinsystems/surveyal/domain/SurveyedLocale.java
```diff
@@ -350,5 +350,14 @@ public static String generateBase32Uuid(String oldStyleIdentifier) {
                 oldStyleIdentifier.substring(4, 8),
                 base32Uuid.substring(10));
     }
-    
+
+    public void setGeoLocation(String geoLocationString) {
```
valllllll2000: Just in case: is it possible that geoLocationString is null?

GAE/src/org/waterforpeople/mapping/app/web/RawDataRestServlet.java
```diff
@@ -214,10 +214,13 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
             qasDao.delete(deletedAnswers);
 
             if (!isMonitoringForm && !isNewInstance) {
-                // Update datapoint name for this locale
+                // Update datapoint name and location for this locale
```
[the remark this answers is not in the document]
muloem: I would add it but maybe in another issue since this is focused on the issue of updating datapoint location

GAE/src/com/gallatinsystems/surveyal/domain/SurveyedLocale.java
muloem: Ahah! good catch!

PR #2325, opened 2017-11-17
Feature/2282 New export UI

loicsans:


loicsans:
Last commit fixes: 

- When selecting the collection date for Data analysis export I do not get the calendar selector as in Data cleaning exports > We have 2 set of date selector with same id, therefore making one of them not working.
-  For the description in the Comprehensive report I do not like the (Min, Max, Frequency, ...) changed to "(min, max, frequency and more)" and added a full stop after "Contains raw data".
-  Collection date placeholder in input not visible in firefox

muloem:
@loicsans We'll have to refactor a bit but merging for now 👍

Issue #2291, opened 2017-10-30
16-digit numbers get rounded in report cells


stellanl:
The changes in #267 did not fully take very large numbers into account.
Users have been collecting 16-digit ID numbers that then misbehave in the exported Excel sheets. They expect them to not be rounded or shown in "scientific" format.
Luckily, in most of Flow, a NUMBER answer is just like a TEXT answer but with different value checking. So no data has been lost.
The answer is not quite clear-cut. We can put any large number in text cells during the export, but users that collect large numbers where not every digit is significant presumably still want them in number cells.

We *could* add a comment to cells that were changed to string to make it more obvious that they need to be changed back if the user needs to do math on them.

https://flowhelp.reamaze.com/admin/conversations/akvoflow-support-request-number-80157-from-aulia-at-akvo-dot-org

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/akvoflow-support-request-number-80157-from-aulia-at-akvo-dot-org

janagombitova:
@stellanl I checked the file you shared this morning and it is a bit odd when I open it. I think there should be a comment there, but I see just a little line. 

<img width="1280" alt="screen shot 2017-11-14 at 12 08 30" src="https://user-images.githubusercontent.com/12456965/32776976-b9ebac80-c934-11e7-8438-f23653a11408.png">

janagombitova:
Why did you close the issue @stellanl ?

stellanl:
Oops, I may have gotten a little too enthusiastic moving it around in waffle...

janagombitova:
<img width="763" alt="screen shot 2017-12-12 at 14 51 49" src="https://user-images.githubusercontent.com/12456965/33892922-60325914-df5a-11e7-94f1-e65bade020d9.png">

Long numbers show fully. All good

PR #2342, opened 2017-11-23
Generation of 16-digit+ numbers as text cells in spreadsheet (Connect #2291)

stellanl:
16-digit numbers got rounded.
#### The solution
output them as text, and warn the user with a note on the column.
## Checklist
* [X] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -1302,7 +1353,7 @@ protected void createRawDataHeader(Workbook wb,
                     if (qgSheetMap.containsKey(group.getKeyId()))   {
                         Sheet groupSheet = qgSheetMap.get(group.getKeyId());
                         int metaEnd = addMetaDataHeaders(groupSheet, true);
-                        addComment(groupSheet);
+//                        addComment(groupSheet, 0, 0, REPORT_COMMENT);
```
muloem: if its commented out I think lets remove it completely for now.  When we need it again we will add it in.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -431,7 +439,8 @@ public int compare(String o1, String o2) {
                 wb.setActiveSheet(isFullReport ? wb.getNumberOfSheets()-1 : 0);
                 wb.write(fileOut);
                 fileOut.close();
-
+                
+                setDocumentProperties(fileName, surveyId);
```
muloem: I think we should also cut this part out. Maybe we can have it in another branch and only if verified that it doesn't cause any performance issues, merge it. For this issue lets only focus on the generating of the text for 16-digit+ numbers.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -442,6 +451,38 @@ public int compare(String o1, String o2) {
...
+        POIXMLProperties.CustomProperties custProp = props.getCustomProperties();
+        custProp.addProperty("Version", "Akvo Flow v1.9.25");
+        custProp.addProperty("Generated", new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ").format(new Date()));
+        
+        FileOutputStream fos = new FileOutputStream(fileName);
+        workbook.write(fos);
+        fos.close();
+    }
+
+    private void addColumnWarnings(Workbook wb) {
```
muloem: As this is unused for now, I think we should also remove till we get round to using it.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
stellanl: Ok, I just thought the big change in reporting formats would be a good time to start tagging the files.

Issue #2344, opened 2017-11-23
Comprehensive reports not generating with new UI 


janagombitova:
No comprehensive reports being generated after the changes to exports and the new UI. Found a bug while testing

stellanl:
The reportMode parameter is not sent to the server.

janagombitova:
The comprehensive report gets generated now, but there are issues in the Raw data sheet. The structure is off. 

Here is an example of a fake data set:

[DATA_CLEANING-214449116 (1).xlsx](https://github.com/akvo/akvo-flow/files/1501206/DATA_CLEANING-214449116.1.xlsx)
[COMPREHENSIVE-214449116.xlsx](https://github.com/akvo/akvo-flow/files/1501207/COMPREHENSIVE-214449116.xlsx)

janagombitova:
works again 👍

PR #2350, opened 2017-11-24
Use sheet method to track last-line-written everywhere. (Connect #2344)

stellanl:
Comprehensive report was broken; all raw data was on the first line
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
@@ -623,7 +623,8 @@ public class GraphicalSurveySummaryExporter extends SurveySummaryExporter {
 
                 
             } else { //just one sheet - do all at once with a global repeat column
-                writeInstanceData(baseSheet, instanceData,
+                int baseCurrentRow = baseSheet.getLastRowNum() + 1;
+                baseCurrentRow = writeInstanceData(baseSheet, baseCurrentRow, instanceData,
                         generateSummary, nameToIdMap, collapseIdMap, model);
             }
         }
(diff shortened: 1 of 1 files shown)


stellanl:
Ok, not quite right yet...

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -747,7 +747,8 @@ private synchronized int writeInstanceData(
 
         Row firstRow = getRow(startRow, sheet);
 
-        writeMetadata(sheet, instanceData, startRow, (int) instanceData.maxIterationsCount, true);
+        //maxIterationsCount is actually the max iteration index; 0 for 1 iteration...
+        writeMetadata(sheet, instanceData, startRow, (int) instanceData.maxIterationsCount + 1, true);
```
muloem: since we already pass in the `instanceData` parameter, do we need to pass in the `maxIterationsCount` as a parameter as well?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -747,7 +747,8 @@ private synchronized int writeInstanceData(
 
         Row firstRow = getRow(startRow, sheet);
 
-        writeMetadata(sheet, instanceData, startRow, (int) instanceData.maxIterationsCount, true);
+        //maxIterationsCount is actually the max iteration index; 0 for 1 iteration...
```
muloem: Maybe its worth it renaming the variable in that case to make it clearer?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
muloem: Does this take into account the situations where there is a missing iteration index. e.g. we have iteration `0, 1, 3`?

stellanl: We never have that. instanceData will create empty filler iterations.

stellanl: And to answer the first question, for split sheets we call with another count.

Issue #2368, opened 2017-11-29
 Importing in new data using the Data cleaning export format


janagombitova:
# Context

Users capture data with Flow. But sometimes they already have data stored somewhere else and they either want to have it stored in Flow with their other data, or they want to use the data for further monitoring in Flow. Simply speaking they want to have their data in Flow.

Currently what is possible is using data cleaning import to import new data into Flow. How:

- User creates a survey form with questions matching their data
- Exports the Raw data report
- Fill in her new data into the sheet
- Does not fill in the Identifier, Instance, Device
- Imports the file using Data cleaning to the survey form
- There is one limitation: you cannot import new data into a monitoring form.

# How it works since 1.9.25 release 
In the 1.9.25 release we released new export structures and **Data cleaning export** is the one users get when cleaning data. The "old" raw data report is not longer downloadable from Flow. 

However, to import in new data users can use the structure of raw data reports. 

# Opportunity

We need to make it possible to import in new data using a data export structure we currently offer in Flow, to not have a third structure hidden in the shadows. 

With the new export structure using data cleaning for import is ok, if there are no repeated question groups in the survey form. However, if there are any, we need to figure out how to make this import possible as the user currently has no way to link that data in the separate sheets (the Flow identifier is not generated yet as they data is to be imported as new data).

Do we want to have this possibility hidden under that name of data cleaning?
How can we make importing new data to Flow possible?

@stellanl can you share your ideas here?

stellanl:
The new RQG responses need to be linked to the new instance (on the base sheet) somehow.
I propose that we link them with a special flavour of unique values in the Identifier (or prerhaps the Instance) column, like this:
"new-1", "new-2"... etc.
That way those values are then easy to recognise and strip out just before the upload to the database.  And with those columns empty they will be treated as new.

janagombitova:
Great! Stellan, once ready to test can you explain here how I should go about importing in new data with repeated groups? The steps and what I need to add and a test plan? Cheers

janagombitova:
@stellanl I tested importing in new data with the Data cleaning export structure for submissions with no repeated questions groups and that worked perfectly. 

Can you tell me how to do the ones that might have a repeated question group? How am I supposed to link the submissions in the main sheet and what should I put in where?

stellanl:
Just like I outlined above. Put new-1, new-2 etc in the first column. They work just like an empty cell, but can be referred to on the repeat sheets.

janagombitova:
[...]

PR #2371, opened 2017-11-30
Import new data from split sheets (Connect #2368) 

stellanl:
Cannot import new data on the data cleaning report sheets.
#### The solution
Enable that by having a special temp identifier pattern.
## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
@@ -461,11 +461,11 @@ public class RawDataSpreadsheetImporter implements DataImporter {
             } else {
                 if (rowIx > repSheet.getLastRowNum()) { //fell off the end
                     rowIx = 1; //start over
-                    }
-                rowIx++;
-                if (rowIx == currentRowIndex) { //back to where we started
-                    return rowIx; //not found; there were 0 iterations
                 }
+            }
+            rowIx++;
+            if (rowIx == currentRowIndex) { //back to where we started
+                return rowIx; //not found; there were 0 iterations
             }
         }
         //Found one row for the instance, read them all
--- a/GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
@@ -455,12 +455,12 @@ public class RawDataSpreadsheetImporter implements DataImporter {
         while (true) {
             row = repSheet.getRow(rowIx);
             if (row != null
-                    && !isEmptyCell(row.getCell(identifierColumnIndex))
+                    && row.getCell(identifierColumnIndex) != null
                     && row.getCell(identifierColumnIndex).getStringCellValue().equals(dataPointIdentifier)) { //found!
                 break;
             } else {
-                if (isEmptyRow(row)) { // a row without any cells defined
-                    rowIx = 1; //end of sheet; start over
+                if (rowIx > repSheet.getLastRowNum()) { //fell off the end
+                    rowIx = 1; //start over
                     }
                 rowIx++;
                 if (rowIx == currentRowIndex) { //back to where we started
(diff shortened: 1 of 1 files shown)


GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -85,6 +85,7 @@
     public static final String DURATION_COLUMN_KEY = "surveyalTime";
     
     public static final String METADATA_HEADER = "Metadata";
+    public static final String NEW_DATA_PATTERN = "^new-\\d+"; // new- followed by one or more digits
```
muloem: Where is this used? i can't seem to locate it..

stellanl:
I know I wrote and tested the new-nnn stripping code before, but must have pressed Undo or something before committing it...
One thing about matching "new-nnn" is that (some) spreadsheet apps like to capitalise strings on manual input, so there may be a case for also allowing "New-nnn" or even a wholly case-insensitive match.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
stellanl: Fixed.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -434,18 +445,18 @@ private Integer parseRepeatsForInstance(InstanceData instanceData,
...
         while (true) {
             row = repSheet.getRow(rowIx);
             if (row != null
-                    && row.getCell(instanceIdColumnIndex) != null
-                    && row.getCell(instanceIdColumnIndex).getStringCellValue().equals(SurveyInstanceId)) { //found!
+                    && row.getCell(identifierColumnIndex) != null
+                    && row.getCell(identifierColumnIndex).getStringCellValue().equals(dataPointIdentifier)) { //found!
                 break;
             } else {
                 if (isEmptyRow(row)) { // a row without any cells defined
```
muloem: My understanding is that we are scanning the datapoint identifier column to find matching rows, but if we use this we end up scanning the whole row perhaps needlessly.  Would it be enough to search here for `isEmptyCell(identifierColumnIndexCell)`? And just ignore those that don't have an identifier?

stellanl: I thought this was slightly more efficient, but sure.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -434,18 +445,18 @@ private Integer parseRepeatsForInstance(InstanceData instanceData,
...
         while (true) {
             row = repSheet.getRow(rowIx);
             if (row != null
-                    && row.getCell(instanceIdColumnIndex) != null
-                    && row.getCell(instanceIdColumnIndex).getStringCellValue().equals(SurveyInstanceId)) { //found!
+                    && !isEmptyCell(row.getCell(identifierColumnIndex))
+                    && row.getCell(identifierColumnIndex).getStringCellValue().equals(dataPointIdentifier)) { //found!
                 break;
             } else {
                 if (isEmptyRow(row)) { // a row without any cells defined
```
muloem: I was actually referring to this line i.e. line 462.  since we are checking for a value in the datapoint identifier column why scan the entire row?

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -434,41 +445,41 @@ private Integer parseRepeatsForInstance(InstanceData instanceData,
...
-                && row.getCell(instanceIdColumnIndex).getStringCellValue().equals(SurveyInstanceId) //TODO: use "identifier" as link instead to handle new data?
+                && row.getCell(identifierColumnIndex) != null
+                && row.getCell(identifierColumnIndex).getStringCellValue().equals(dataPointIdentifier)
                 && row.getCell(repeatIterationColumnIndex) != null
                 && row.getCell(repeatIterationColumnIndex).getCellType() == Cell.CELL_TYPE_NUMERIC) {
             Long rep = (long) row.getCell(repeatIterationColumnIndex).getNumericCellValue(); //might throw on huge number
-            //check rep for sanity
-            if (rep<1) { break;}
+            //check repeat no for sanity
+            if (rep < 1) { break;}
```
muloem: Should we perhaps `continue` and just skip such a row but process the rest that are ok?

Issue #2373, opened 2017-12-01
Update translation strings for new reports UI


muloem:
The new reports UI is missing language translations

janagombitova:
Translations missing as they should in Transifex. Now we need to get them translated and in the build, but this issue can be considered done 👍

PR #2374, opened 2017-12-01
Update translation strings (connect #2373)

muloem:
#### The solution
* Add the missing translation strings that were added to new exports UI

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/locale/en.properties
```diff
@@ -69,6 +72,7 @@ Before\ you\ save\ or\ publish\ a\ survey,\ you\ have\ to\ give\ the\ survey\ a\
 Before\ you\ save\ or\ publish\ a\ survey,\ you\ have\ to\ select\ a\ survey\ type = Before you save or publish a survey, you have to select a survey type
 Build\ chart = Build chart
 Bulk\ upload\ data = Bulk upload data
+By\ repeating\ this\ question\ group\ you\ can\ ask\ questions\ in\ this\ group\ multiple\ times,\ e.g.\ when\ you\ want\ to\ ask\ the\ same\ questions\ to\ each\ member\ of\ a\ household.\ You\ will\ decide\ on\ the\ number\ of\ repetitions\ when\ capturing\ the\ answers\ in\ the\ Akvo\ Flow\ app.\ You\ cannot\ make\ a\ question\ outside\ of\ this\ group\ dependent\ on\ a\ question\ inside\ the\ group. = By repeating this question group you can ask questions in this group multiple times, e.g. when you want to ask the same questions to each member of a household. You will decide on the number of repetitions when capturing the answers in the Akvo Flow app. You cannot make a question outside of this group dependent on a question inside the group.
```
valllllll2000: I am trying to understand the logic behind it: so on the left is the key which has the spaces escaped with "\", correct? In that case shouldn't the keys be shorter and more "explanotary" of where they are used, say for example "title\ left\ tab" or "question\ help\ description".

GAE/src/locale/ui-strings.properties
```diff
@@ -1,4 +1,3 @@
-a_link_emailed = A link will be emailed to you shortly.
```
valllllll2000: Why is this file diferent from en.properties? keys do not follow the same convention (_ instead of \ ), the naming is diferent: why ui-strings.properties and the other en.properties, does this one not get translated?

muloem: the ui-strings.properties is the one with the placeholders but the en.properties is the source file in transifex.  I believe its setup this way in order to not have placeholders appear for other languages but rather to have a plain english text string.

GAE/src/locale/en.properties
muloem: in the en.properties we have the exact string. The one on the left is the key and thats why its escaped, and the one on the right is the string.

Issue #2319, opened 2017-11-15
Remove "No additional settings for this question type" 


janagombitova:
# Context

In the survey editor once the user selects the question type, she has some more options to choose from that we call Settings. Some questions however do not have any more settings to work with. The we show a grey space with the sentence that there are no additional settings for the question type.

<img width="1219" alt="screen shot 2017-11-15 at 09 01 12" src="https://user-images.githubusercontent.com/12456965/32825097-ea1b167a-c9e3-11e7-9435-9d46ffdc5eb6.png">

The questions are: 
- photo 
- video 
- date 
- signature 

# Opportunity 
As there are so Settings for the above mentioned question types, the grey space with the text "No additional settings for this type of question" does not really tell anything to the user extra. It takes up space. Could we remove the entire grey box and the text and thus declutter the screen?

Maseline:
Test Plan
- Open any survey
-Try adding a new question and keep changing the question type before saving
-Additional settings text box should show up for only questions that have additional settings

janagombitova:
Nicely done, all works as expected on Firefox, Chrome and Safari

PR #2376, opened 2017-12-05
Remove no additional settings for questions with no extra settings (connect #2319)

Maseline:
When adding a question to a form, after selecting the question type, additional settings may be displayed in a grey text box. 
For question types with no additional settings e..g. video,date,signature,photo, a grey text box with "No additional settings for this type of question" is displayed. This was to be removed in this issue

#### The solution
Removed the {#if} clause displaying the message "No additional settings for this question type" from question-view.handlebars .
Added a function in question-view.js to identify question types with extra settings in the template.
Added a condition for extra settings to be displayed in question-view.handlebars: {{if view.hasExtraSettings}}, thus the div with extra settings is displayed unless the question type has no extra settings
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/views/surveys/question-view.js
+++ b/Dashboard/app/js/lib/views/surveys/question-view.js
@@ -113,15 +113,12 @@ FLOW.QuestionView = FLOW.View.extend({
 	}
   }.property('this.type').cacheable(),
 
-  hasExtraSettings: function(){
+  hasExtraSettings: function () {
     var val;
     if (!Ember.none(this.type)) {
       val = this.type.get('value');
-      if(val === 'PHOTO' || val === 'VIDEO' || val === 'DATE' || val === 'SIGNATURE'){
-        return false;
-      }else{
-        return true;
-      }
+      return val === 'GEOSHAPE' || val === 'CASCADE' || val === 'NUMBER' || val === 'GEO' 
+      || val === 'FREE_TEXT' || val === 'SCAN' || val === 'OPTION' || val === 'CADDISFLY';
     }
   }.property('this.type').cacheable(),
 
--- a/Dashboard/app/js/lib/views/surveys/question-view.js
+++ b/Dashboard/app/js/lib/views/surveys/question-view.js
@@ -113,14 +113,6 @@ FLOW.QuestionView = FLOW.View.extend({
 	}
   }.property('this.type').cacheable(),
 
-  amNoOptionsType: function () {
-    var val;
-    if (!Ember.none(this.type)) {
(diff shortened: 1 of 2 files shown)


Dashboard/app/js/templates/navSurveys/question-view.handlebars
```diff
@@ -62,9 +62,8 @@
   {{/if}}
 
     <!-- Question specific material -->
-<div class="questionOption floats-in">
-
     {{#if view.amOptionType}}
+      <div class="questionOption floats-in">
```
muloem: Does this need to be inside each individual `{{if}}` block?  Why not outside the block like it was so that its not repeated over and over?

Maseline:
@muloem it's because leaving it as an external tag means the grey text box is present regardless of whether there are extra settings or not, so I might have removed the text but leaving the div that way would leave the empty grey box there anyway, putting the div within the if's means it shows up only when there are extra settings

Maseline:
@muloem , new commit Refactored code with new solution, also edited the pull request comment, specifically the solution part

Dashboard/app/js/templates/navSurveys/question-view.handlebars
```diff
@@ -138,7 +139,7 @@
         {{else}}
         <div class="errorLoading">{{t _failed_load_caddisfly_tests}}</div>
         {{/if}}
-        </label>
+        </label
```
muloem: You somehow dropped the ending `>` here...

Dashboard/app/js/lib/views/surveys/question-view.js
```diff
@@ -113,11 +113,15 @@ FLOW.QuestionView = FLOW.View.extend({
 	}
   }.property('this.type').cacheable(),
 
-  amNoOptionsType: function () {
+  hasExtraSettings: function(){
     var val;
     if (!Ember.none(this.type)) {
       val = this.type.get('value');
-      return val === 'PHOTO' || val === 'VIDEO' || val === 'DATE' || val === 'SIGNATURE';
+      if(val === 'PHOTO' || val === 'VIDEO' || val === 'DATE' || val === 'SIGNATURE'){
```
muloem: the condition is `{{#if hasExtraSettings}}{{/if}}` and there are a number of if statements following for the question types that have extra settings. Its probably cleaner here to simply put that list e.g. 

`return this.get('amOptionType') || this.get('amFreetextType') || .... ;`

and that way if any type doesn't match the result will be false.

Issue #2309, opened 2017-11-08
Remove localised header code from report engine


stellanl:
Since we are removing the UI for translated report headers in #2286, we can clean up the report engine code.

stellanl:
One question: translated options etc. are still exported that way. Keep or remove?

janagombitova:
@stellanl do you mean when a form is translated to one language, if we still want to add the translations to the export (to show the original language and then the translation)? Like in this issue: https://github.com/akvo/akvo-flow/issues/1828

stellanl:
Well, right now I think we do not so much add as replace. We output only the translated text.

janagombitova:
With @stellanl we agreed everything related to translations can be removed from the exports code.

janagombitova:
@stellanl anything in particular I need to test or to keep in mind when testing?

stellanl:
No, there is no way (any longer) to provoke the report engine by asking for Spanish/whatever from the UI.

janagombitova:
Ok. Then this one is also good for releasing

PR #2377, opened 2017-12-05
[#2309]Remove all report localisation code. (Connect #2309)

stellanl:
Localised report code was unused
#### The solution
Remove it

The removal is a little hard to test without a UI that lets you specify a locale, and we recently removed that, but the tester should verify that the reports still output the expected column headers and other labels (such as in the comprehensive stats).

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -372,18 +203,8 @@ public void export(Map<String, String> criteria, File fileName,
         try {
             Map<QuestionGroupDto, List<QuestionDto>> questionMap = 
                     loadAllQuestions(surveyId, performGeoRollup, serverBaseUrl, apiKey);
-            //minimal data plus cascade level names
-            if (useQuestionId && DEFAULT_LOCALE.equals(locale)) {
-                loadQuestionOptions(surveyId, serverBaseUrl, questionMap, apiKey);
```
muloem: Where will the question options be loaded from?

stellanl: They don't need translated names anymore.

Issue #2385, opened 2017-12-13
In Monitoring tab remove question IDs


janagombitova:
# Context

When a user views data in the monitoring tab she first selects a data point where she sees the list of submitted forms, and then she selects the form she want to view. Here she sees items: ID, Question and Answer. 

<img width="1085" alt="screen shot 2017-12-13 at 10 29 04" src="https://user-images.githubusercontent.com/12456965/33931493-785f6694-dff0-11e7-9314-cd63534f8b21.png">

# Change 

In the detailed submission view the user does not need to see the Question ID, as this is an internal item we use and makes no sense or gives no value to the user. It only clutters her screen. The idea is to remove the question ID column to only show Questions and Answer.

bogere:
## Test plan
* In the monitoring tab after you clicked the data navigation tab
*  select the folder or survey whose data you want to monitor
*  When you  view data in the monitoring tab you  first selects a data point where you will see the list of submitted forms, and then  select the form you  want to view. 
* you will see that Question ID column is no longer present

janagombitova:
@bogere we have some issues related to your changes I believe: 

# 1 Data is not loading

After I select my survey and see the list of data points, I hit View details but not details load. I see the table structure but no content. There is no error in the Console I can point out. 

![monitoring](https://user-images.githubusercontent.com/12456965/35285148-e8de8850-005c-11e8-95ed-36dda457fc56.gif)

# 2 UI changes to the view window

I think you might have changed how the UI looks once user hits View data. The 'x' to close is missing and replaced with a text aligned to the left. This change is also visible in the Inspect data tab view viewing a submitted form. Please restore to what it was (shown as then in the images below).  

![screen shot 2018-01-23 at 16 51 11](https://user-images.githubusercontent.com/12456965/35285479-c5402ab0-005d-11e8-81bd-c284728174d5.png)
![screen shot 2018-01-23 at 16 53 46](https://user-images.githubusercontent.com/12456965/35285590-198533ea-005e-11e8-86a8-c38a00196f71.png)

stellanl:
This might be the result of #2198 changing a typo in main.css. There was a "clodeDialog" style that was corrected to closeDialog...

janagombitova:
Ah. Can you confirm and check @stellanl and @bogere ?

stellanl:
The style that is now no longer hidden by the typo is:
   &.closeDialog {
        position: absolute;
        top: 10px;
        right: 10px;
        display: block;
[...]

PR #2391, opened 2017-12-14
In monitoring tab remove question ID (connect #2385)

bogere:
In the detailed submission view the user does not need to see the Question ID, as this is an internal item we use and makes no sense or gives no value to the user. It only clutters her screen. The idea is to remove the question ID column to only show Questions and Answer.
#### The solution
Removed the question ID column to only show Question and Answer
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


bogere:
Maybe I  specify the alignment in the main.scss file because if the part is not styled, the table headers will be wrongly aligned
table alignment

Issue #2397, opened 2017-12-20
Remove metrics


janagombitova:
# Context 

We have a functionality on Flow called metrics that has not been enabled to all instance, was used only on one as a custom build feature. The partner has not used the feature and now is not using Flow anymore either. 

In the future, the problem this functionality was aiming to solve and value it provided will be re-researched and re-discovered so we can see if it is something valuable for all Flow users. 

# Opportunity 
Remove functionality and everything related. Make sure nothing else breaks.

stellanl:
We could resurrect the "Statistics" tab for everyone but use it to show per-instance stats, like number of forms submitted etc.

janagombitova:
@stellanl we definitely need to find a way to show usage stats, but let's finish up the work we have planned out first for the exports and then we can pick up this one. We have an issue here: https://github.com/akvo/akvo-product-design/issues/265 and can pick up the discussion again once we close the work on exports

stellanl:
In the backend org.waterforpeople.mapping.app.web.rest.MetricRestService implements the /metrics endpoint.
I am removing that and the domain and dao classes for Metrics and SurveyMetricMapping.

Question also had metrics-related field, and SurveyalRestServlet would include metric info.

janagombitova:
Not sure how to test this out considering it was only enabled on akvoflow-12 and that is closed for a while now. I guess just regression testing to ensure all works well? @stellanl what are your thoughts?

stellanl:
Yep. Typical case of trying to prove a negative...

janagombitova:
Regression testing going well. Ready for release 👍

PR #2404, opened 2018-01-02
[#2397]Remove attributes and statistics from UI. (connect #2397)

stellanl:
Metrics UI code no longer in use
#### The solution
Cut it out
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

To test, set showStatisticsFeature = true in config file and redeploy. Nothing should change in the UI.


stellanl:
I thought we might reuse it to control showing of instance-statistics in the UI, but perhaps that should either be on for everybody or have a new variable.

muloem:
I think we may as well launch it to all when the time comes. no need to have more feature flags and some instances running something and others not.

Issue #2170, opened 2017-06-27
Question IDs confusion 


janagombitova:
Most of our users do not know what the question IDs can be used for. 

Even though the explanation says what the functionality is for; the location of the feature (defining it when defining the question) is good and the validation to ensure all IDs are unique is also well implemented, we have heard during few interviews around data exports that it is a shame you cannot define the value name to not have to deal with the long question names. Thus, users realise this is important to have but either they: 
- do not know at all that this is possible via the question IDs
- do not know what the question ID stands for 
- do not know how to get it out in the exports (handled in a separate issue) 

<img width="672" alt="screen shot 2017-06-27 at 07 39 44" src="https://user-images.githubusercontent.com/12456965/27572360-d15a0246-5b0b-11e7-8276-6cd6fb4c332c.png">

The idea of the question ID is to act as a value name in data export, so the analyst does not have to deal with long column headers (normally question names) 

With this issue we want to answer the question: How can we make the value of question IDs more clear to our users?

stellanl:
One way to make their function clearer is generating a default id automatically. It could be as simple as q001, q002, q003. The user can then either improve it or keep a usable default value. Either way someting shows up in the reports...

janagombitova:
I disagree on that one Stellan. The ID is considered a variable name that users need to be aware of as they based on it know what to search for in the data. If we generate another ID for them, they will have no use of it as they won't know where the information came from and what it means. Secondly, from the case we have in Ghana, the link Flow data in their national database based on these IDs, that were first created in the database. So then they would have to go into each form and not only create their own IDs but also delete ours.

stellanl:
Oh, I did not mean to generate it hidden, silently. I mean when you create a q it gets a default ID so that you see that this is an expected thing. But it's right there to be edited from the start.

muloem:
Perhaps we should consider renaming the field. First because it already causes confusion internally.  question Id currently refers to two completely different things.  Second because it perhaps doesn't mean anything to a user who is an analyst.  Something like _question code_ might be more appropriate to the users

janagombitova:
@stellanl I still think we should not generate the IDs for users, visibly. It is not something everyone uses (as many of our users identify the data based on the question and do not import Flow data to other tools).

I agree with you @muloem. Maybe a new name will make it more understandable and some newsletter communication and a good support example can do the trick. Making it more clear in exports, is a different discussion. 

Options: 
- question code
- variable name (mentioned by CDC) 
[...]

Issue #2143, opened 2017-06-06
Question IDs - simplify validation


janagombitova:
# Current implementation 
When creating a question ID we validate two things: the format, uniqueness. 

If the input does not pass the validation we show in red the info above the text field. The user can hit Save, to save the question, but he will receive a modal dialogue telling him the same info as the red error message. He needs to confirm, what brings him back to the question. He cannot save the question until he makes adjustments to the ID

<img width="1196" alt="screen shot 2017-06-06 at 12 29 37" src="https://user-images.githubusercontent.com/12456965/26837607-27cfb9d8-4ade-11e7-85c4-227b55dead82.png">

<img width="1216" alt="screen shot 2017-06-06 at 12 29 46" src="https://user-images.githubusercontent.com/12456965/26837619-335293e8-4ade-11e7-9377-4af4b068e788.png">

**Validation error messages**

<img width="1221" alt="screen shot 2017-06-06 at 12 29 26" src="https://user-images.githubusercontent.com/12456965/26837646-433519b6-4ade-11e7-8e9b-82503c13f254.png">

<img width="1081" alt="screen shot 2017-06-06 at 12 29 56" src="https://user-images.githubusercontent.com/12456965/26837661-4be4b648-4ade-11e7-8cf3-a20a0863a66a.png">

 **Modal dialogues after trying to save the question** 

If the user hits Cancel, then the ID is not saved. 

# Improvement 
Would it be possible to instead of showing that the ID format is not correct, being able to click on save and getting the pop up message, to simply validate the field (as we do) and disable the save button, removing the modal dialogues? 

Secondly, the text for the uniqueness validation needs to be improved: 
Current text: the question id is not unique 
New text: 
- option A: Question ID is not unique 
- option B: Question ID already exists 
- option C: ... to be defined ...

@muloem and @kymni what are your thoughts?

janagombitova:
Before we make any improvements to the validation of the question IDs we need to agree on if we change how we call this feature: https://github.com/akvo/akvo-flow/issues/2170

kymni:
#2170 remains open, what's the way forward for this?

janagombitova:
My suggestion would be to 1st handle #2170 so move away from Question ID and implement Variable name throughout all Flow components and then handle this issue by disabling Save is the Variable name is not correct and simplify the validation notes.

janagombitova:
[...]

PR #2420, opened 2018-01-09
Simplify question ID (variable name) validation (connect #2143, #2170, and #2444)

kymni:
Users were able to click on save while editing a question even if there was a question ID validation error which would trigger a modal alert with the validation error
#### The solution
Disabled the save button if there is a validation error so modal alert is never displayed and changed all phrasing of "question ID" to "variable name"
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/views/surveys/question-view.js
```diff
@@ -512,7 +512,7 @@ FLOW.QuestionView = FLOW.View.extend({
         if (isUnique) {
           args.success();
         } else {
-          args.failure('the question id is not unique');
+          args.failure(Ember.String.loc('_question_id_not_unique'));
```
muloem: May I suggest we rename the placeholders as well in order to keep terms consistent... i.e. `_variable_name`

GAE/src/locale/ui-strings.properties
```diff
@@ -376,11 +376,12 @@ _qg_delete_header = Are you sure you want to delete this question group, with al
 _question = Question
 _question_help_tooltip = Question help tooltip
 _question_help_tooltip_optional = Question help tooltip: (optional)
-_question_id = Question ID
-_question_id_mandatory = The question ID is mandatory
-_question_id_must_be_valid_and_unique = Question ID must be valid and unique
-_question_id_only_alphanumeric = The question ID can only contain characters a-z, A-Z, 0-9, - and _
-_question_id_tooltip = The question ID can be used as the column header in data exports. Many tools such as QGIS or ArcGIS require this. It can also be used when importing data into a different database through the Akvo API
+_question_id = Variable name
```
muloem: Same here re the placeholders naming

Issue #1838, opened 2016-11-23
Modified date in survey overview does not update


janagombitova:
When looking at the list of surveys the users gets two date indications, of the date when the survey was **created** (not changeable) and of the date when the survey was **modified** (should change with the survey). However the **modified** date does not update after changing the survey, with or without publishing. 

![modified date](https://cloud.githubusercontent.com/assets/12456965/20554349/a6b320f4-b15b-11e6-8b3e-23422daf32e9.gif)

kymni:
It appears the survey group update endpoint only triggers when the survey group basics (Survey title, privacy type, etc) are modified. Potential fix is to set the survey group's "lastUpdateDateTime" property with the latest timestamp each time a survey update is triggered

janagombitova:
I agree. The modified date should change when any change is made to the survey or one of the forms in the survey, so covering actions as: 
- change to survey name
- change to survey privacy 
- change to monitoring enabled 
- change to data approval enabled 
- change to form name 
- change to form translation
- change to question group 
- change to any question

kymni:
The "lastUpdateDateTime" property of both the surveys and survey_groups endpoint  is system generated. We can therefore take advantage of this by setting the property to empty before committing to trigger the update.

muloem:
@janagombitova @kymni So we currently have a solution for this issue where we show the latest modified date of a survey, however, the consequence is that we have two extra calls to the backend for saving a question (save a survey and save the associated form).  I am not sure how this behaves on a slower connection and whether or not it will disrupt things a little so before we look into that I have a bit of devil's-advocate type question.  Do we _really_ need to have the modified date time on a survey?  What do the users really use it for?  And can we address that _what_ in another way?  We've made some good progress on improving performance so I would like us to be sure that if we add something that may degrade it, there is a strong argument why and no other alternatives.

janagombitova:
@muloem I agree that any changes we made cannot compromise performance, we should aim to always speed things up. So if the fix is slowing things down, then we should reconsider the approach. 

What is the value of the modified date: 

[...]

PR #2439, opened 2018-01-18
Update survey group modified date (connect #1838)

kymni:
The modified date was not updating when some changes such as adding questions were made to the survey
#### The solution
Update the "lastUpdateDateTime" property for all survey related changes
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


kymni:
If we don't trigger the "lastUpdateDateTime" property change on the front end, then the alternative would be to trigger it on the back-end. This should be done on the questions, survey_groups, and surveys endpoints by calling a method that would trigger the property change

kymni:
Closing this as we look for an alternative solution

Issue #1765, opened 2016-09-27
Forms of copied survey does not show to users


janagombitova:
### Description

If a user has permissions in a specific folder and someone makes of copy of a suvrey with multiple forms into that folder the user does not see the forms, only the survey. The workaround currently in place is to rename the forms and save the changes. The user will be able to see and work on the forms after reloading the dashboard. 
### Steps to take to recreate issue
1. Have a test user with access to folder B 
2. Copy a monitored survey (with multiple forms) from folder A to folder B
3. Log into the dashboard with the test user
4. Open the copied survey in folder B
5. You should be able to see the survey but as if it had no forms created 
### Support tickets with this issue:
- https://akvoo.reamaze.com/admin/conversations/copied-surveys-not-visible-to-certain-users
- private email from ICCO from 27 September: ... A strange issue: recently I have copied a survey into the icco instance Ethiopia folder. When my colleague in Ethiopia opens the folder and survey the survey forms (and hence the groups and questions) do not appear...

janagombitova:
Actually, the survey does not necessarily need to be monitored (changed the name of the issue). Simply the forms do not show

The workaround is to rename the survey and forms, save the changes and then for the user to reload Flow and all works again.

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/support-for-ushhd-project

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/akvoflow-support-request-number-89344-from-kwandolowski-at-waterforpeople-dot-org

kymni:
@janagombitova on it

stellanl:
Does the access check depend on the path field? Because that is not reliable...

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/user-unable-to-see-all-forms-of-survey

kymni:
PR with fix ready

janagombitova:
All seems to work well. We will know for sure if we do not receive any support tickets with this issue. Ready for release 👍

PR #2441, opened 2018-01-18
Display copied forms for users (connect #1765)

kymni:
Copied forms we're not always displaying for all users
#### The solution
Set the ancestor IDs of copied forms before triggering the survey copy
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/app/web/rest/ActionRestService.java
+++ b/GAE/src/org/waterforpeople/mapping/app/web/rest/ActionRestService.java
@@ -360,6 +360,7 @@ public class ActionRestService {
         surveysAncestorIds.add(savedProjectCopy.getKey().getId());
 
         for (Survey sourceSurvey : sourceSurveys) {
+            sourceSurvey.setAncestorIds(surveysAncestorIds);
             SurveyDto surveyDto = new SurveyDto();
             surveyDto.setCode(sourceSurvey.getCode());
             surveyDto.setName(sourceSurvey.getName());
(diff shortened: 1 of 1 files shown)


GAE/src/org/waterforpeople/mapping/app/web/rest/ActionRestService.java
```diff
@@ -360,14 +360,14 @@ private String copyProject(Long targetId, Long folderId) {
...
         for (Survey sourceSurvey : sourceSurveys) {
+            sourceSurvey.setAncestorIds(surveysAncestorIds);
             SurveyDto surveyDto = new SurveyDto();
             surveyDto.setCode(sourceSurvey.getCode());
             surveyDto.setName(sourceSurvey.getName());
             surveyDto.setPath(projectCopy.getPath() + "/" + sourceSurvey.getName());
             surveyDto.setSurveyGroupId(savedProjectCopy.getKey().getId());
             Survey surveyCopy = SurveyUtils.copySurvey(sourceSurvey, surveyDto);
             surveyCopy.setSurveyGroupId(savedProjectCopy.getKey().getId());
-            sourceSurvey.setAncestorIds(surveysAncestorIds);
```
muloem: Actually, it may make more sense to set the ancestorIds of the `surveyCopy` so as the avoid an unnecessary step of setting the `sourceSurvey` and then copying that across.

Issue #1981, opened 2017-03-10
Creating new forms should open the new form 


janagombitova:
Salim can create multiple forms in his survey. This happens usually for monitoring cases. However, every time he clicks on `Add form` he does not land on the new form right away, but he either stays on the form he was working on first, or if he was working on a monitoring form, Flow opens the registration form.

An easy thing would be, that if Salim clicks on `Add form` we open up the new form for him to directly start editing.

![1](https://cloud.githubusercontent.com/assets/12456965/23787319/cb81e838-0571-11e7-9507-c896e06883b1.gif)

kymni:
## Test plan
* Navigate to survey editing tab
* Enable monitoring feature on survey
* Click "Add new form" button
* Newly created form should be opened

janagombitova:
@muloem comment from the PR: closing this PR without merging as this issue will be addressed under a wider refactor.

janagombitova:
Tested on uat2. Fix works well and I see forms are now ordered by creation date rather than alphabetically. Let's see what users feedback us on that change. But there is this odd behaviour happening when creating a new form. It first shows a 1st in the rom and then jumps to the end. 

![newform](https://user-images.githubusercontent.com/12456965/48346092-5ae29f00-e67a-11e8-955f-7257dbd1fadc.gif)

kymni:
Reason could be that `createDateTime` property is server generated therefore is initially `null` then assigned a value. Possible fix is to assign the property a future date that will be auto replaced with a real date by the server

kymni:
Possible to handle in this issue for this release but fix PR will go into release branch

janagombitova:
Lets move our suggestion to the next release. Created a separate issue for it here: https://github.com/akvo/akvo-flow/issues/2923

This change is ok for now.

PR #2446, opened 2018-01-22
When user clicks "Add form", display created form (connect #1981)

kymni:
When users clicked on add form in a monitoring survey, the newly created form was not immediately displayed
#### The solution
Created a "newForm" property that is checked before displaying a form
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


kymni:
Makes sense. Problem is, however, the old implementation sets selected survey as the first object each time content is loaded. We'll therefore still need to flag when a new form is created to stop the observer from changing the selected form

muloem:
Maybe we remove the observer all together. What happens if its gone?

Issue #2455, opened 2018-01-24
Indicate survey must be selected in Monitoring tab


janagombitova:
# Context
Based on the discussion here: https://github.com/akvo/akvo-flow/issues/2380#issuecomment-360069140 we have agreed to streamline how we show users that they need to select a survey/form before completing an action. 

In the Monitoring tab the user must select a survey to be able to see the data points. Currently if she does not select anything and hits 'Find' we show 'No results found'. This error message should apply to when there is no data for a survey and not when a survey is not selected. 
<img width="1122" alt="screen shot 2018-01-24 at 10 29 59" src="https://user-images.githubusercontent.com/12456965/35324635-940950ae-00f1-11e8-8353-2560731b15fb.png">

# Changes 
According to the comment linked above: 
- highlight the dropdown when the user hits Find but did not select a survey
- ensure that No results found only shows when a survey is selected that is empty, has no data submitted to it

Maseline:
Test plan
Navigate to monitoring tab
Select the find option without setting any survey, notice a red border added tothe area around the select field
Next, select all the required surveys/form and select Find, notice that results are returned and the red border is removed
In another scenario, try selecting only some of the surveys but not all, notice the red border is still added.

kymni:
For consistency's sake I suggest we consider the approach currently implemented in the data cleaning and reports exports tabs where when the user tries to perform an action without having selected a survey, they get popup prompting them to select one. What do you thing @janagombitova @bogere?

janagombitova:
@kymni based on the discussion here: https://github.com/akvo/akvo-flow/issues/2380#issuecomment-360069140 we decided to go for an in page approach that should then we reused across the other parts of Flow too

kymni:
Should we therefore tackle the dropdowns under the data cleaning and exports tabs in this issue or a separate one?

janagombitova:
That is up to you guys to decide. ;)

bogere:
@kymni  basing on this issue [#2380],  they are being handled in separate issues

bogere:
After discussion with charles, we decided to combine the related issues(#2454 , #2456 and #2380) into a single issue(#2455)

janagombitova:
@bogere and @kymni I took a quick look and I like how the highlighted box remains until the user gets to the form (or survey for Monitoring tab). But there is something odd in how the drop downs behave (resize) with the highlighted box and without. This is not a showstopper - but needs to be addressed, if not now, in the next iteration. 

[...]

PR #2472, opened 2018-02-05
Indicate survey must be selected in monitoring tab(connect #2455)

Maseline:
When the 'Find' option was selected with no survey selected, it would return No results found. 
The proposed solution was to highlight the border of  entire stretch for survey selection to cater especially for cases where final form selection went to another line
#### The solution
Created a css class for the red border effect
Added  a property, misssingSurvey that is set to true in the view function when the function dealing with Find is called but no survey is selected. missingSurvey is set back to false when the surveys are all selected. 
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/css/main.scss
+++ b/Dashboard/app/css/main.scss
@@ -3343,6 +3343,7 @@ div.chooseSurveyData {
     border-color: red;
     padding-top: 8px;
     padding-bottom: 5px;
+    width: 99% !important;
 }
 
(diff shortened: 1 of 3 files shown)


muloem:
Handled in https://github.com/akvo/akvo-flow/pull/2847

Issue #2474, opened 2018-02-06
Change how variable name validation error message is displayed


kymni:
Currently the variable name validation error message is hard-coded on the end-point. Change this so that the displayed message is pulled from transifex

kymni:
### Test plan
* Confirm that a string '_variable_name_validation_failure_reason' for handling variable name validation error message exists
* Confirm that all instances of question ID on the dashboard are now replaced by variable name

stellanl:
# More test plan:
The variable name occurs at many places. We need to make sure it still works everywhere, both old and newly set variable names.

- Test setting some and check that displaying a mix of them in the UI still works.
- Test that copying a survey copies the variable names correctly.
- Test that both old and new variable names are shown correctly in report headers.

janagombitova:
Found a problem: 

Before we had two variable name error messages:
- not unique 
- containing characters we do not support 

Now the not unique message is showing also when it should show that the variable name contains characters that we do not support.

<img width="470" alt="screen shot 2018-03-06 at 09 12 39" src="https://user-images.githubusercontent.com/12456965/37021086-9497dade-211e-11e8-92ba-738c7a95b389.png">

janagombitova:
All good now 👍

PR #2475, opened 2018-02-06
Change variable name properties and validation failure error message (connect #2474 and connect #2143)

kymni:
Variable name properties read as question ID and validation error message was hard coded
#### The solution
Rephrase properties and create UI string for validation error message
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


stellanl:
It's easy to rename the field in the backend. But what about old data?
We could
 1. Save the QuestionDto.variableName to both Question.questionId as well as the new Question.variableName. 
 2. On a fetch use variableName, and if not set, use questionId.
 3. Release.
 4. Run a data script that copies questionId to variableName.
 5. Update code (like reporting engine) that reads/writes questionId.
 6. Remove questionId field.

muloem:
@stellanl sounds like a plan.  I think we should cover all but perhaps we leave out the data migration bit.  It should be enough to always check whether there is a variableName and if not check whether there is a questionId and return it as variable name.  And in the saving always save to variableName.  At some point down the road we can run a scan on the questionId field and perhaps drop if its empty

Dashboard/app/js/lib/models/models.js
```diff
@@ -389,7 +389,7 @@ FLOW.QuestionAnswer = FLOW.BaseModel.extend({
   collectionDate: DS.attr('number'),
   surveyInstanceId: DS.attr('number'),
   iteration: DS.attr('number'),
-  questionID: DS.attr('string'),
+  questionID: DS.attr('string'), //TODO should be number?
```
muloem: Should be a number but is quite widely used so better we tackle in a separate issue.

GAE/src/com/gallatinsystems/survey/domain/Question.java
```diff
@@ -62,6 +62,7 @@
     private Long questionGroupId;
     private Long surveyId;
     private String questionId;
```
muloem: Perhaps we can mark this one as deprecated and put a note of the transformation to `variableName`.

Dashboard/app/js/lib/models/models.js
stellanl: That's why I made it a TODO, and not simply changed it :)

kymni:
#2143 was already closed as it covered the front-end bit of changing question ID to variable name. Commits for changing variables, properties, and functions have been made against #2474 which is where the test plan will be

Dashboard/app/js/lib/models/models.js
stellanl: See #2516.

Issue #2400, opened 2017-12-21
Stop splitting single select options to columns in Data analysis export


janagombitova:
# Context
Currently users can export options question answers 2 ways from Flow:

- via Data Cleaning Export: all selected options are combined in one cell and pipe separated for multiple select option questions
- via Data Analysis Export: one column with all the selected options, followed by column per option

## Example of single select option question
https://user-images.githubusercontent.com/12456965/32102822-43813052-bb1e-11e7-8dfc-3eaacaf30d66.jpg

## Example of multi select option question

https://user-images.githubusercontent.com/12456965/32102826-456a588a-bb1e-11e7-9d70-36fb70b78b23.jpg

In the research done here: akvo/akvo-product-design#273 we learned that having separate columns for option questions (multiple select) is a helpful way to structure data. This avoids Salim or Petra having to have to clean up the data on their own. We also, however, learned that having these options split for all option questions, regardless of single or multiple select, makes no sense for single select option questions.

# Ideas
## For Data Cleaning Exports
- create separate columns for each option but **only** for option that have multiple select enabled (follow the same pattern as in Data analysis export - so one column with all answers selected and then separate columns per option) 
- do not separate columns for option questions that are only single select
## For Data Analysis Exports
- keep as it is for option question with multiple select enabled
   - summary column with the question name (or variable name) and all selected answers in the cell divided by pipe
   - each column per option and indicated with 1/0 is option selected
- do not create separate columns for options where multiple select is not enabled, so for single select options

# Questions 
How will this affect data cleaning?

janagombitova:
Based on the discussion we had around ensuring column names are not duplicated (here: https://github.com/akvo/akvo-flow/issues/2315) I suggest to keep the **Data cleaning export** structure as it is. For the **Data analysis export** we already split the option questions into multiple columns. 

However, the issue is that for single select option questions the splitting of options to separate columns in not helpful and not needed. The change should then ensure that for Data analysis exports we only create the separate option columns for option questions where more than one answer is allowed (multi select option questions)

stellanl:
Currently the comprehensive report also splits options into separate columns.
(It also - like the data cleaning report - does not put variable names in the column headers.)
This could be a bug that should be fixed.

janagombitova:
@stellanl should we create an issue to how to handle the data in Comprehensive reports? I think it should have one structure or the other (Data cleaning export structure or Data analysis export structure). What are your thoughts?

stellanl:
[...]

PR #2484, opened 2018-02-13
[#2400]Only split columns if multiple options allowed. Optimise fetch of options.

stellanl:
Options were split into columns even if the re could only be one answer.
#### The solution
Prevent that.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -1108,7 +1124,7 @@ private String buildOptionString(List<Map<String, String>> optionNodes) {
         cells.add(optionString);
 
         // if needed, build cells for options
-        if (useQuestionId) {
+        if (useQuestionId && allowMultiple) { //Split options into own columns, iff multiselect
```
valllllll2000: sorry there is a typo in iff

Issue #2479, opened 2018-02-13
Indicate geoshape question must be selected in Export reports tab


Maseline:
Currently under the Export reports tab, the last step to download Geoshape data is to select a geoshape question.
When a user hits the download option without first selecting a Geoshape question, a pop up notification with a message to select a form or question is displayed.

**Changes**
In relation to issue #2380 comments:

- a red border should highlight the select question drop down when a user hits Download but did not select a question

Maseline:
Test plan
Navigate to Export reports tab, select surveys and form for Geoshape data.
Hit the download button without selecting a question.
Notice that the border color changes to red.
Select a question and hit download again.
Notice that the border color now returns to the original color

janagombitova:
Released with 1.9.38

PR #2485, opened 2018-02-14
Indicate geoshape question must be selected in exports tab(connect #2479)

Maseline:
When a user hit the Download button for geoshape Data without selecting the geoshape question, a pop up notification with a message 'select form or geoshape question' would be displayed
#### The solution
The solution was to get rid of the pop up notification when a question isn't selected and highlight the select question drop down instead.
Added a css class to change the border color of the select question drop down.
Used bindAttr to set the class of the responsible element dynamically basing on a missingQuestion property defined in the export-reports-view.js

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/css/main.scss
+++ b/Dashboard/app/css/main.scss
@@ -3337,10 +3337,8 @@ div.chooseSurveyData {
     }
 }
 
-.redBorderColor {
-    .ember-select {
-        border-color: red;
-    }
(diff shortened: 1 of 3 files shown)


muloem:

Issue #2481, opened 2018-02-13
Code cleanup: remove unused velocity templates


muloem:
There are a number of velocity templates present in the repo that are not used anywhere or that are in parts of the code that are deprecated.  We need to remove these along with the code that is not used anymore. Connected to the cleanup #555 and the the move to maven builds #2406

janagombitova:
A lot has been merged for this issue. Is this part of the release? Anything I need to test? @muloem

PR #2490, opened 2018-02-15
Remove unused velocity template and its dependent methods (connect #2481)

muloem:
* The google earth velocity template is used in code that is no longer used in our repo.

#### The solution
* Remove the template and the code that uses it

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


muloem:
Yeah. Unfortunately its used in other places that would need a more thorough cleanup so since its not the main objective of this PR we can do that in a follow up clean up.

Issue #2456, opened 2018-01-24
Indicate form must be selected in Data cleaning tab


janagombitova:
# Context
Based on the discussion here: https://github.com/akvo/akvo-flow/issues/2380#issuecomment-360069140 we have agreed to streamline how we show users that they need to select a survey/form before completing an action. 

In the Data cleaning tab tab the user must select a form to be able to export data or import data. Currently if she does not select anything and hits 'Download' we do not show anything. If she hits 'Import' without selecting a form we show her a pop up message. 

<img width="1116" alt="screen shot 2018-01-24 at 10 33 19" src="https://user-images.githubusercontent.com/12456965/35324767-06f3d094-00f2-11e8-826e-de9098209cf9.png">

# Changes 
According to the comment linked above: 
- highlight the dropdown when the user hits Download or Import but did not select a form 
- remove the pop up message currently in place

Maseline:
Test plan
Navigate to the data cleaning tab under Data.
Hit the Download option without selecting the surveys. 
A red border should appear around the survey selection area.
Select all the required surveys and form and hit the download option, the red border should be removed

muloem:
@bogere This issue is not actually handled in https://github.com/akvo/akvo-flow/pull/2871 so we have to reopen and resolve the conflicts in https://github.com/akvo/akvo-flow/pull/2859 before merging

valllllll2000:
works as described!

PR #2495, opened 2018-02-19
Indicate form must be selected in data cleaning tab (connect #2456)

Maseline:
When a user tries to download without selecting a survey, nothing happens, there is no error prompt. However, an error is logged in the console, "Cannot read property get of null"
The solution is to highlight the survey selection area by adding a red border around it when a user hits Download without selecting a survey.
#### The solution
Added a css class, dataCleaningErrorBorder in main.scss to create the red border around the survey selection area.
Added  a function in export-reports-view.js, to check if a survey is selected, this function is called in the showDataCleaningReport function
Added a div tag to enclose the survey selection elements and set the errorBorder dynamically in data-cleaning.handlebars
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/css/main.scss
+++ b/Dashboard/app/css/main.scss
@@ -3337,6 +3337,14 @@ div.chooseSurveyData {
     }
 }
 
+.dataCleaningErrorBorder {
+    padding: 8px 0 8px 0%;
+    border-style: solid;
+    border-width: 1px;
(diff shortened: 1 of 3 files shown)


muloem:

Issue #2330, opened 2017-11-22
Remove old-format data cleaning import code


stellanl:
When we the users are happy with the new data cleaning imports, as handled in issue #2295 and have no old data to upload, we can remove the parseSingleSheet() method of RawDataSpreadsheetImporter.java

stellanl:
Test plan:
Try to import a spreadsheet in the old format (or the analysis format) - should fail.

janagombitova:
It fails as it should fall. Well done @stellanl. Plus I like the error you added: Invalid RAW DATA file: DATA_CLEANING-238329118 (2).xlsx - Errors: First header cell must contain 'Metadata'. It is a good touch to inform the user what is going wrong. 👍

PR #2504, opened 2018-02-20
[#2330]Drop support for legacy format. (Connect #2330)

stellanl:
We supported a legacy format for data cleaning.
#### The solution
Simplify code by dropping that support. Generate an appropriate error message instead.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -152,25 +149,20 @@ public void executeImport(File file, String serverBase, Map<String, String> crit
                     criteria, questionIdToQuestionDto.values());
 
             List<InstanceData> instanceDataList = new ArrayList<>();
```
valllllll2000: I wonder if initializing it to empty array is necessary since the line 152 you assing a new value to it. I would just have ` List<InstanceData> instanceDataList = parseSplitSheets(....`

stellanl: Quite right, since the other fork is gone now.

Issue #2290, opened 2017-10-27
Have "other, please specify" always in a separate column 


janagombitova:
# Context 

Option questions are one of the most popular question types used. The user specifies a list of option the respondent can select from. However, in many cases when creating the form you cannot think of all options possible thus you ca select in Flow to add "other, please specify". In the app is the responded goes for "other, please specify", she write down her own answer. 

In the data exports, data cleaning exports, the "other, please specify" is treated as the form defined options and added to the answer cell. 

# Idea 
The idea is to create a separate column for "other, please specify" if selected for an option question for all export types. 

# Why 
It is confusing for the Salim or Petra analysing the data, as s/he does not know if the option selected was defined in the survey form or written by the user when answering the question. "Other, please specify" is sometimes in data analysis treated as one variable, regardless of what the users answer. Currently it is complicated to divide the give form options from the user specified ones.

janagombitova:
This change should apply on both export types, Data cleaning and Data analysis.

stellanl:
I'll assume also for comprehensive reports.

stellanl:
In the same way as #2315, the current "other" columns have a non-unique header: "--OTHER--". I propose we add that to the id or variable name to make it unique.

muloem:
> In the same way as #2315, the current "other" columns have a non-unique header: "--OTHER--". I propose we add that to the id or variable name to make it unique.

@stellanl this is for the data analysis report, right?

stellanl:
Since I have to change the import logic anyway, I was going to do it for both.

stellanl:
@janagombitova With ID I meant the numerical question id, which is put in the data cleaning headers so that the can be linked to the orrect quesiton on import.

stellanl:
One question: the usual for an option is to be put into a spreadsheet cell as code:text (if there is a code). Modern answers seem to set code OTHER for all such answers. It will look a bit repetitive with a lot of OTHER:Foo and OTHER:Bar in the "other" column. I suggest we strip those, and add them when importing.

stellanl:
Test plan:

 * Order reports in all 3 formats. They should all have a separate "other" column.
 * Make changes to the "other" column in the data cleaning report and import it. Data should change.
 * Order an analysis report on a form with two option questions that both allow other and with no variable names set; the "other" column headers should be unique.

janagombitova:
👍 works like a charm. Well done.

PR #2506, opened 2018-02-20
[#2290]Count more things. Fix shellscript wrapper. (Connect #2290)

stellanl:
Data script for checking old option data was too simple. Shell script was not the working version. 
#### The solution
Update to what I actually used.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/scripts/data/src/org/akvo/gae/remoteapi/CheckOptions.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/CheckOptions.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2014 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2018 Stichting Akvo (Akvo Foundation)
  *
  *  This file is part of Akvo FLOW.
  *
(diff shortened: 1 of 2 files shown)


scripts/data/src/org/akvo/gae/remoteapi/CheckOptions.java
```diff
@@ -47,18 +50,25 @@ public void execute(DatastoreService ds, String[] args) throws Exception {
...
+            for (Entity e : pq.asIterable(FetchOptions.Builder.withChunkSize(1000))) {
                 String val = (String) e.getProperty("value");
-                if (val.startsWith("[")) {
-                    optionsJson++;
+                if (val == null) {
+                    val = ((com.google.appengine.api.datastore.Text) e.getProperty("valueText")).getValue(); //it was too big for string
+                    optionsLong++;
+                }
+                if (val == null) {
+                    optionsNull++;
```
kymni: Can this go after L57 and we remove the condition repetition?

stellanl: Sure I can move it but the condition remains. Only if *both* fields are null should it be counted as a null value.

kymni: Makes sense

Issue #2508, opened 2018-02-22
Improve structure-check data script robustness


stellanl:
The data script would crash if a question had a null survey pointer.

janagombitova:
I noticed this has been merged recently. Is is part of the release branch and is there something I need to test here @stellanl ? Cheers

stellanl:
It will be in the release, but since it is only run by developers, I think it has been tested enough. (It stopped crashing.)

PR #2509, opened 2018-02-22
[#2508]Test for null survey pointer. (Connect #2508)

stellanl:
Data script could crash on bad db fields.
#### The solution
Test for null.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/scripts/data/src/org/akvo/gae/remoteapi/CheckSurveyStructure.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/CheckSurveyStructure.java
@@ -135,7 +135,7 @@ public class CheckSurveyStructure implements Process {
             String questionText = (String) q.getProperty("text");
             Long questionGroupSurvey = (Long) qgToSurvey.get(questionGroup);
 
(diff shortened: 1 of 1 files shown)


scripts/data/src/org/akvo/gae/remoteapi/CheckSurveyStructure.java
```diff
@@ -135,7 +135,7 @@ private void processQuestions(DatastoreService ds) {
             String questionText = (String) q.getProperty("text");
             Long questionGroupSurvey = (Long) qgToSurvey.get(questionGroup);
 
-            if (questionGroup == null || questionGroupSurvey == null) { // in no group or a nonexistent group; hopelessly lost
+            if (questionSurvey == null || questionGroup == null || questionGroupSurvey == null) { // in no survey, group or a nonexistent group; hopelessly lost
```
valllllll2000: should it be "if no survey"?

Issue #2380, opened 2017-12-08
Indicate form must be selected in Exports page 


janagombitova:
# Context 

Once a user decided she wants to export data she goes to the Export tab. She first has to select the folder > survey > form she wants to get the data out of and then she selects which export option is the best for her. 

However, if she goes directly to the export option and hits download without selecting the form, there is no indiction in the page telling her she skipped a step. 
![export](https://user-images.githubusercontent.com/12456965/33768456-64de5fd0-dc26-11e7-9477-cd0d23a61b70.gif)

# Change 

Do we need to indicate in the UI that the folder > survey > form were not selected? 

How might we indicate to the user that she needs to select a form (folder > survey > form) from the drop downs first, before she can download any export option? 

@loicsans and @Kiarii What are your thoughts?

**Side note**: if we decide to do a UI indication of some sort, then we should apply it also in the Inspect data tab, Data cleaning tab, Monitoring tab

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/issues-data-export

Kiarii:
- the Data Cleaning has a pop-up notification: "you must select a survey to run a report"
- Inspect data has an in-page message: "No results found"
- Monitoring tab has no indication

Aligning these will depend on the resources we have with potential solutions ranging (effortwise) from  
- a pop-up,
- an error indication for a required field; as in a form
- to a clear progressive (workflow), emphasizing the relation and flow of actions.. 
- we could also consider holding-in-place the survey within which a user tends to work from as a (changeable) default to reduce the need for always having to pick the survey..

janagombitova:
After a talk today with @Maseline and @Kiarii we agreed to create an error indication of the dropdown menu in this tab to show users that they must select the form before being able to export data. 

The error should appear if the user hits "Download" button in any of the provided export options but did not select in the dropdown menus a form. 

As in the dropdown we already tell the user to "Choose a ..." we will not display an error message to not duplicate the text and test out if this indication is enough or not. 
 
![image](https://user-images.githubusercontent.com/12456965/35323835-3bf18910-00ef-11e8-9d54-5e16d66507d2.png)

Maseline:
@janagombitova check this out,  you mentioned testing out if the change in color is enough for indication, this is what either of the solutions could look like
![sln1](https://user-images.githubusercontent.com/3306929/35503389-0b249254-04f1-11e8-91fc-9a4608dfa019.png)
[...]

PR #2529, opened 2018-03-06
Indicate form must be selected in exports page 2 (connect #2380)

Maseline:
Under the export reports tab, if one of the download options was selected without first selecting the survey:
-Data cleaning export: nothing would happen
-Data analysis export: a pop up dialog with the message error generating report would be displayed
-Comprehensive report: a pop up dialog with the message 'your report is being prepared' would be displayed
-Geoshape data: a pop up dialog with the message 'please select a form and a geoshape question' would be displayed
-Survey Form: a pop up dialog with the message 'you must select a survey to run a report' would be displayed
The forwarded solution was to highlight the entire survey selection area when a survey wasn't selected
#### The solution
Added a property 'downloadCall' that is set to true when any of the functions dealing with downloading reports is called 
Created an observer for downloadCall, when downloadCall is true, the value of the selectedSurvey is checked, if it is null, a property 'missingSurvey' is set to true
Missing survey is a property used to add a redError border dynamically in the handlebars  template
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/css/main.scss
+++ b/Dashboard/app/css/main.scss
@@ -3337,6 +3337,13 @@ div.chooseSurveyData {
     }
 }
 
+.redErrorBorder {
+    border-style: solid;
+    border-width: 1px;
+    border-color: red;
+    width: 98%;
+}
+
 .filterData {
     border-top: 1px solid white;
     .dataDeviceId {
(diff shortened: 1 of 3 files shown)


muloem:

Issue #2525, opened 2018-03-05
We could store *who* changed a datastore entity


stellanl:
Currently we store time of creation and time of last modification for most entities. We could add a field for which user (last) did it.
For developers, this would help with investigating support issues.
For users, displaying this would help when more than are working on a survey.

muloem:
@stellanl this field actually exists but was not actively used.  I think its a great idea indeed to store that
https://github.com/akvo/akvo-flow/blob/04d156aa376ac138e6ed833ed21fece8ae9c8159/GAE/src/com/gallatinsystems/framework/domain/BaseDomain.java#L49

janagombitova:
@stellanl I guess there is nothing here to test from the users perspective. Correct? Anything I should check elsewhere?

stellanl:
If you want to , you can see them being added to the entities in the datastore, but no, no user-visible part (yet).

janagombitova:
Hey Stellan, I checked in the data store randomly around some entities but the Create User ID has entires to it for the most recent things. However the user ID is a number. Is this expected? Is there a difference in how users (visible in the UI) are logged and in how us super users are logged? 

<img width="706" alt="screen shot 2018-03-29 at 09 06 29" src="https://user-images.githubusercontent.com/12456965/38074839-85ea7950-3330-11e8-953a-58114d3adff1.png">

stellanl:
Yes, the userIds are references into the user "table", for all kinds of users.
That's not a change, the fields have been around for years, but were not filled in.

janagombitova:
Ohh ok. Now it makes sense. Then I assume all is at it should be and we are ready to release this.

PR #2540, opened 2018-03-14
Issue/2525 log who changed entity(Connect #2525)

stellanl:
We had no idea who created/changed an entity, just when
#### The solution
Remember that
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/com/gallatinsystems/framework/dao/BaseDAO.java
+++ b/GAE/src/com/gallatinsystems/framework/dao/BaseDAO.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2010-2017 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2010-2018 Stichting Akvo (Akvo Foundation)
  *
  *  This file is part of Akvo FLOW.
  *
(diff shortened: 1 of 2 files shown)


GAE/src/com/gallatinsystems/framework/dao/BaseDAO.java
```diff
@@ -97,9 +101,12 @@ public void setDomainClass(Class<T> e) {
      */
     public <E extends BaseDomain> E save(E obj) {
         PersistenceManager pm = PersistenceFilter.getManager();
+        Long who = org.waterforpeople.mapping.app.web.CurrentUserServlet.getCurrentUserId();
         obj.setLastUpdateDateTime(new Date());
+        obj.setLastUpdateUserId(who);
```
valllllll2000: What would happen if user is null?

stellanl: Then we would store that null, which is what we have been storing since forever anyway.

Issue #2539, opened 2018-03-14
Bad filenames being entered into DeviceFileJobQueue


stellanl:
In org.waterforpeople.mapping.dao.SurveyInstanceDAO.save() the filenames added to the DeviceFileJobQueue are not extracted from the JSON package they nowadays come in; they end up looking like:

> 630e1777-d6c9-45c7-8b96-50ba17456408.jpg","location":{"accuracy":11.5,"altitude":120.5,"latitude":-16.35772,"longitude":34.6957768}}

stellanl:
This is probably contributin gto the #2636 problem, as the devices will get

1. Very long filenames that
2. Will not match any (image) files, so
3. The DeviceFileJobQueue will just grow and grow.

stellanl:
Test plan:

1. Submit a survey with an image.
2. Use the GAE console to look in the QuestionAnswerStore "table" and check that the filename is valid.

janagombitova:
@stellanl I just followed your steps in the Test plan and this is what I see in Console. Oddly the last entry is from yesterday. Is that ok? I am new to this part of Console. 

<img width="1259" alt="screen shot 2018-03-28 at 14 38 21" src="https://user-images.githubusercontent.com/12456965/38029277-b879be26-3295-11e8-8cd3-b7d7417f92d2.png">

stellanl:
No, that is a broken filename (See the "} at the end?) It's supposed to be "naked".
An image filename is checked for 3 times, at 5-minute intervals, before giving up and creating a DFJQ entry.
So it will be at least 15 mins after the upload until a DFJQ entry shows up.

janagombitova:
@stellanl I checked again this morning and my entry from yesterday does not show up in the table.

janagombitova:
@stellanl can you please do a 2nd test for me - just to be sure it works as it should? Thanks

PR #2547, opened 2018-03-19
[#2539]Extract filename from JSON object. Update copyright year. (Connect #2539)

stellanl:
Since we changed the format of the IMAGE value, bad flenames have been entered into the DeviceFileJobQueue.
#### The solution
Extract the filename using an objectMapper.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
+++ b/GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2010-2015, 2017 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2010-2015, 2017, 2018 Stichting Akvo (Akvo Foundation)
  *
  *  This file is part of Akvo FLOW.
  *
@@ -16,6 +16,7 @@
 
 package org.waterforpeople.mapping.dao;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
```diff
@@ -143,19 +146,34 @@ public SurveyInstance save(SurveyInstance si, DeviceFiles deviceFile) {
...
-                        .param(ImageCheckRequest.FILENAME_PARAM, filename)
-                        .param(ImageCheckRequest.DEVICE_ID_PARAM, deviceId)
-                        .param(ImageCheckRequest.QAS_ID_PARAM, String.valueOf(qas.getKey().getId()))
-                        .param(ImageCheckRequest.ATTEMPT_PARAM, "1");
-                queue.add(to);
+                String value = qas.getValue();
+                String filename = null;            
+                if (value.startsWith("{")) { //JSON
+                    try {
+                        Map<String,String>imgMap = objectMapper.readValue(value,
```
muloem: the object mapper is a memory sink so I was thinking that in this case, given that we dont need the whole JSON object but just the filename, could we use regex parsing instead?

GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
```diff
@@ -143,19 +146,31 @@ public SurveyInstance save(SurveyInstance si, DeviceFiles deviceFile) {
...
-                        .param(ImageCheckRequest.DEVICE_ID_PARAM, deviceId)
-                        .param(ImageCheckRequest.QAS_ID_PARAM, String.valueOf(qas.getKey().getId()))
-                        .param(ImageCheckRequest.ATTEMPT_PARAM, "1");
-                queue.add(to);
+                String value = qas.getValue();
+                String filename = null;        
+                if (value.startsWith("{")) { //JSON
+                    final String key = "\"filename\":\"";
+                    int i = value.indexOf(key);
+                    if (i >- 1) { //key found, grab all until next "
```
muloem: formatting here looks a little odd... like `>-` is an operator.

Issue #2340, opened 2017-11-23
Add filter to export only latest submission of monitoring form to data point 


janagombitova:
In this issue https://github.com/akvo/akvo-flow/issues/2339 we removed the so called **Latest data only** filter from the UI. This filter allows users for monitoring forms to only export the latest submissions to a monitoring form for each data point. 

# Changes to discuss: 
- The filter should show for Data cleaning export and for Data analysis export 
- Rename the filter to make it clear what it filters out 
- Show the filter only once a monitoring survey is selected or only once a monitoring form to a monitoring survey is selected (and not the registration from)

janagombitova:
# Context 
With monitoring our partners submit one monitoring form to a data point multiple times. Data collection when monitoring can happen constantly or in rounds (once a year all the data points are visited to capture data on the current status). 

Once users export the data, they select the monitoring form and hit export. In this case they get all the data submitted to the monitoring form and thus some submissions point to one data point. If data was collected in rounds, Salim is only interested in the newest data for each data point as he has already cleaned and analysed the older data from the previous round.

# Changes 
- once a user selects a monitoring form from a monitoring survey (not the registration form) show the option the export only the last submission for each data point 
- add this option to Data cleaning exports and to Data analysis export

stellanl:
The code for this filter is still present in the report engine, and should just need the "lastCollection" flag to be passed in from the UI.

janagombitova:
@stellanl just to confirm - what the code that is currently in the report engine does, it it exports for every data point the last submission made to the selected monitoring form, right? 

So we need a way to allow Salim to decide in the UI that we wants this "lasCollection" for Data Cleaning and Data Analysis and ensure it only shows if Salim selects a monitoring form in the dropdown.

janagombitova:
Here is the revised design provided by @Kiarii https://projects.invisionapp.com/share/HMG8IDOVCX7#/screens/283709674

We still need to decide on how to name the last submission (in designs as Recent submission) to the data points filter. 

The expected behaviour is: 
- submission date and Recent submission (or how we call it) are available filters for both Data cleaning and Data analysis exports 
- the Recent submission is disabled if the user has selected a non monitoring survey or a registration form to a monitoring survey 
- the Recent submission is only enabled for monitoring forms of a monitoring survey 
- the user can either filter the export on submission date or on recent submission, not on both

loicsans:
@stellanl I've added the radio buttons.

janagombitova:
[...]

PR #2515, opened 2018-02-28
Fixed collection date set for both export options (connect #2363)

Maseline:
#### Before the PR (what is the issue or what needed to be done)
Under the reports tab, when a user selected a date for the Data Cleaning section, the same date would show up under the data analysis section.
#### The solution
Changed from DateField to DateField2 in export-reports.handlebars in order to use an abstract function for the datepicker.
Changed from one instance of exportReportAppletView in the template to multiple instances for each report section
Added reportFromDate and reportToDate properties in exportReportAppletView and set them as the valueBinding for the date input fields  in the template
Added observers for reportFromDate and reportToDate inorder to change the min  and maxDates for the datepickers 
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

PR #2585, opened 2018-04-12
Offer option to either export latest submissions or filter by date for monitoring surveys (connect #2340)

kymni:
No option to filter only latest submissions
#### The solution
Enable option to filter by either latest submissions or by date. Option to export only latest submissions available only when a monitoring survey is selected
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


kymni:
#2515 changes restored

Dashboard/app/js/lib/views/reports/export-reports-views.js
```diff
@@ -66,7 +71,10 @@ FLOW.ReportLoader = Ember.Object.create({
       });
     }
 
-    criteria.opts.lastCollection = '' + (exportType === 'DATA_CLEANING' && FLOW.selectedControl.get('selectedSurveyGroup').get('monitoringGroup') && !!FLOW.editControl.lastCollection);
+    var exportOption = exportType === 'DATA_CLEANING' ? this.cleaningExportOption : exportType === 'DATA_ANALYSIS' ? this.analysisExportOption : "" ;
```
muloem: The nested ternary statement looks a little strange. Since there are more than two options here perhaps we can instead use an `if ... else if ... else` statement to make it more readable.

Dashboard/app/js/templates/navReports/export-reports.handlebars
```diff
@@ -16,71 +16,63 @@
...
-        evt.currentTarget.className += " active";
-       }
-    </script>
     <section class="exportContainer">
       <ul class="exportSelect">
         <li class="dataCleanExp trigger" onclick="openExportOptions(event, 'dataCleanExp_options')">
           <h2>{{t _data_cleaning_export}}</h2>
           <h6>{{t _importable_back_to_akvo_flow}}</h6>
           <p class="expDescr">{{t _combines_options}}</p>
-        {{#view FLOW.ExportReportsAppletView}}
```
muloem: These were also present in the changes made by @Maseline. please restore them.

Dashboard/app/js/templates/navReports/export-reports.handlebars
```diff
@@ -16,71 +16,63 @@
...
+              {{/if}}
             <a {{action showDataCleaningReport target="this"}} class="button trigger2">{{t _download}}</a>
           </div>
-        {{/view}}
         </li>
         <li class="dataAnalyseExp trigger" onclick="openExportOptions(event, 'dataAnalyseExp_options')">
           <h2>{{t _data_analysis_export}}</h2>
           <h6>{{t _not_importable_back}}</h6>
-          <p class="expDescr">{{t _replaces_question}}</p>  
-          {{#view FLOW.ExportReportsAppletView}}
```
muloem: were in changes made by @Maseline

Dashboard/app/js/lib/views/reports/export-reports-views.js
```diff
@@ -147,7 +175,21 @@ FLOW.ExportReportsAppletView = FLOW.View.extend({
...
+  exportOption: "range",
+  dateRangeText: Ember.String.loc('_collection_period'),
+  onlyRecentText: Ember.String.loc('_only_recent_submissions'),
+
+  dateRangeSelectedObserver: function () {
+    if (this.get("exportOption") == "range") {
+      this.set('dateRangeSelected', true);
+    } else {
+      this.set('dateRangeSelected', false);
+      FLOW.dateControl.set('fromDate', null);
```
muloem: We previously made code changes to stop using the `FLOW.dateControl` and instead use `reportFromDate` and `reportToDate` properties.

Dashboard/app/js/lib/views/reports/export-reports-views.js
```diff
@@ -187,17 +229,28 @@ FLOW.ExportReportsAppletView = FLOW.View.extend({
...
+    if (FLOW.selectedControl.selectedSurveyGroup && FLOW.selectedControl.selectedSurvey) {
+      //if not a monitoring form, export should be filtered by date
+      if (FLOW.selectedControl.selectedSurvey.get('keyId') == FLOW.selectedControl.selectedSurveyGroup.get('newLocaleSurveyId')) {
+        FLOW.ReportLoader.set('cleaningExportOption', "range");
+        $('input:radio[name=cleaning-export-option]').filter('[value=range]').prop('checked', true);
+        FLOW.ReportLoader.set('analysisExportOption', "range");
+        $('input:radio[name=analysis-export-option]').filter('[value=range]').prop('checked', true);
+      }
+    }
+    return FLOW.Env.showMonitoringFeature && FLOW.selectedControl.selectedSurveyGroup
```
muloem: We have all instances running monitoring features now so we actually don't  need to check this anymore. `FLOW.ENV.showMonitoringFeature`

kymni:
> When the last submission is selected, only the radio button becomes active. Maybe it would help to highlight the background as well, the way we highlight the background of the date filter section. And at the same time we would remove the highlighting from the date filter section to emphasise that it is no longer selected.

Should be possible given that we already have a check for what's selected

janagombitova:
> Maybe it would help to highlight the background as well, the way we highlight the background of the date filter section. And at the same time we would remove the highlighting from the date filter section to emphasise that it is no longer selected.

Great idea. 

> include a hint in the text around (most recent submissions) I think we can consider that in the next iteration. 

For now, if the option is only available for monitoring forms, that should be a strong enough hint. If we learn that users do not understand, or the support page is viewed a bit too often, we can add the hint in a following release. 

> Last thing is an idea to perhaps grey out the text "Most recent submissions" in the cases of forms where this option will not be selectable.. i.e. in non-monitoring forms

But the option is then disabled right? Because Salim selected a non-monitoring survey form, or a registration form? Do you want to add greying it out to be more visually disabled? Do I understand correctly?

muloem:
> Do you want to add greying it out to be more visually disabled? Do I understand correctly?

Exactly, for the case of non-monitoring forms, the option is disabled but visually it still looks like its possible to select it and only when you try to click on it, would you realise its disabled.  Greying it out will at least give the visual hint.  And now I realise that if we grey it out we don't need the text hint

Issue #2466, opened 2018-02-02
Add walkthrough to bulk upload page 


janagombitova:
# Context
In the light of the GDPR regulation we are improving how we handle data in Flow to ensure it is secure. We are changing how we store data on the app. This change has an affect on how users move the data from the app to their PC to upload to Flow if captured offline. For this task they use **bulk upload**. 

# Current use 
Bulk upload is done mostly by Salim and it is not a common task done. Most of the time when data is captured offline and Alphonso comes back home or to the office, he turns on wifi and the data syncs via the network to Flow. But sometimes Salim gets the data out of the device and into Flow using bulk upload. 

Here is some feedback from our support team: 
*The bulk upload feature has only been used once or twice in the past couple of years by partners in this [South Asia] region. No one has used it regularly or even themselves - I had in fact facilitated it for the partner.*

*In our region almost everybody upload data directly from the phone. Great internet coverage you. Bulk upload is only used when something goes wrong, which rarely happens* 

Because this is not a commonly done task, Salim uses flowsupport to check if he takes all the correct steps. 

# Opportunity 
With the slight change in workflow on how you get the data out of the app for bulk uploading due to the changes in the app and because bulk uploading data is not a commonly done task we decided to limit the mystery on what steps you need to take when doing this and the need to open flowsupport to read what you need to do. 

We will add a visual to show the steps you need to take in the bulk upload page. 
- The walkthrough will hold images, title and a short text to make the walkthrough as easy to understand even without reading 
- The walkthrough is located above the place where Salim drops the file 
- Salim can close the walkthrough if he wants with the close 'x' icon, but this will make the walkthrough show every time he comes back to the page 
- Salim can decide to never want to see it again, by clicking on 'Don't show again'. This will mean that next time he navigates to the page or logs into Flow the walkthrough will not show anymore. 

For more details check the designs here: https://projects.invisionapp.com/share/ZJFL5Q4VQUG#/screens/276279057

<img width="860" alt="screen shot 2018-02-02 at 07 42 39" src="https://user-images.githubusercontent.com/12456965/35720121-aef8a0de-07ec-11e8-8036-9a11af290dde.png">

# Next steps 
Test full workflow (app and Flow online space) with a few partners and internal Akvo users

janagombitova:
@loicsans I have just tested the recent changes on https://uat2.akvoflow.org/admin/ and here are my remarks: 
[...]

PR #2591, opened 2018-04-17
[#2466] optimised vertical space of the steps

loicsans:


loicsans:
After deployment on test, we decided to optimise the vertical space for smaller screen:
From left to right:
screen shot 2018-04-17 at 15 50 06

Issue #2558, opened 2018-03-27
Maven should produce a exporterapplet.jar


dlebrero:
Other projects depend on the code of akvo-flow, so we need to generate a jar that those other projects can depend on.

PR #2607, opened 2018-04-20
Issue 2558/exporterapplet jar (connect #2558)

dlebrero:
#### Before the PR

Updating flow-service and flow-api code dependencies with Flow was a manual and cumbersome task, duplicating part of the build process on those two projects

#### The solution

Flow to publish the jar with the classes and dependencies to be used by other projects.

Travis will publish to Clojars whenever the project is tagged.

The jar will contain all the Flow classes. We tried to pull just the subset of classes into a different maven module but the classes are just too entangled with the rest of the code. 

The "Kubernetes" version of flow-services is already using the jar published during testing.

## Checklist
* [x] Connect the issue
* [x] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


dlebrero:
Can we please review and merge? If we do not publish the jar it means that we cannot make any changes in the Java side of flow services or flow api, which will I am sure will cause some panic in the near future.

Issue #2617, opened 2018-04-24
New datatype and endpoint for reports


stellanl:
As mentioned in #2505, we need a new entity class and REST endpoint: /reports that will be used by

 1.  The UI to find out which reports have been created, their state and where to download the result from. (optionally?) filtered by user.
 2. The report engine to create and update the state as the generation proceeds.
 3. A (cron) background cleanup job that removes the reports when they get too old.

Child of #2505

janagombitova:
child #2505

stellanl:
The initial proposed architecture was something like this:
```
UI                 GAE                            FlowServices

Button-----------------------------------------------> /?

              /rest/reports <--POST new report state-- Clojure
                                  report state, with id ->
JS---GET--->  /rest/reports           
<- state list
              /surveyrestapi <--GET questions-------------- Java

              /databackout   <--GET answers--------------- Java

              /rest/reports/nnn  <--PUT finished report state-- Clojure
 
JS---GET--->  /rest/reports           
<- state list

JS---GET-------------------------------------------->  /.../report.xsl           
<- file
```

stellanl:
The main problem here is that FlowServices is not logged into GAE, and so uses an API key for the /surveyrestapi and /databackout endpoints.

stellanl:
General shape of the POST/PUT body:
{"report":{
"state:"IN_PROGRESS", //or FINISHED_SUCCESS or FINISHED_ERROR
 "user":"foo@example.com",
 "id":12345678, // not for POST
 "reportType":"DATA_CLEANING",
 "message": null //mostly for errors
}}

stellanl:
![sequence](https://user-images.githubusercontent.com/399534/40486341-4dbf8fc2-5f61-11e8-8b36-455ad82fe28c.png)

PR #2630, opened 2018-05-08
Issue/2617 reports class and REST endpoint (Connect #2617)

stellanl:
No way for UI to know what Reports have been run.
#### The solution
Add a new kind of entity and a rest endpoint.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+            }
+        }
+
+        response.put("meta", statusDto);
+        response.put("report", dto);
+        return response;
+    }
+
+    // find all reports belonging to the current user
+    //TODO: get an unfiltered list if superAdmin?
```
muloem: I think even the superAdmin should only get their own list of files.

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+    @RequestMapping(method = RequestMethod.GET, value = "")
+    @ResponseBody
+    public Map<String, Object> listMyReports() {
+        final Map<String, Object> response = new HashMap<String, Object>();
+        List<ReportDto> results = new ArrayList<ReportDto>();
+        List<Report> reports = reportDao.listAllByCurrentUser();
+        if (reports != null) {
+            for (Report r : reports) {
+                ReportDto dto = new ReportDto();
+                DtoMarshaller.copyToDto(r, dto);
```
muloem: Ditto re replacing with the spring `BeanUtils`.

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+    // TODO: restrict use to owner+superAdmins
+    @RequestMapping(method = RequestMethod.GET, value = "/{id}")
+    @ResponseBody
+    public Map<String, ReportDto> findReport(@PathVariable("id") Long id) {
+        final Map<String, ReportDto> response = new HashMap<String, ReportDto>();
+        Report qo = reportDao.getByKey(id);
+        ReportDto dto = null;
+        if (qo != null) {
+            dto = new ReportDto();
+            DtoMarshaller.copyToDto(qo, dto);
```
muloem: Ditto

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+        }
+        response.put("report", dto);
+        return response;
+
+    }
+
+    // delete report by id
+    @RequestMapping(method = RequestMethod.DELETE, value = "/{id}")
```
muloem: Deleting the actual file would also need to be handled but it can be in a follow up I created. https://github.com/akvo/akvo-flow/issues/2662

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+            // delete report group
+            reportDao.delete(qo);
+            statusDto.setStatus("ok");
+        }
+        response.put("meta", statusDto);
+        return response;
+    }
+
+    // update existing report
+    //TODO: only allow status changes?
```
muloem: I think this makes sense.  Only update the status 👍

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+            // if the reportDto has a key, try to get the report.
+            if (keyId != null) {
+                qo = reportDao.getByKey(keyId);
+                // if we find the report, update it's properties
+                if (qo != null) {
+                    BeanUtils.copyProperties(reportDto, qo, doNotCopy);
+                    //TODO: look up user (but why would it change?)
+                    qo = reportDao.save(qo); //Also stores lastUpdateDateTime
+                    dto = new ReportDto();
+                    DtoMarshaller.copyToDto(qo, dto);
```
muloem: Ditto BeanUtils

GAE/src/org/akvo/flow/rest/dto/ReportDto.java
```diff
@@ -0,0 +1,137 @@
...
+    private String state;
+    private String filename;
+    private String message;
+    private String user;
+    private Date startDate;
+    private Date endDate;
+    private Boolean lastCollectionOnly;
+    private Long questionId;
+    private Date createdDateTime;
+    private Date lastUpdateDateTime;
```
muloem: Do we need both dates? Ideally we only need one property in the model on the UI thats something like `generationDate` but that corresponds to the `lastUpdateDateTime` that should reflect the successful completion of the report.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,196 @@
...
+                    final String email = uDao.getByKey(r.getUser()).getEmailAddress();
+
+                    //Gleaned from export-reports-views.js
+                    ReportBody body = new ReportBody();
+                    body.criteria = new ReportCriteria();
+                    body.criteria.opts = new ReportOptions();
+                    body.criteria.appId = PropertyUtil.getProperty("appId");
+                    body.criteria.email = email;
+                    body.criteria.surveyId = r.getFormId();
+                    body.criteria.appId = PropertyUtil.getProperty("appId");
```
muloem: You already have this property set on line 117. 👍

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,196 @@
...
+                           wr.write( postData );
+                        }
+                        final int sts = con.getResponseCode();
+                        log.info(" got  " + sts);
+
+                        if (sts == 200) {
+                            //Success, we are done!
+                            return null;
+                        }
+                        //if permanent error, fail this report
```
muloem: The formatting feels a bit weird here. maybe this comment should go inside the `else if` branch?

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,196 @@
...
+                        else if ((sts % 100) == 4) { //4xx: you messed up
+                            r.setState(Report.FINISHED_ERROR);
+                            r.setMessage("Unexpected result when starting report: " + sts);
+                            rDao.save(r);
+                        }
+                        else {
```
muloem: `} else {`

GAE/war/WEB-INF/rest-servlet.xml
```diff
@@ -37,6 +37,7 @@
     <bean name="/question_groups" class="org.waterforpeople.mapping.app.web.rest.QuestionGroupRestService" lazy-init="true" />
     <bean name="/question_options" class="org.waterforpeople.mapping.app.web.rest.QuestionOptionRestService" lazy-init="true" />
     <bean name="/questions" class="org.waterforpeople.mapping.app.web.rest.QuestionRestService" lazy-init="true" />
+    <bean name="/reports" class="org.akvo.flow.rest.ReportRestService" />
```
muloem: make it a `lazy-init` class as well so that it only initialised by spring when we are actually going to use it.

GAE/war/WEB-INF/web.xml
```diff
@@ -451,17 +462,23 @@
...
-		<url-pattern>/rawdatarestapi</url-pattern>
-	</filter-mapping>
+    <filter-mapping>
+        <filter-name>restAuthFilter</filter-name>
+        <url-pattern>/rawdatarestapi</url-pattern>
+    </filter-mapping>
+    <!-- Experiment. Also protected by Spring. -->
+    <filter-mapping>
+        <filter-name>restAuthFilter</filter-name>
+        <url-pattern>/rest/reports</url-pattern>
```
muloem: Spring already secures this end point so the restAuthFilter should actually be for securing the `/app_worker/reports` endpoint

GAE/war/WEB-INF/webapp-security.xml
```diff
@@ -42,6 +42,9 @@
         <intercept-url pattern="/rest/users" access="hasRole('USER')"/>
         <intercept-url pattern="/rest/user_roles" access="hasRole('ADMIN')"/>
         <intercept-url pattern="/rest/user_auth" access="hasRole('ADMIN')"/>
+        <intercept-url pattern="/rest/reports" method="GET" access="hasRole('USER')"/>
+        <intercept-url pattern="/rest/reports/*" method="PUT" access="permitAll"/>
+        <intercept-url pattern="/rest/reports" method="POST" access="permitAll"/>
```
muloem: I believe we should simply have a generic rule that one has to have a `USER` role to be able to execute any of the method on this endpoint.

GAE/src/org/akvo/flow/domain/persistent/Report.java
```diff
@@ -0,0 +1,114 @@
...
+    public static final String FINISHED_SUCCESS = "FINISHED_SUCCESS";
+    public static final String FINISHED_ERROR = "FINISHED_ERROR";
+
+    private static final long serialVersionUID = -1;
+    private String reportType;
+    private Long formId;
+    private String state;
+    private String filename;
+    private String message;
+    private Long user;
```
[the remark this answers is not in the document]
stellanl: Because the report engine does not get authorized as a user. The user id will be 0; but since the report engine no longer creates the Report entity, it is less of a problem.

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+import org.waterforpeople.mapping.app.web.rest.dto.RestStatusDto;
+
+import com.gallatinsystems.user.dao.UserDao;
+import com.gallatinsystems.user.domain.User;
+import com.google.appengine.api.taskqueue.Queue;
+import com.google.appengine.api.taskqueue.QueueFactory;
+import com.google.appengine.api.taskqueue.TaskOptions;
+
+/*
+ * @startuml
```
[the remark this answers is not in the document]
stellanl: The Plantuml plugin in Eclipse made it very convenient. I can move it to a separate file, of course, but then I cannot see it in the editor.

GAE/src/org/akvo/flow/rest/ReportRestService.java
stellanl: ok

GAE/war/WEB-INF/rest-servlet.xml
stellanl: ok

GAE/war/WEB-INF/web.xml
stellanl: Yes, but the report engine cannot use Spring auth

muloem: I suggest that since you already created the `ReportServlet` to run the task, the report engine can also use it to update when reports are generated.

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+                r = reportDao.save(r);
+
+                //Queue it
+                Queue queue = QueueFactory.getDefaultQueue();
+                TaskOptions options = TaskOptions.Builder.withUrl("/app_worker/reportservlet")
+                        .param(TaskRequest.ACTION_PARAM, ReportTaskRequest.START_ACTION)
+                        .param(ReportTaskRequest.ID_PARAM, Long.toString(r.getKey().getId()));
+                queue.add(options); //overwrite any supplied state
+                dto = new ReportDto();
+                DtoMarshaller.copyToDto(r, dto);
```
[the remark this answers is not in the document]
stellanl: Ok.

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,217 @@
...
+
+                BeanUtils.copyProperties(reportDto, r, doNotCopy);
+                r.setUser(user.getKey().getId());
+                r.setState(Report.QUEUED);
+                // Save it, so we get an id assigned
+                r = reportDao.save(r);
+
+                //Queue it
+                Queue queue = QueueFactory.getDefaultQueue();
+                TaskOptions options = TaskOptions.Builder.withUrl("/app_worker/reportservlet")
```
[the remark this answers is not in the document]
stellanl: Ok

GAE/src/org/akvo/flow/servlet/ReportServlet.java
stellanl: yes

GAE/war/WEB-INF/web.xml
stellanl: I had been hoping to keep the complexity down. Especially since it looked like we were redoing auth soon anyway. But ok. Changing the API again.

GAE/war/WEB-INF/webapp-security.xml
stellanl: That was to let the engine do a PUT, since it is not Spring authorized.

GAE/src/org/akvo/flow/rest/dto/ReportDto.java
stellanl: I don't know which dates will be wanted in the UI so I am exposing both of them.

GAE/src/org/akvo/flow/rest/ReportRestService.java
stellanl: If we are moving the updates from the report engine to the "servlet" API then no changes will be needed through this API at all. PUT can be completely removed.

muloem: Sure. if its not needed then we get rid of it till we actually need it.

GAE/war/WEB-INF/webapp-security.xml
muloem: Understood.  We will need to have a rule `/rest/reports` rule for the UI to authorized through spring.

GAE/src/org/akvo/flow/rest/dto/ReportDto.java
```diff
@@ -0,0 +1,137 @@
...
+public class ReportDto extends BaseDto {
+
+    private static final long serialVersionUID = 3197857074377585732L;
+
+    private String reportType;
+    private Long formId;
+    private String state;
+    private String filename;
+    private String message;
+    private String user;
```
muloem: If we dont want the user to be tampered with, maybe we can leave the `user` field out of the DTO

GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -0,0 +1,212 @@
...
+ * FlowService --> UI: Report file
+ * @enduml
+ */
+
+@Controller
+@RequestMapping("/reports")
+public class ReportRestService {
+    private static final Logger log = Logger.getLogger(ReportRestService.class.getName());
+
+    private final String[] doNotCopyUser = {
```
muloem: If we leave out the `user` field from the DTO, to only allow for it to be set internally within the backend and not from any external sources, then we don't need this.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+    private static final Logger log = Logger.getLogger(ReportServlet.class.getName());
+
+    private static final long serialVersionUID = -9064136799930675167L;
+    private static final String SERVLET_URL = "/app_worker/reportservlet";
+
+    private ReportDao rDao;
+    private UserDao uDao;
+
+    class ReportOptions implements Serializable {
+        /**
```
muloem: This part looks a little strange with the empty comments.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+        public Long questionId; //only for GeoJSON
+        public Date from;
+        public Date to;
+        public Boolean lastCollection;
+        public String imgPrefix;
+        public String uploadUrl;
+    }
+
+    class ReportCriteria implements Serializable {
+        /**
```
muloem: Ditto re empty comments section

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+        private static final long serialVersionUID = 1L;
+        public ReportOptions opts;
+        public String exportType;
+        public String appId;
+        public Long surveyId;
+        public String email;
+    }
+
+    class ReportBody implements Serializable {
+        /**
```
muloem: Ditto

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+        return restRequest;
+    }
+
+    @Override
+    protected RestResponse handleRequest(RestRequest req) throws Exception  {
+        ReportTaskRequest stReq = (ReportTaskRequest) req;
+        String action = stReq.getAction();
+        Long id = stReq.getId();
+        log.info("action: " + action + " id: " + id);
+        if (action == null) {
```
muloem: should this ever happen given there is a check for a missing action parameter when initialising the `ReportTaskRequest` object?

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+            case ReportTaskRequest.START_ACTION:
+                Report r = rDao.getByKey(id);
+                if (r != null) {
+                    if (!r.getState().equals(Report.QUEUED)) {
+                        //wrong state
+                        log.warning("Cannot start report that is " + r.getState());
+                        //TODO do anything else?
+                        return null;
+                    }
+                    log.info(" ====Starting========");
```
muloem: maybe `log.debug` instead? so we don't overcrowd log files.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+
+                    //hit the services server
+                    try {
+                        final int sts = engage(r);
+                        log.info(" got  " + sts);
```
muloem: Also perhaps have it as `log.debug`?

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+
+        Queue queue = QueueFactory.getDefaultQueue();
+        TaskOptions options = TaskOptions.Builder.withUrl(SERVLET_URL)
+                .param(TaskRequest.ACTION_PARAM, ReportTaskRequest.START_ACTION)
+                .param(ReportTaskRequest.ID_PARAM, Long.toString(r.getKey().getId()))
+                .countdownMillis(Constants.TASK_RETRY_INTERVAL);
+        queue.add(options);
+    }
+
+    private int engage(Report r) throws JsonGenerationException, JsonMappingException, IOException {
```
muloem: I was trying to figure out what `engage()` was. 😄 maybe a more descriptive name?

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+                            //permanent error, fail this report
+                            r.setState(Report.FINISHED_ERROR);
+                            r.setMessage("Unexpected result when starting report: " + sts);
+                            rDao.save(r);
+                        } else {
+                            //if we get a transient error, re-queue
+                            requeueStart(r);
+                        }
+                    }
+                    catch (MalformedURLException e) {
```
muloem: Formatting for this and the next catch is off.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+                        Log.error("Bad URL");
```
muloem: Here you are using a different logger

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+        criteria.opts.exportMode = r.getReportType();
+        criteria.opts.reportId = r.getKey().getId();
+        criteria.opts.from = r.getStartDate();
+        criteria.opts.to = r.getEndDate();
+        criteria.opts.lastCollection = r.getLastCollectionOnly();
+        criteria.opts.questionId = r.getQuestionId();
+        criteria.opts.imgPrefix = PropertyUtil.getProperty("photo_url_root");
+        criteria.opts.uploadUrl = PropertyUtil.getProperty("surveyuploadurl");
+        ObjectMapper objectMapper = new ObjectMapper();
+        String crit = java.net.URLEncoder.encode(objectMapper.writeValueAsString(criteria), "ISO-8859-1");;
```
muloem: Strictly speaking no harm done but there are two `;;` at the end.

muloem: Shouldn't it be UTF-8 encoding?

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+
+        URL url = new URL(PropertyUtil.getProperty("flowServices") + "/generate?criteria=" +  crit);
+        HttpURLConnection con = (HttpURLConnection) url.openConnection();
+        con.setRequestMethod("GET");
+        log.info("Preparing to GET " + url);
```
muloem: Also `log.debug`

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+                    }
+                }
+                break;
+            case ReportTaskRequest.PROGRESS_ACTION:
+                Report r2 = rDao.getByKey(id);
+                if (r2 != null) {
+                    if (!r2.getState().equals(Report.QUEUED)
+                            && !r2.getState().equals(Report.IN_PROGRESS)) {
+                        //wrong state
+                        log.warning("Cannot set progress on report that is " + r2.getState());
```
muloem: Just a thought that for debugging the report key id may be useful to include in this log statement.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,237 @@
...
+                    }
+                    catch (IOException e) {
+                        log.warning("====IOerror: " + e);
+                        //call it a transient error, re-queue
+                        requeueStart(r);
+                    }
+                }
+                break;
+            case ReportTaskRequest.PROGRESS_ACTION:
+                Report r2 = rDao.getByKey(id);
```
muloem: in both (`switch`) cases you will need the report so maybe its worth it to `getByKeyId()` before the switch statement and use the same report object?

GAE/src/org/akvo/flow/rest/dto/ReportDto.java
```diff
@@ -0,0 +1,137 @@
...
+
+public class ReportDto extends BaseDto {
+
+    private static final long serialVersionUID = 3197857074377585732L;
+
+    private String reportType;
+    private Long formId;
+    private String state;
+    private String filename;
+    private String message;
```
muloem: Is this the same `message` that goes to the `Messages` tab?

GAE/src/org/akvo/flow/servlet/ReportServlet.java
stellanl: Oops

GAE/src/org/akvo/flow/rest/dto/ReportDto.java
stellanl: It's more details for the state, especially if the state is FINISHED_ERROR.

GAE/src/org/akvo/flow/rest/dto/ReportDto.java
stellanl: It was a long-term inclusion; if we ever wanted to show all reports to an admin, it would identify who each of them belonged too. But gone now.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
stellanl: Ok

GAE/src/org/akvo/flow/servlet/ReportServlet.java
stellanl: yes

GAE/src/org/akvo/flow/servlet/ReportServlet.java
stellanl: fixed

GAE/src/org/akvo/flow/servlet/ReportServlet.java
stellanl: fixed

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -0,0 +1,228 @@
...
+        switch (action) {
+            case ReportTaskRequest.START_ACTION:
+                if (r != null) {
+                    if (!r.getState().equals(Report.QUEUED)) {
+                        //wrong state
+                        log.warning("Cannot start report " + id + " that is " + r.getState());
+                        //TODO do anything else?
+                        return null;
+                    }
+                    log.info(" ====Starting========");
```
muloem: `log.debug`?

GAE/src/org/akvo/flow/servlet/ReportServlet.java
stellanl: Probably not.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
stellanl: oops

Issue #1558, opened 2016-01-05
View Form & View Folder permissions give you View Data by default


rumca:
I've set up a user so that they have the ``view form`` and ``view folder`` permissions. When I log in their permission set looks like:

```
 pathPermissions: {"<id>":["PROJECT_FOLDER_READ","FORM_READ"]}
```

from discussion with @janagombitova I think it's expected that this user should not be able to subsequently view data for surveys within the folder they are permissioned for, however I can navigate to the inspect data tab and view responses. 

Is it possible to just hide the "Inspect Data" and "Monitoring" tabs if the user does not have ``"DATA_READ"`` in their pathPermissions?

janagombitova:
Furthermore, the user can 'View data' with any set of permission once the role is assigned to a folder/survey/form. 

Example: created a role from Manage cascades + Manage devices and had to link it with a folder. As a result the user could view the survey in the folder (ok) but he could also view the collected data in the Data tab for this folder (not ok)

janagombitova:
To summarise, if the use does not have View data permission, she should not be able to see the Data tab's Inspect data, Monitoring, Charts, Export reports and Maps. 

In the very unlikely case that a user does not have 'view data' but has 'edit data', he should still not be able to view the above tabs. He must have 'view data' and 'edit data' to be able to edit data. 

Do you guys agree?

janagombitova:
@muloem ohh that was not my intention. I share your concern. I might not have been explicit enough in my comment, my apologies. What I mean is, if a user has no 'view data' permission at all, she should not see those tabs. But indeed if she has 'view data' for only specific folders, or specific forms, she should only be able to see those in the dropdown when viewing the data (or exporting it). 

So lets say I can 'view forms' for all folder but only 'view data' for folder Households. In the survey list I see all the folders and surveys on the instance, but in the Data tab's Inspect data, Monitoring, Charts, Export reports and Maps I can only see the folder Households in the dropdown and the surveys and forms in it. 

Does this make more sense now?

muloem:
And for that reason i am actually of the view that we shouldn't hide the tabs at all but rather show them and only list folders / surveys where the user has view data permissions.

janagombitova:
That also sounds good.

kymni:
We won't be hiding elements but rather just not displaying folders and surveys in the drop downs for which the user does not have access. How about data displayed on the maps tab?

muloem:
We will have to add that as well for the maps tab. 😓

[...]

PR #2655, opened 2018-05-21
Load data viewing elements based on permissions set (connect #1558)

kymni:
Data viewing parts of the dashboard were being loaded even when the user did not have read permissions
#### The solution
Hide elements unless the data read permission is set
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/controllers/permissions.js
+++ b/Dashboard/app/js/lib/controllers/permissions.js
@@ -268,7 +268,7 @@ FLOW.permControl = Ember.Controller.create({
       }
     }
     return false;
-  }
+  }.property()
 });
 
 
--- a/Dashboard/app/js/lib/controllers/permissions.js
+++ b/Dashboard/app/js/lib/controllers/permissions.js
@@ -259,6 +259,16 @@ FLOW.permControl = Ember.Controller.create({
       }
       return false;
   }.property(),
+
+  canReadData: function () {
+    var currentUserPermissions = FLOW.currentUser.get('pathPermissions');
+    for (key in currentUserPermissions) {
+      if (currentUserPermissions[key].indexOf("DATA_READ") > -1) {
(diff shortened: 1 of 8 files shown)


Dashboard/app/js/templates/application/navigation.handlebars
```diff
@@ -2,12 +2,12 @@
   {{#view view.NavItemView item="navSurveys" }}
     <a {{action doNavSurveys}}>{{t _surveys}}</a>
   {{/view}}
-  {{#view view.NavItemView item="navDevices" }}
 	{{#if view.showDevicesButton}}
+  {{#view view.NavItemView item="navDevices" }}
```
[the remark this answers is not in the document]
kymni: Reason for keeping this was that for users without device permissions, the view would be created but the option would not loaded leaving an awkward spacing between the surveys and data tabs

Dashboard/app/js/lib/controllers/permissions.js
```diff
@@ -259,6 +259,30 @@ FLOW.permControl = Ember.Controller.create({
       }
       return false;
   }.property(),
+
+  userCanViewData: function (formID) {
+    var permissions = this.formPermissions(formID);
+    return permissions.indexOf("DATA_READ") >= 0;
+  },
+
+  formPermissions: function (formId) {
```
muloem: if we pass in the `form` object itself instead of the `formId` only, we could skip the extra loop in the `FLOW.surveyControl` to get the `form` object.

muloem: It also means you would be able to reuse the `FLOW.permControl.permissions(form)` function to get its permissions instead of rewriting that code.

Issue #1582, opened 2016-02-19
An 'all folder' role overrules the permissions assigned to a specific folder


janagombitova:
A user has two roles assigned to him: 
1. view only role for all folders and surveys 
2. edit role for one specific folder 
However, he cannot edit in the specific folder. It seems as if the other 'read only' role in combination with all folders is overruling the edit role for a single folder. 

This issue was reported by Welthungerhilfe. As shown in the picture below, the user'e editing role did not work in the Malawi folder and he could only view. 
![image](https://cloud.githubusercontent.com/assets/12456965/13176634/03ab9a0c-d714-11e5-92c0-a60b23acf136.png)

akvotender:
Link: https://akvoo.reamaze.com/admin/conversations/update-on-flow-roles-and-permissions-2
Assignee: Emmanuel Mulo

janagombitova:
Link: https://akvoo.reamaze.com/admin/conversations/akvoflow-support-request-number-13838-from-jigmy-at-akvo-dot-org
Assignee: Jana Gombitova

Nikki-Akvo:
Link: https://akvoo.reamaze.com/admin/conversations/access-dashboard
Assignee: Nikki Sloan

kymni:
Same issue is affecting surveys, inspect data, and monitoring data tabs so fix implemented applies for them as well.

janagombitova:
Checked the cases mentioned above in the linked support issues and they are nicely resolved. Well done 👍 

<img width="1245" alt="screen shot 2018-05-25 at 09 15 18" src="https://user-images.githubusercontent.com/12456965/40531310-480d979a-5ffc-11e8-9381-4eeab4787ce4.png">

PR #2656, opened 2018-05-21
Permissions set at root level override permissions set down the nest (Connect #1582)

kymni:
Permission set at root level overriding permissions set for specific surveys and forms down the folder structure
#### The solution
When loading specific forms and surveys, check if there are specific permissions set beyond the root level
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/controllers/permissions.js
+++ b/Dashboard/app/js/lib/controllers/permissions.js
@@ -205,9 +205,13 @@ FLOW.permControl = Ember.Controller.create({
   /* takes a form (ember object) and checks with user permissions
   whether the current user has edit permissions for the form */
   canEditForm: function(form) {
-    var permissions;
+    var permissions, userPerms = FLOW.currentUser.get('pathPermissions');
     if (!Ember.none(form)) {
-      permissions = this.permissions(form);
+      if (form.get("keyId") in userPerms) {
+        permissions = userPerms[form.get("keyId")];
+      } else {
+        permissions = this.permissions(form);
+      }
     }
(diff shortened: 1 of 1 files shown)


Dashboard/app/js/lib/controllers/permissions.js
```diff
@@ -176,6 +171,15 @@ FLOW.permControl = Ember.Controller.create({
       }
     }
 
+    // return superAdmin permissions
```
muloem: I am wondering if we even need this whole section.  The idea behind it was to shortcut the whole checking of permissions for all the ancestors for superAdmins but it actually doesn't do and the permissions on the root folder i.e. with the `id === 0` will always happen

Dashboard/app/js/lib/controllers/permissions.js
```diff
@@ -195,7 +199,7 @@ FLOW.permControl = Ember.Controller.create({
   /* takes a survey (ember object) and checks whether the current user
     has edit permissions for the survey */
   canEditSurvey: function(survey) {
-    var permissions;
+    var permissions, userPerms = FLOW.currentUser.get('pathPermissions');
```
muloem: what is the `userPerms` var for? It doesn't seem to be used anywhere

Dashboard/app/js/lib/controllers/permissions.js
kymni: True, removing the redundant code

Issue #2658, opened 2018-05-23
Remove dependency with maven repo ebi.ac.uk


dlebrero:
This repo is failing right now, which is causing the flow-services and flow-api builds to fail and will cause the flow build to fail if the TravisCI cache is cleared.

It seems that the functionality imported from that repo is either not used or can be replaced with other more popular libraries

iperdomo:
Related to #555

janagombitova:
I assume there are no user facing changes to test here.

Issue #555, opened 2014-05-29
Legacy Code Cleanup


muloem:
Legacy code cleanup that is not attached to any particular feature/implementation should be monitored via this issue.

stellanl:
Marked the [gs]etSublevel[1-6] methods of SurveyedLocale as @deprecated as part of #1759

muloem:
## Things to refactor/cleanup
* [ ] Remove unused/unnecessary jar libraries
* [ ] Standardise the logging library used (#1933)

janagombitova:
@muloem I assume, that if Flow works with all regression tests, then we can consider all the changes here successful, correct? If so, then all works as expected 

How do you want to go about the check boxes in your comment above?

muloem:
Exactly most of the issues here should be covered by regression tests and its usually code that is not even executed.  We may have to use different issues to keep better track of code cleanup. and only use this issue to track the cleanup overall.

janagombitova:
👍  sounds like a plan. Or we can create an issue in the product design repository to track the clean up tasks and issue in flow to link the code changes. But as far as the changes go, Flow works fine

muloem:
## Test Plan

Any broken issues should be identified in regression testing

janagombitova:
Nothing found in regression round. We are good to go 👍

muloem:
I'll leave this open for now but create separate issues per code cleanup.

PR #2659, opened 2018-05-23
Remove dependency on `com.google.gdata.data/gdata-core` (connect #2658)

iperdomo:
* We had an implicit dependency on commons-lang/commons-lang
  via org.apache.velocity/velocity, we just made it explicit
* We're using javax.xml.bind.DatatypeConverter for Base64 encoding
  but we need to move to Base64 API on Java 8 when possible
* Fixed the @Nullable annotation to use `javax.annotation.*` and
  not a library one
* We're not using TokenUtility class
* Related #555 

## Checklist
* [x] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [x] Code formatting
* [ ] Documentation


GAE/src/com/gallatinsystems/common/util/MD5Util.java
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2010-2012 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2010-2018 Stichting Akvo (Akvo Foundation)
```
muloem: Small note: following the strict convention this would be `2010-2012,2028`

iperdomo: are we following the _strict convention_?, if so, we need to fix *lots* of copyright headers

dlebrero:
The pom has a reference to the eib.ac.uk. Could you please remove it and test that we do not have any other dependency against that repo? Maybe by removing your own local maven repo (or moving it temporarily to another path)

iperdomo:
Tested by _clearing_ the cache. Locally works as expected.

````diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 74d86a718..16a5e9e2c 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,5 +16,5 @@ services:
      - 5005:5005
     volumes:
      - .:/app/src/:delegated
-     - ~/.m2:/home/akvo/.m2:delegated
-     - ~/.m2:/root/.m2:delegated
+     - /tmp/m2:/home/akvo/.m2:delegated
+     - /tmp/m2:/root/.m2:delegated
````

GAE/src/com/gallatinsystems/common/util/MD5Util.java
iperdomo: e.g. according to `git log` this file was modified in 2011,2012,2014,2018

     git log --date=short GAE/src/com/gallatinsystems/common/util/MD5Util.java | awk '/^Date:/ {print $2}'

I guess that something similar applies to several files

muloem: This of course applies to all files but following the boy scout rule, this is the file you were working on at the moment and can actually do something about without disrupting too much else.

Issue #2661, opened 2018-05-24
Filter datapoints on the maps tab based on a selected survey


muloem:
Currently when one accesses the maps tab _all_ datapoints are retrieved from the backend.  We now intend to filter these datapoints in order to only show those for a selected survey.

janagombitova:
We allow users to filter to survey level. For monitoring surveys the data points show answers to registration forms. 

# From testing 
Filtering data points on both maps works correctly. Carto maps work correctly, also to view data submissions - the issue with clicking on the point and the points disappeared is resolved. Mapbox maps work correctly too. 

We have an issue with showing answers to new data points on the maps, but this will be handled separately.

PR #2668, opened 2018-05-29
Filter datapoints for maps backend by `surveyId` (connect #2661)

muloem:
* All datapoints returned when a request is made to the maps backend

#### The solution

* We filter datapoints based on a `surveyId` thats provided in the request.  
* When maps are accessed through the publicly visible page, no survey is provided.  Here we return all datapoints but without any details of internal identifiers and therefore a non-authorized user can only see the datapoints but cannot access the associated content. 

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/app/web/rest/PlacemarkRestService.java
+++ b/GAE/src/org/waterforpeople/mapping/app/web/rest/PlacemarkRestService.java
@@ -19,6 +19,7 @@ package org.waterforpeople.mapping.app.web.rest;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


GAE/src/org/waterforpeople/mapping/app/web/rest/PlacemarkRestService.java
```diff
@@ -50,6 +51,7 @@
     @RequestMapping(method = RequestMethod.GET, value = "")
     @ResponseBody
     public Map<String, Object> listPlaceMarks(
+            @RequestParam(value = "surveyId", defaultValue = "-1") Long surveyId, // default to non-existing surveyId
```
valllllll2000: How about to extarct this "-1" to a constant something like NON_EXISTING_SURVEY_ID and then no need for comment

PR #2669, opened 2018-05-31
Restore devices tab (connect #1558)

kymni:
Devices tab not loading because view is loaded after permissions are checked
#### The solution
Load view first then check permissions to display the link
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/templates/application/navigation.handlebars
+++ b/Dashboard/app/js/templates/application/navigation.handlebars
@@ -2,11 +2,11 @@
   {{#view view.NavItemView item="navSurveys" }}
     <a {{action doNavSurveys}}>{{t _surveys}}</a>
   {{/view}}
-	{{#if view.showDevicesButton}}
   {{#view view.NavItemView item="navDevices" }}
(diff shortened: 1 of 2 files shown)


Dashboard/app/js/templates/application/navigation.handlebars
```diff
@@ -2,11 +2,11 @@
   {{#view view.NavItemView item="navSurveys" }}
     <a {{action doNavSurveys}}>{{t _surveys}}</a>
   {{/view}}
-	{{#if view.showDevicesButton}}
   {{#view view.NavItemView item="navDevices" }}
+	{{#if view.showDevicesButton}}
     <a {{action doNavDevices}}>{{t _devices}}</a>
-	{{/view}}
   {{/if}}
+	{{/view}}
```
muloem: The indentation looks a bit weird here. can you check it? and the `{{if ..}}` statement before it.

Issue #2660, opened 2018-05-23
Importing other does not work for single select 


janagombitova:
You have an option question with three options (a, b, c) in your survey. It has 'other' enabled and you have either have it as a single select option question or multi select option question. For single select this means that you can select a, b, c or type in other. For multiple select you can select a and b and c and type in other. This is clear. 

When working with data cleaning and importing in data you add answers for the multiple select question with other enabled and the other is saved properly with import. If you have a **single** select question and you import in data (new or cleaned) and you also have other in the OTHER column (so an row does not have a or b or c only other) the other answer is not saved at import. 

On UAT1 in _1.9.31 1.9.32 .. > Import other > New form there are two option questions with other enabled: 
- the first has three options (a, b, c) - multiple select 
- the second has two options (x, y) - single select

I imported in data for this sheet. For the 'a, b, c' question each b answer also had other. This got saved correctly as it is multiple select. For the 'x, y' question I added to each third submission an other. So I skipped the x or y option in the column and left the cell empty and in the other column added the answer. These other answers were not saved. 

Here is the sheet I imported and the one I exported once the import was completed: 
[import sheet.xlsx](https://github.com/akvo/akvo-flow/files/2030903/import.sheet.xlsx)
[DATA_CLEANING-257669445 (2).xlsx](https://github.com/akvo/akvo-flow/files/2030905/DATA_CLEANING-257669445.2.xlsx)

valllllll2000:
I have tried importing Jana's file, all seems to work. I tried comparing the comprehensive report and the new one is different but I am not able to appreciate if the issue has been fixed. Please provide a simple way to test it as do not know at what to look.

valllllll2000:
This covers an edge case hard to reproduce, should be covered by regression tests

PR #2681, opened 2018-06-04
[#2660]Look for 'other' even if that is the only answer. (Connect #2660)

stellanl:
Option questions with no non-other answer were treated as empty. 
#### The solution
Check for other answer even if main cell is empty.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -84,27 +85,28 @@
...
      * @throws Exception
      */
-    public Sheet getDataSheet(File file) throws Exception {
+    public Sheet getDataSheet(File file) {
         stream = new PushbackInputStream(new FileInputStream(file));
         Workbook wb = null;
         try {
             wb = WorkbookFactory.create(stream);
         } catch (Exception e) {
+            Log.error("Workbook creation exception:" + e);
```
valllllll2000: It seems this have added a new import: org.jfree.util.Log, why not use the same log object as line 204? log.error("Workbook creation exception", e); ? Is there a reason behind using 2 different ways of logging errors?

Issue #2619, opened 2018-04-25
Show who has User admin role in User list 


janagombitova:
# Context 

The list of users is a very handy overview for User administrators. Petra need to keep an eye on who has access to their data and what access they have. If Petra works in a large organisation with a HQ and regional locations they usually have strict policies on who can add new users and what access and roles these users can have. 

For her it is also very important to know who is also a User admin (who has the role to see the User's tab, the 'Administrator - Create, Modify and Delete Users'). Sometime people who should not have User admin rights according to the organisation's policy are given this role and Petra needs to keep a tight overview of how access is distributed. 

![screen shot 2018-04-25 at 10 04 56](https://user-images.githubusercontent.com/12456965/39233269-34f53850-4870-11e8-8089-0e5530cf2dd1.png)

# Problem
For Petra to know who is a user admin, she currently, needs to open every user to check. It would be easier for her to see this directly in the UI list. 

# Opportunity
We could make use of the column in the user list that indicates if one has the API key or not. As Salim no longer needs the API key with Flow's v2 API and we will remove that functionality (Connected to https://github.com/akvo/akvo-flow/issues/2394), we can make use of the UI elements and instead of showing API key we can show who has User admin rights instead. 

The title can be: USER ADMIN and the check indication can remain. 

![screen shot 2018-04-25 at 09 59 08](https://user-images.githubusercontent.com/12456965/39233051-8cab838e-486f-11e8-9d07-b1af4bf34ad4.png)

PR #2682, opened 2018-06-04
Show who has User admin role in User list [#2619]

dlebrero:

See #2619 

#### The solution

See #2619 

Reusing same "admin?" logic as in the edit user checkbox

## Checklist
* [x] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [x] Code formatting
* [ ] Documentation


Dashboard/app/cljs/src/org/akvo/flow/dashboard/users/users_list.cljs
```diff
@@ -98,9 +98,9 @@
                            {:user user
                             :user-auth-store user-auth-store
                             :roles-store roles-store})}
-     {:title (t> _api_keys)
-      :class "text-center"
-      :component api-user-mark}
+     {:title     (t> _admin)
```
iperdomo: There is another string using the `_administrator` translation key, why do we have 2?

dlebrero: That key is:

_administrator = Administrator - Create, Modify and Delete Users

which is very long for that column.

iperdomo: You're right. Then let's _fix_ the value of the new translation key to: `User Admin` ... not all upper case. The uppercase text transformation is done in the UI via CSS.

dlebrero: done

dlebrero:
Merged as part of https://github.com/akvo/akvo-flow/pull/2689

Issue #2505, opened 2018-02-20
Exports not longer sent via email but accessible in new page  


janagombitova:
# Context 

When we did our data security audit under the work for becoming compliant with GDPR we realised that the way we make downloading captured data possible is not the best practice.  https://github.com/akvo/akvo-product-design/issues/283 We currently send data exports to users via email. As it can take quite some time (from a few minutes to up to an hour) to generate the data export file depending of the size of the data set, in the past, we decided not to have the user wait in the UI for the file to be generated, but let her freely navigate and then send her the ready file to her email. 

As we want to support our partners in becoming more responsible when handling data, we feel we should change how we make data accessible. We do not want to send data to emails but make sure you can only access it via the UI, respecting your rights and permissions. Once our users download the data to their computers it is up to them how they share it further and handle the data. 

## Current situation
Salim goes to his Flow instance. Decides he wants to work with the dataset of a particular form. He goes to the Reports tab, Export reports page and selects his form. Then he selects which export option suits him the best and hits download. He gets a message telling him that the file is being generated and that he will receive it in his email inbox. A while later (depending on the file size) he receives the email with a link. Once he clicks on the link the data file is downloaded to his computer. 

# Opportunity 

**Note:** This issue holds the entire story, if needed please split the implementation to smaller issues.

**Goal:** How might we make it possible for Salim to download a dataset, not have to wait for it to generate, and once it is ready know that it is ready and download the data from Flow's UI respecting his roles and permissions. 

We decided to create a place in Flow where Salim can get the data from. This place will be close to the current reports page as the actions are linked. Here he will see all the exports he previously created and the newest one on the top. This way he might not need to regenerate an export over and over again if he can just re-download it. He will only see the export he has created and not the exports created by his colleagues on the instance, even though he might have access to the form. We will keep exports that are up to 1 year old, once older we will remove the entry from the UI and the file too (as Salim can always re-download the data set again).

In the UI we will show:  
- Salim goes to the Reports tab and lands on this new page that will hold the generated exports and a button that brings him to the current Export reports page where he selects a form and export option to download
- 1st visit he will not have any generated exports in there, so he will see an empty screen with the prompt to export using the button above 
- To create a new export he hits the button, what brings him to the current exports page. Once he selects his form and export option and hits download, Flow brings him back to the initial page 
- Here he sees
   - the icon of the export type he selected (same as in the current Export reports page)   
   - the name of the exported file: (current naming convention) Export_type-[form ID].[filetype]
[...]

PR #2698, opened 2018-06-12
Downloadable exports UI (connect #2505)

kymni:
Exports were generated on the server then a link sent to user when report is ready
#### The solution
Move the operation to client side and have the reports downloadable from the dashboard
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/views/reports/export-reports-views.js
```diff
@@ -1,140 +1,28 @@
...
-	  },
-	  jsonpCallback: 'FLOW.ReportLoader.handleResponse',
-	  dataType: 'jsonp',
-	  timeout: this.timeout
-	});
+    newReport.set('reportType', exportType);
+    newReport.set('formId', surveyId);
+    newReport.set('filename', '');
+    newReport.set('state', 'QUEUED');
+    //newReport.set('lastCollectionOnly', ('' + (exportType === 'DATA_CLEANING' && FLOW.selectedControl.get('selectedSurveyGroup').get('monitoringGroup') && !!FLOW.editControl.lastCollection));
```
muloem: Should we remove this?

kymni: left it there as a placeholder for after we merge #2585

Dashboard/app/js/lib/views/views.js
```diff
@@ -473,6 +510,31 @@ Ember.Handlebars.registerHelper('sgName', function (property) {
   return sg && sg.get('name') || sgId;
 });
 
+Ember.Handlebars.registerHelper('surveyPath', function (property) {
```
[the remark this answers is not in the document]
kymni: True, problem however is that surveys content might not be loaded yet when the user first visits the reports list tab hence the reason for generating the path using the survey groups instead

Issue #2702, opened 2018-06-12
Remove logging of non-Spring auth when saving objects.


stellanl:
This was added when the "save who did it" code was new, but clutters the logs unnecessarily.

PR #2703, opened 2018-06-12
Issue/2702 stop logging non spring auth (Connect #2702)

stellanl:
Unnecessary logging. Warnings about unchecked casts.
#### The solution
Remove logging. Suppress warnings.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


stellanl:
I think it is because the annotation only applies to that declaration on the next line. The Eclipse quick-fix did it that way...

Issue #2719, opened 2018-06-20
Send more parameters to report engine


stellanl:
Apparently the Clojure code expects more parameters when starting a report

kymni:
Nothing to test on front-end

PR #2720, opened 2018-06-20
[#2719]Add parameters. (Connect #2719)

stellanl:
Not enough params to report engine
#### The solution
Add some
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/akvo/flow/rest/ReportRestService.java
+++ b/GAE/src/org/akvo/flow/rest/ReportRestService.java
@@ -30,7 +30,6 @@ import org.akvo.flow.rest.dto.ReportDto;
 import org.akvo.flow.rest.dto.ReportPayload;
 import org.akvo.flow.servlet.ReportServlet;
 import org.springframework.beans.BeanUtils;
-//import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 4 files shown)


GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -41,7 +45,10 @@
 @Controller
 @RequestMapping("/reports")
 public class ReportRestService {
-    private static final Logger log = Logger.getLogger(ReportRestService.class.getName());
+	@Autowired
```
muloem: Just a small note here.. if we can dig a little deeper lets try to stay away from any @Autowired spring stuff.  We removed as much as possible of the spring auto setup to improve performance. More so if the main motivation is to know which protocol is used.

GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -159,39 +163,41 @@ protected RestResponse handleRequest(RestRequest req) throws Exception  {
...
 
-    public static void queueStart(Report r) {
-        log.info("Forking to task with action START");
+    public static void queueStart(String baseUrl, Report r) {
         Queue queue = QueueFactory.getDefaultQueue();
-        TaskOptions options = TaskOptions.Builder.withUrl(SERVLET_URL)
-                .param(TaskRequest.ACTION_PARAM, ReportTaskRequest.START_ACTION)
-                .param(ReportTaskRequest.ID_PARAM, Long.toString(r.getKey().getId()));
+        TaskOptions options = getTaskOptions(baseUrl, r);
+        log.info("Forking to task with options: " + options.toString());
```
muloem: I think we should also switch to debug logging

PR #2733, opened 2018-06-26
Remove AccessPoint (connect #2511)

iperdomo:
Removes all references to `AccessPoint` class and related functionality.

## Checklist
* [x] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/com/gallatinsystems/survey/SurveyHarness.java
```diff
@@ -29,7 +29,6 @@
     public static void main(String[] args) {
```
muloem: Seems like this whole class can go as well.

iperdomo: ok, removing

PR #2736, opened 2018-06-26
Fix report link (connect #2505)

kymni:
When a report is not yet ready for download, an invalid link is loaded in place
#### The solution
Make link unclickable until a link to report is available
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/views/views.js
```diff
@@ -396,6 +396,11 @@ Ember.Handlebars.registerHelper("reportFilename", function (filename) {
   return url.split('/').pop().replace(/\s/g, '');
 });
 
+Ember.Handlebars.registerHelper("reportLink", function (filename) {
+  var url = Ember.get(this, filename);
+  return !url || url == "" ? "#" : url;
```
muloem: From what I can tell, here we will never get a `#` since the first check works if `url == undefined` or if `url === ""`

PR #2763, opened 2018-07-04
[#2505] remove hard-coded height

kymni:
Hard-coded height causing empty spaces on some tabs
#### The solution
Remove
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/css/main.scss
+++ b/Dashboard/app/css/main.scss
@@ -4020,13 +4020,9 @@ label.languageSelector:after {
 .reportsSection > div > div > section.reportTools select,
 .reportTools > div > select {}
 .dataSection {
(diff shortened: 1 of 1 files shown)


muloem:
FYI this undoes the changes made by this commit https://github.com/akvo/akvo-flow/pull/2735/files.  Doesn't that result in the footer bar overlapping the reports section again? i.e. overlapping the report types when selecting which report to generate?

stellanl:
Yes, that was a bit weird in a not-very-tall browser...

kymni:
Should probably then restore the min-height but restrict it to the exports tab

Issue #2754, opened 2018-07-03
Pre-load surveys when the dashboard loads


kymni:
At the moment we're only pre-loading surveyGroups and cascade resources when the dashboard loads. After #2505, we realised the needs to also pre-load surveys in order to be able to generate a report's path without making additional calls

kymni:
##Test plan
* Inspect the XHR requests that are executed when dashboard is first loaded
* Confirm that call to surveys end-point is one of the requests executed

valllllll2000:
All works as expected!

PR #2781, opened 2018-07-19
Preload surveys on dashboard load (connect #2754)

kymni:
Only survey groups and cascade resources pulled when dashboard is loaded
#### The solution
Add surveys to items pulled on dashboard load
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/controllers/survey-controllers.js
+++ b/Dashboard/app/js/lib/controllers/survey-controllers.js
@@ -615,6 +615,10 @@ FLOW.surveyControl = Ember.ArrayController.create({
     }
   }.observes('FLOW.selectedControl.selectedSurveyGroup'),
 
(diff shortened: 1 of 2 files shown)


Dashboard/app/js/lib/controllers/survey-controllers.js
```diff
@@ -615,6 +615,10 @@ FLOW.surveyControl = Ember.ArrayController.create({
     }
   }.observes('FLOW.selectedControl.selectedSurveyGroup'),
 
+  populateAll: function () {
+    FLOW.store.find(FLOW.Survey);
```
muloem: So you probably need to set the retrieved survey array to `content` property of this controller

Issue #2779, opened 2018-07-17
Have a timeout for report generation


stellanl:
If report generation fails for some reason, the report will stay in the list as IN_PROGRESS until deleted. A timeout of, say, an hour or a day that will set them to FINISHED_ERROR state could be implemented, perhaps as part of the cron job that deletes them after a year (#2638).

muloem:
We have the report generation task that kicks off the report generation and requeues the report if the initial generation returns any other codes besides 4xx.  Maybe we can add a limit to the retries here and say after 5 retries we consider this a completely failed report and set to `FINISHED_ERROR` state.

stellanl:
Test plan:
1. Run the local server, standard docker-compose.yml (with dummy flow-services).
2. Start an export.
3. The logs should show that the task tried to start the report 5 times, roughly 30s apart and then gave up.
4. The report list should show it as failed.

PR #2799, opened 2018-07-30
[#2779]Stop trying to start a report after 5 tries. (Connect #2779)

stellanl:
Task would try to start report forever.
#### The solution
Stop after 5 tries.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/akvo/flow/servlet/ReportServlet.java
+++ b/GAE/src/org/akvo/flow/servlet/ReportServlet.java
@@ -134,14 +134,14 @@ public class ReportServlet extends AbstractRestApiServlet {
                             rDao.save(r);
                         } else {
                             //if we get a transient error, re-queue
-                            requeueStart(stReq, r);
+                            requeueStart(stReq, r, sts);
                         }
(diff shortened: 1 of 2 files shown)


GAE/src/org/akvo/flow/servlet/ReportServlet.java
```diff
@@ -167,27 +168,36 @@ protected RestResponse handleRequest(RestRequest req) throws Exception  {
 
     public static void queueStart(String baseUrl, Report r) {
         Queue queue = QueueFactory.getDefaultQueue();
-        TaskOptions options = getTaskOptions(baseUrl, r);
+        TaskOptions options = getTaskOptions(baseUrl, 1L, r); //First time
```
valllllll2000: Why is attempt a Long and not an "int"?

stellanl: ReportTaskRequest.attempt is Long so I can handle if it is missing

Issue #2821, opened 2018-08-17
Survey page loading errors 


janagombitova:
We have received a message from a user in Haiti having issues with loading his Flow instance, the survey page. I have tested opening up larger Flow instances on my normal network connection and had the same issue. Below are the errors I see in Console. 

<img width="1164" alt="screen shot 2018-08-17 at 08 59 25" src="https://user-images.githubusercontent.com/12456965/44252331-128f1e80-a1fc-11e8-89dc-acad9498da10.png">
<img width="1162" alt="screen shot 2018-08-17 at 08 59 40" src="https://user-images.githubusercontent.com/12456965/44252333-14f17880-a1fc-11e8-8ad6-16d024792bd3.png">
<img width="1164" alt="screen shot 2018-08-17 at 08 59 55" src="https://user-images.githubusercontent.com/12456965/44252334-14f17880-a1fc-11e8-8ea8-411d3c25352c.png">
<img width="1163" alt="screen shot 2018-08-17 at 09 00 07" src="https://user-images.githubusercontent.com/12456965/44252335-158a0f00-a1fc-11e8-9500-45e16bee5352.png">
<img width="1162" alt="screen shot 2018-08-17 at 09 00 19" src="https://user-images.githubusercontent.com/12456965/44252339-17ec6900-a1fc-11e8-9e54-1bbabfa627f1.png">
<img width="1156" alt="screen shot 2018-08-17 at 09 00 28" src="https://user-images.githubusercontent.com/12456965/44252340-17ec6900-a1fc-11e8-8c33-cab0953d7424.png">
<img width="1161" alt="screen shot 2018-08-17 at 09 00 41" src="https://user-images.githubusercontent.com/12456965/44252341-1884ff80-a1fc-11e8-8682-6c2f332bd7d1.png">

kymni:
## Test plan
Confirm error no longer shows

janagombitova:
No error showing in logs on uat1

janagombitova:
As agreed after the release I tested out the loading on few larger instances and smaller ones. On these instances I did not get any errors in console and the loading spinner stopped: 
https://lifewater.akvoflow.org/admin/
https://pilots.akvoflow.org/admin/ 
https://welthungerhilfe.akvoflow.org/admin/
https://greencoffee.akvoflow.org/admin/

But on others the error shows causing loading to not stop. Here are some instances and a screen shot. Here are some examples:

https://icco.akvoflow.org/admin/ 
[...]

PR #2836, opened 2018-08-23
Filter out forms without ancestors (connect #2821)

kymni:
Orphan forms resulting in parts of dashboard breaking
#### The solution
Filter out orphan forms
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/models/models.js
+++ b/Dashboard/app/js/lib/models/models.js
@@ -129,9 +129,9 @@ FLOW.SurveyGroup = FLOW.BaseModel.extend({
 FLOW.Survey = FLOW.BaseModel.extend({
   didLoad: function () {
     // set the survey group name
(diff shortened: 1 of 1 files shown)


Dashboard/app/js/lib/models/models.js
```diff
@@ -129,7 +129,8 @@ FLOW.SurveyGroup = FLOW.BaseModel.extend({
 FLOW.Survey = FLOW.BaseModel.extend({
   didLoad: function () {
     // set the survey group name
-    if (this.get('surveyGroupId')) {
+    var ancestorIds = this.get('ancestorIds');
+    if (this.get('surveyGroupId') && ancestorIds.length > 1) {
```
valllllll2000: why not extract this.get('surveyGroupId') to a var?

Issue #2823, opened 2018-08-17
New reports page not loading for big tenants


dlebrero:
The new reports page is making a call to /rest/surveys that is timing out for big instances, as it takes more than 60 to build the response for **1.6k** surveys.

Possible causes a possible cause is that for each survey, Flow [tries to retrieve](https://github.com/akvo/akvo-flow/blob/e02b8134b95f2a664f965f8d81712e2371aa26a6/GAE/src/org/waterforpeople/mapping/app/web/rest/SurveyRestService.java#L79) a [SurveySummaryInstance](https://github.com/akvo/akvo-flow/blob/91485e66f757eb3c13dd528fdc8627c7637628b9/GAE/src/org/waterforpeople/mapping/analytics/domain/SurveyInstanceSummary.java#L32), causing more than a thousand calls to the GAE storage. This is a typical n+1 problem.

The only piece of data returned to the Dashboard is the [instanceCount](https://github.com/akvo/akvo-flow/blob/de487f485c45b8c54bfff0087eae6902ac0b82ac/Dashboard/app/js/lib/models/models.js#L154), which is not used at all.

Digging a little bit deeper, it looks like SurveySummaryInstance is just read in this REST endpoint, so the proposal would be to delete the SurveySummaryInstance and all the code around it.

### Additional work

1. Do not call /rest/surveys until the data is actually needed. 
2. Review the performance of the [dto conversion](https://github.com/akvo/akvo-flow/blob/e02b8134b95f2a664f965f8d81712e2371aa26a6/GAE/src/org/waterforpeople/mapping/app/web/rest/SurveyRestService.java#L76)

kymni:
Client-side issue was addressed under hotfix 1.9.36.2. Back-end issue should be fixed by #2841

kymni:
Surveys loading slightly faster. Loaded >900 surveys in 25s

PR #2841, opened 2018-08-27
Remove code related to instanceCount (connect #2823)

iperdomo:
We're building aggregates `SurveyInstanceSummary` entities that are
not used anywhere in the current state of the application.

Related to issue code clean up (issue #555)

## Checklist
* [x] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/waterforpeople/mapping/app/web/DataProcessorRestServlet.java
```diff
@@ -200,10 +199,6 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
         } else if (DataProcessorRequest.FIX_OPTIONS2VALUES_ACTION
                 .equalsIgnoreCase(dpReq.getAction())) {
             fixOptions2Values();
-        } else if (DataProcessorRequest.SURVEY_INSTANCE_SUMMARIZER
```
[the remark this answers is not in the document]
iperdomo: @muloem I deleted the source classes and checked where we had a broken reference. `RawDataRestServlet` doesn't do the instance count. In the case of `SurveyInstanceDAO` the update is to increment a `SurveyQuestionSummary` not a `SurveyInstanceSummary`.

My suggestion is to keep this PR related to `instanceCount`.

GAE/src/org/waterforpeople/mapping/app/web/DataProcessorRestServlet.java
```diff
@@ -230,11 +225,6 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
             if (dpReq.getSurveyInstanceId() != null) {
                 deleteSurveyResponses(dpReq.getSurveyInstanceId());
             }
-        } else if (DataProcessorRequest.SURVEY_RESPONSE_COUNT.equalsIgnoreCase(req
```
muloem: We need to restore this as well.

stellanl: Why was this? Who uses this endpoint?

PR #2847, opened 2018-09-05
Indicate survey must be selected via monitoring tab and inspect-tab(connect #2455)

bogere:
In the Monitoring  and inspect data tab,  the user must select a survey to be able to see the data points. Currently if she does not select anything and hits 'Find' we show 'No results found'. This error message should apply to when there is no data for a survey and not when a survey is not selected.

#### The solution
* highlight the dropdown when the user hits Find but did not select a survey
* ensure that No results found only shows when a survey is selected that is empty, has no data submitted to it
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/css/main.scss
+++ b/Dashboard/app/css/main.scss
@@ -3332,7 +3332,6 @@ div.chooseSurveyData {
     position: relative;
     display: block;
     margin: 0px 0 10px 0;
-    width: 100%;
     overflow: auto;
     > select {
         max-width: 20%;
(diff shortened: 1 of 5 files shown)


kymni:
The select filters are getting resized when the "missingSurvey" check is run which should not be the case. Let's try to keep the filters as they are and only surround with the outline to indicate the needed action. Also let's apply some padding so that the outline doesn't directly border the filters. 

Also if the outline is displayed then the user user selects a form, the outline should disappear without the user having to trigger it by performing an action

screen shot 2018-09-06 at 11 32 27

PR #2859, opened 2018-09-14
Indicate form must be selected via data cleaning(connect #2456)

bogere:
Based on the discussion here: https://github.com/akvo/akvo-flow/issues/2380#issuecomment-360069140 we have agreed to streamline how we show users that they need to select a survey/form before completing an action. 
In the Data cleaning tab tab the user must select a form to be able to  import data. Currently if she does not select anything and hits  'Import' without selecting a form we show her a pop up message
#### The solution
* Remove the popup message and instead the highlight the dropdown when the user does not select the survey forms first
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


muloem:

PR #2863, opened 2018-09-17
Delete data based on property

joycarpediem:
New class to delete data based on a property from the SurveyInstance
Kind

We could only delete ALL the data for a form. Now we can selectively delete based on any string property(submitter/deviceid etc) in the surveyInstance kind
#### The solution
Prepares a list of survey instance ID's based on survey ID and the other property. Uses this list to filter QAStore and other relevant kind ids and deletes them.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


muloem:
@joycarpediem Should we close this one?

muloem:
Talking to @joycarpediem he says we don’t need this for now so closing it.

Issue #2742, opened 2018-06-28
Handling option and cascade codes and values in data exports 


janagombitova:
# Context 
When exporting data we provide Salim with different data export options. One is **Data cleaning one** that is human readable and in a structure that allows Flow to import the data. The 2nd one is the **Data analysis export** that was meant for the data to be imported into another tool and used there. 

Secondly, the Salim has the option to define codes for options or cascades, meaning an option in an option question consist of code + value, for example 1:yes, 0:no. If Salim created the codes for options in option questions or for a cascade, then at data exports in both export formats he gets the code and value in one cell. 

# Problem 
We have received feedback that when importing data into SPSS Salim has to go through each column and split the text to column to divide the code from the value. Thus making our intention to make this format machine readable not fulfilled. 

## Partner feedback 

"The excel still gives the value codes and the value labels in 1 cell together. It would really help and save a lot of time for the importing of excel data into SPSS, if the value codes and the value labels would be in different cells. Otherwise we have to do column by column “text to column” in Excel to prepare the data. SPSS works with numbers, and not well with text in a cell. As you can imagine, this prep takes a lot of time – especially if you have a large questionnaire. I remember that for the baseline data for my project it took me about 5 hours to prepare the excel to be able to export it to SPSS properly, because I needed to do text to column for about 100 columns, and excel only allows you to do 1 column at a time." 

"Data analysis exports require manual deletion or replacement of text into codes to inject data into statistical package e.g. SPSS. Would prefer 2 analysis exports: pure code & pure pure text." 

# Opportunity 
We should consider: 
- either providing the option for Salim to decide before generating the file is he wants codes or values
- or by default, if you export data with Data analysis structure you only get the codes (same principle as we do with variable names vs question text) 

# Extra 
This is a different topic, but I thought might be handy to mention. Another feedback we received for the Data analysis export is if Salim can decide if he gets repeated question groups in row, sheets, or columns. 

@stellanl and @muloem thoughts?

janagombitova:
Created via Reamaze:

Link: /admin/conversations/data-analysis-export

janagombitova:
@stellanl this is the issue we discussed in our planning session about handling how we export option and cascade codes and to apply the same rules on how we handle variable names and question text

stellanl:
Estimate for these two cases: 
1. In Data analysis format, show only code if defined, otherwise show only value. Keep current handling for other formats. An hour's work for me.
[...]

PR #2915, opened 2018-11-05
[#2742]Just ouput codes for OPTION+CASCADE in analysis+comprehensive (connect #2742)

stellanl:
For cascade, **code:name** was output unless all had codes, then just **code**.
For option, **code:name** if code defined, otherwise **name**.
#### The solution
For analysis+comprehensive formats, output **code** if it exists, otherwise **name**.
For data cleaning format, as before, **code:name** if code defined, otherwise **name**.
## Checklist
* [x] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -160,6 +160,7 @@
     private boolean variableNamesInHeaders; // Also turns on splitting of answers into separate columns (options, geo, etc.) and turns off digests
     private boolean splitIntoColumns; // Turns on splitting of answers into separate columns (options, geo, etc.) and turns off digests
     private boolean separateSheetsForRepeatableGroups;
+    private boolean justCodes; //Only output the codes fro multiple-choice answers (option and cascade)
```
valllllll2000: I think there's a tipo you meant "from" and not "fro" right?

Issue #2770, opened 2018-07-08
Form dropdown on Maps


janagombitova:
# Context 
A few releases ago we implemented folder > survey filters on maps in the Flow online workspace. Salim has to first select the folder and then the survey to see data points on the map. Once he selects a data point the most recent submission will show. If the survey he selected is monitored and there is a registration and a monitoring form submitted, we show the most recent one - the monitoring form. 

# Problem
When inspecting submission on the Map Salim often wants to see the points that have the monitoring form submitted. Right now he sees all and has to click one by one to see if there is a monitoring form submission made to the point by checking the answers.

Created via Reamaze:

Link: /admin/conversations/akvoflow-support-request-number-128356-from-andrew-at-akvo-dot-org

# Opportunity
The idea is to allow Salim to filter the points on the **form** level. He would select the folder, then the survey and then the form. Then the points will show if they have submissions of that form. So if Salim selected a monitoring form and only 5 out of 10 registered data points have the monitoring forms submitted, only the 5 points will show.

If there is more than one submission made to the monitoring form, we will show the most recent submission made to the form in the side panel. In a later iteration we should see how to (and if) to show multiple monitoring form submissions.

kymni:
After assessing the feasibility with @muloem, we agreed to first tackle this for the coming iteration by enabling the form filter then either highlighting on the details panel if a clicked point does not have responses or loading the responses as normal.

A later iteration will see us do a refactor which will make it possible to filter points displayed on the map by form

janagombitova:
The form dropdown works well but there is one usability issue that makes it confusing. Once I select my data point the first thing I see is that there are no submissions while Flow is actually loading them in and then they appear in the side bar. Can we instead to a loading spinner before we confuse the user that there is no data even though there is data? 

![map](https://user-images.githubusercontent.com/12456965/48478932-05d39400-e806-11e8-8fd9-3f5b97299b06.gif)

Relates to this issue: https://github.com/akvo/akvo-flow/issues/2882

kymni:
What we can do for this release is remove the `No submissions for this...` error message as the data is being pulled unless point has no data. We can then optimise how fast data is showed on #2882

janagombitova:
Perfect!

PR #2924, opened 2018-11-14
Manage how the form filter handles submissions(connect #2770)

kymni:
`No submissions...` message displayed to user before data has completed pulling
#### The solution
Change how submissions are managed
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/controllers/maps-controllers-common.js
+++ b/Dashboard/app/js/lib/controllers/maps-controllers-common.js
@@ -152,6 +152,7 @@ FLOW.placemarkDetailController = Ember.ArrayController.create({
   dataPoint: null,
 
   dataPointCollectionDate: null,
+  noSubmissions: false,
 
   dataPointDisplayName: function () {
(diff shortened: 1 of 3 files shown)


Dashboard/app/js/lib/controllers/maps-controllers-common.js
```diff
@@ -129,6 +124,20 @@ FLOW.MapsController = Ember.ArrayController.extend({
         }
     },
 
+    clearMarker: function () {
```
valllllll2000: Whi is the function called clearMarker but then we also clear things from the sidebar?

kymni: The primary purpose of the function is to clear the highlighting added after clicking on a point on the map. We also set the marker's associated properties to null so that the next point the user clicks starts on a clean slate

Issue #2891, opened 2018-10-16
Simplify editing options to option questions 


janagombitova:
# Context

In the survey editor Petra can create an option question with any number of options the respondent then needs to choose (one or many) from. 

We know editing and creating a survey is one of the tasks Petra spends most time on, so any ways we can reduce this effort for her, the better. 

# Issue 
Currently when adding the options she can rename them, delete any and add a new option to the bottom of the list. But this means that if the order of the options is important and she needs to add an option somewhere in the middle she cannot do so. She either has to rename all previous ones, or delete the previous ones and add them again. 

![screen shot 2018-10-16 at 16 34 53](https://user-images.githubusercontent.com/12456965/47025122-32db4b00-d163-11e8-9879-ab4ad6cbee1e.png)

# How might we 
How might we make it possible for Petra to easily add another option to the list but in any position (not only to the bottom of the list), or allow her to rearrange the order of the options? 

# Created via Reamaze:

Link: /admin/conversations/akvoflow-support-request-number-156800-from-aulia-at-akvo-dot-org

stellanl:
That can be done just as for questions and groups; add an 'order' field to the data, and UI for reordering; perhaps by adding "Move Up" and "Move Down" before the "Delete".

stellanl:
Actually, the order field is already implemented and filled out for options. Only the reordering part of the UI is needed.

muloem:
The last option in an option question disappears when attempting to reorder, if there is a dependent question

![options-reorder](https://user-images.githubusercontent.com/533336/53481725-125ebe00-3a7e-11e9-8423-9715262ec701.gif)

muloem:
Works as expected! 👍

PR #2982, opened 2019-02-21
Options ordering (connect #2891)

kymni:
Changing the order of options when designing survey was not efficient
#### The solution
Make it easier to change order of options by adding links that allow you to move an option up or down
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


kymni:
addressed

PR #2986, opened 2019-02-22
Uppgrade jackson version

muloem:
#### The solution
This is part of the move to upgrade the spring library to a  more recent (and secure) version. The first step is to upgrade the version of Jackson for JSON reading and writing on which spring depends.

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/akvo/flow/domain/DataUtils.java
```diff
@@ -106,15 +103,16 @@ public static String jsonResponsesToPipeSeparated(String optionResponses) {
...
      */
     public static List<Map<String, String>> jsonStringToList(String data) {
+        FlowJsonObjectReader jsonReader = new FlowJsonObjectReader();
+        TypeReference<List<Map<String, String>>> typeReference = new TypeReference<List<Map<String, String>>>() {};
         try {
-            return JSON_OBJECT_MAPPER.readValue(data,
-                    new TypeReference<List<Map<String, String>>>() {});
+            return jsonReader.readObject(data, typeReference);
         } catch (IOException e) {
             // Data is not JSON-formatted
```
valllllll2000: I would recommend logging this to the console or this code may fail and we will never find out why, unless you are 100% sure the exception will never occur and then you should write a comment //ignored to show that we are awaire we are swallowing the exception and it is not something we forgot.

GAE/src/test/java/org/akvo/flow/util/FlowJsonObjectReaderTests.java
```diff
@@ -0,0 +1,180 @@
...
+
+        try {
+            surveyList = reader.readDtoListObject(DTO_LIST_JSON_OBJECT, typeReference);
+        } catch (IOException e) {
+            //
+        }
+
+        assertNotEquals(surveyList, null);
+        assertEquals(surveyList.size(), 1);
+        assertEquals(surveyList.get(0).getName(),"1.10.36 all questions");
```
valllllll2000: This is a minor detail but the the assertZZZ methods work like this assertZZZ(expected, actual), this is valid for the whole of your test files. Also please format the file as some "," have no space after them.

valllllll2000: https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html

muloem: 🤦‍♂️ no thats a major detail! 😛

GAE/src/org/akvo/flow/domain/DataUtils.java
muloem: I opted for ignoring the exceptions because the assertions will fail if the exception occurs anyway and we can figure out what happened

Issue #1502, opened 2015-11-24
Add JSX transpiler to dashboard build process (estimated: 1)


muloem:
As part of the introduction of react components in the dashboard, we need to be able to compile [JSX](https://facebook.github.io/jsx/) to JS code that will be embedded in different parts of the ember dashboard code. This requires including JSX transpiler tools in our current build process.

Child of #2796

kymni:
Issue with survey/form filter. Investigating

janagombitova:
With today's testing I found the following issues: 
* [ ] previewing different cascades acts out and does not load properly. But this only happens when I try to view the cascade 'France' so there might be something isolated going on. Not sure. 

<img width="1263" alt="Screen Shot 2019-03-13 at 15 42 56" src="https://user-images.githubusercontent.com/12456965/54288197-5639e300-45a7-11e9-90cc-cd2a6de70137.png">

janagombitova:
I hope this is the last issue to be found: 
* [ ] answers to repeated groups do not show in the Inspect or Monitoring tabs. The repetitions show but the answers are blank. In Maps all works fine. 

![Screen Shot 2019-03-14 at 16 06 09](https://user-images.githubusercontent.com/12456965/54367798-47206700-4673-11e9-8061-a1436f728d60.png)

kymni:
@janagombitova seems the reason this particular response does not display is because the question is type option but the values submitted are free text (maybe was initially text then changed type after submitting?). Flow therefore tries to look for text property in the response but not finding it returns null hence the blank. Issue is therefore not webpack related but we still need to catch such cases

janagombitova:
@kymni that you for checking. Thas it very odd. I will try to reproduce it on a production instance and create a separate issue. Then all webpack issues should be handled. I did not find anything more.

valllllll2000:
everything works as expected

PR #3000, opened 2019-03-07
#1502 webpack (connect #1502)

finnfiddle:
## Checklist
* [x] Connect the issue
* [x] Test plan
* [ ] Copyright header
* [x] Code formatting
* [ ] Documentation


Dashboard/.eslintrc
```diff
@@ -0,0 +1,83 @@
+{
+  "extends": "airbnb",
```
muloem: This is based on some styling from airbnb? I guess its relatively stable right? i.e. we wont have broken checks if airbnb decides to refurbish?

Dashboard/Assetfile
```diff
@@ -17,115 +17,6 @@ class LoaderFilter < WebFilters::MinispadeFilter
   end
 end
 
-class EmberAssertFilter < Filter
```
muloem: You probably need to cleanup `AssetfilePublic` the same way, right?

Dashboard/app/js/lib/controllers/general-controllers-common.js
```diff
@@ -8,19 +10,23 @@ FLOW.dashboardLanguageControl = Ember.Object.create({
...
+  },
 });
- 
- 
-FLOW.selectedControl = Ember.Controller.create({
+
+FLOW.selectedControl = Ember.Controller.create(observe({
+  'this.selectedSurveyGroup': 'deselectSurveyGroupChildren',
+  'this.selectedSurvey': 'deselectSurveyChildren',
+  // 'this.selectedSurveyOPTIONQuestions': 'OptionQuestions',
```
muloem: Do we still need this one? if not maybe better to remove completely.

Dashboard/webpack.config.dev.js
```diff
@@ -0,0 +1,170 @@
...
+    }),
+    new HtmlWebpackPlugin({
+      ...HTML_CONFIG,
+      template: 'app/public.ejs',
+      chunks: ['pub'],
+      filename: 'pub.html',
+    }),
+    new CopyPlugin([
+      { from: 'app/js/plugins', to: 'js' },
+      { from: 'app/js/templates', to: 'templates' },
```
muloem: Do we need to copy the templates? (same qn for the other config)

Dashboard/.eslintrc
finnfiddle: yes this is one of the most commonly used styleguides - if not THE most used. no need to worry

Dashboard/app/js/lib/controllers/general-controllers-common.js
finnfiddle: it was commented out before so I just left it that way

Dashboard/Assetfile
finnfiddle: yes youre right. we can do this at a later stage though because the byhbrid build is still getting the job done as it is

Dashboard/.eslintrc
finnfiddle: also the version is locked so it wont just suddenly break

Issue #2896, opened 2018-10-19
Upgrade spring libraries


muloem:
We need to upgrade the spring-framework libraries to address reported security vulnerabilities in the ones we are currently using.

https://github.com/akvo/akvo-flow/network/alert/GAE/pom.xml/org.springframework:spring-core/open

muloem:
The upgrade guide gives a list of changes that should be made.

* [Upgrading spring-security 3.x to 4.x](https://docs.spring.io/spring-security/site/migrate/current/3-to-4/html5/migrate-3-to-4-xml.html)
  * [ ] Change initialization for `accessDecisionManager`
  * [ ]

janagombitova:
@muloem I guess we have keep this one open for the remaining part of the implementation, right? Or do you want to create a new issue?

muloem:
We can keep it open for this round.

PR #3001, opened 2019-03-08
Upgrade spring library part 2 (connect #2896)

muloem:
#### The solution

* Upgrade the spring and spring-security library versions
* Make necessary changes related to migration of the two versions

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/com/gallatinsystems/user/app/gwt/client/PermissionConstants.java
+++ b/GAE/src/com/gallatinsystems/user/app/gwt/client/PermissionConstants.java
@@ -37,6 +37,6 @@ public class PermissionConstants {
     public static final String EDIT_TOKENS = "EDIT TOKENS";
     public static final String VIEW_MESSAGES = "VIEW MESSAGES";
     public static final String EDIT_IMMUTABLITY = "EDIT IMMUTABILITY";
-    public static final String ROLE_ADMIN = "ROLE_ADMIN";
-    public static final String SUPER_ADMIN = "ROLE_SUPER_ADMIN";
+    public static final String ADMIN = "ADMIN";
+    public static final String SUPER_ADMIN = "SUPER_ADMIN";
 }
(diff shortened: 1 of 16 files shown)


GAE/src/com/gallatinsystems/user/app/gwt/client/PermissionConstants.java
```diff
@@ -37,6 +37,6 @@
     public static final String EDIT_TOKENS = "EDIT TOKENS";
     public static final String VIEW_MESSAGES = "VIEW MESSAGES";
     public static final String EDIT_IMMUTABLITY = "EDIT IMMUTABILITY";
-    public static final String ADMIN = "ADMIN";
-    public static final String SUPER_ADMIN = "SUPER_ADMIN";
+    public static final String ROLE_ADMIN = "ROLE_ADMIN";
```
valllllll2000: Why one is called ROLE_ADMIN? and the other SUPER_ADMIN, should it be called ROLE_SUPER_ADMIN?

stellanl: No, strangely super-admin is not a role.

muloem:
The ROLE_SUPER_ADMIN is not used anywhere so I removed it completely otherwise yes I would have had to rename as you say.

stellanl:
I still see ROLE_SUPER_ADMIN in the code.

muloem:
> I still see ROLE_SUPER_ADMIN in the code.
In UserRestService.java, for example.

@stellanl yes but thats `AppRole.ROLE_SUPER_ADMIN` and not `PermissionConstants.ROLE_SUPER_ADMIN`

muloem:
@stellanl @valllllll2000 so after some digging around turns out that we may not even need that PermissionConstants class.  It is pretty much not used anywhere.  The field `UserDto.admin` which is affected by that code is never passed to the UI nor to the backend.  I will make a follow up PR to remove it since I dont want to include that as part of this.

Issue #2802, opened 2018-08-02
Group copy never ends dialog


valllllll2000:
Copying a group the "copying" dialog stays on forever. If we leave the screen and come back the dialog is gone and copy completes successfully. This is more of a cosmetic issue.

valllllll2000:
![screenshot from 2018-08-02 11-24-05](https://user-images.githubusercontent.com/923280/43575862-2abc7756-9648-11e8-893b-d2e51f84b049.png)

valllllll2000:
Also happens in production

kymni:
## Test plan
* Copy a questions group
* Checks for change of state from `COPYING` to `READY` should run every 5s during which a `Copying` animation on the copied question group should display then disappear when the questions group is ready

kymni:
Status never seems to change from `COPYING` to `READY` and inspecting logs reveals that request never completes

kymni:
Querying a copied qg by its ID to identify if its status has changed always returns status as `COPYING` so will have to query all qgs in survey and pull individual qg's status from that

kymni:
Guess this is done

PR #3008, opened 2019-03-14
Issue/2802 endless group copying (connect #2802)

kymni:
Status not changing when querying if a copied question group is ready resulting in ndless copying animation
#### The solution
Check status by querying all question groups in survey to force check
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


kymni:
@muloem addressed

Dashboard/app/js/lib/views/surveys/survey-details-views.js
```diff
@@ -493,51 +493,28 @@ FLOW.QuestionGroupItemView = FLOW.View.extend({
...
-        if (!this.get('qgCheckScheduled')) {
-          this.set('qgCheckScheduled', true);
           qgQuery = setInterval(function () {
               // if the question group has a keyId, we can start polling it remotely
               if (self.content && self.content.get('keyId')) {
+                // we have an id and can start polling remotely
                 if (self.content.get('status') == "READY") {
-                  self.set('qgCheckScheduled', false);
+                  //load new group's questions and clear interval
+                  FLOW.store.findQuery(FLOW.Question, {
```
muloem: One thing I am wondering about here.. The question group polling starts when a question group has just been created, no?  If that's the only case where polling is done, then it should not yet have questions added to it, so I was just wondering how come we are trying to get its questions.

kymni: When we copy a question group, the newly created group's status is set to copying which the `pollQuestionGroupStatus` function is set to observe. We when set up an interval to get all question groups every 5s and check if the copied question group's status has changed to `READY` in which case we pull its questions and clear the interval. So we only pull a question group's questions once the group is ready

Issue #2694, opened 2018-06-08
Stop using backends on GAE


muloem:
[Backends](https://cloud.google.com/appengine/docs/standard/java/backends/) is a deprecated feature that we currently use as a second copy of our application with a different scaling configuration.  However, we use it in  a few select cases that can be optimised such that we do not need an instance with more processing power and longer execution times.

Parent of #2968 
Parent of #2969
Parent of #2970 
Parent of #2971
Parent of #2972

stellanl:
Backends are mentioned in build.xml for update and rollback targets.

It is configured in WEB-INF/backends.xml :
<backends>
	<backend name="dataprocessor">
		<class>B2</class>
		<instances>1</instances>
		<options>
			<dynamic>true</dynamic>
		</options>
	</backend>
</backends>

stellanl:
Backends are called from a number of classes.
The URLs are:
com.gallatinsystems.survey.dao.CascadeNodeDao.deleteRecursive(Long, Long)  and
com.gallatinsystems.survey.dao.CascadeResourceDao.delete(CascadeResource) :
 /app_worker/dataprocessor?action=deleteCascadeNodes

com.gallatinsystems.survey.dao.SurveyUtils.copySurvey(Survey, SurveyDto) :
 /app_worker/dataprocessor?action=copySurvey

org.waterforpeople.mapping.app.web.rest.QuestionGroupRestService.copyGroup(QuestionGroupDto)
 /app_worker/dataprocessor?action=copyGroup

org.waterforpeople.mapping.app.web.RawDataRestServlet.handleRequest(RestRequest) L375
 /app_worker/dataprocessor?action=rebuildQuestionSummary (IF backendpublish property set)

org.waterforpeople.mapping.app.web.SurveyAssemblyServlet.handleRequest(RestRequest)
  /app_worker/dataprocessor?action=assembleSurvey (IF backendpublish property set AND >80 questions)

stellanl:
Interestingly, the importRemoteSurvey action takes a very long time (I find several examples very near the 600s/10 minutes allowed on a task queue on uat1)  but is not sent to a backend by 
org.waterforpeople.mapping.app.web.TestHarnessServlet.doGet(), just to the default task queue.

stellanl:
To eat this elephant we will divide it into smaller piece; creating separate issues for them.

PR #3014, opened 2019-03-18
Remove backends deployment configuration and deployment code (connect #2694)

muloem:
#### The solution
* We will no longer be using the backends so we remove the configuration and deployment code

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


stellanl:
Maybe it should go from build.xml, too

muloem:
We never use the build.xml anymore right? Maybe we leave it in there and clear out the entire build file in a different issue?

Issue #3034, opened 2019-03-29
remove ruby build for public site


finnfiddle:
currently the `AssetfilePublic` AND webpack are overlapping certain build tasks - and we need to remove the former

PR #3045, opened 2019-04-03
remove ruby build for public site #3034

finnfiddle:
Webpack and rake were running in parallel to build the public site static assets. 
#### The solution
Needed to remove the ruby element.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


finnfiddle:
I know theres a lot going on here. I had to completely rework the webpack setup. Happy to explain when you have time

finnfiddle:
@muloem the minified bundle is just less than 1MB. Normally, this is an acceptable for people with a normal connection. I dont think we will be able to bring it down much more than this but what we can look at is splitting the bundle up into smaller bundles - one for each page for example

PR #3071, opened 2019-04-16
Issue/3020 fix exports with group 1 repeatable

stellanl:
Forms w all repeatable groups and responses with >100 repetitions resulted in broken reports.
#### The solution
Handle it, by writing one row at a time.
## Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
@@ -19,6 +19,7 @@ package org.waterforpeople.mapping.dataexport;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
@@ -39,6 +40,8 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+
+import javax.annotation.Nonnull;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import org.akvo.flow.domain.DataUtils;
@@ -76,7 +79,9 @@ import org.waterforpeople.mapping.domain.CaddisflyResult;
 import org.waterforpeople.mapping.domain.response.value.Media;
 import org.waterforpeople.mapping.serialization.response.MediaResponse;
 import static org.waterforpeople.mapping.dataexport.ExportImportConstants.*;
+import static org.waterforpeople.mapping.dataexport.ExportImportUtils.parseCellAsString;
 
+import com.gallatinsystems.common.util.StringUtil;
 import com.gallatinsystems.survey.dao.CaddisflyResourceDao;
 
 import static com.gallatinsystems.common.Constants.*;
@@ -85,6 +90,10 @@ import static com.gallatinsystems.common.Constants.*;
  * Enhancement of the SurveySummaryExporter to support writing to Excel and including chart images.
  *
  * @author Christopher Fagiani
+ */
+/**
+ * @author stellan
+ *
  */
 public class GraphicalSurveySummaryExporter extends SurveySummaryExporter {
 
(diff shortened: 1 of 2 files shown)


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -529,80 +532,96 @@ private void analyseData(final List<InstanceData> allData) {
...
+        createCell(r, col++, dto.getKeyId().toString());
+        createCell(r, col++, ExportImportUtils.formatDateTime(dto.getCollectionDate()));
+        createCell(r, col++, sanitize(dto.getSubmitterName()));
+        String duration = getDurationText(dto.getSurveyalTime());
+        createCell(r, col++, duration);
+        Double v = dto.getFormVersion();
+        if (v != null) {
+            createCell(r, col++, v);
+        }
+        //TODO else a cell with 0.0??
```
muloem: Is this still needed?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -654,18 +689,21 @@ private synchronized int writeInstanceData(
         }
 
         // Calculate the digest
+        /* only for split sheets
+        //TODO this may go outside the window!
```
muloem: Also still needed?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -529,80 +532,96 @@ private void analyseData(final List<InstanceData> allData) {
...
+            Row iterationRow = getRow(startRow + i, sheet);
+            writeMetadataRow(iterationRow, instanceData, i + 1, groupSheet);
+            for (QuestionDto qd : whichQuestions) {
+                final Long questionId = qd.getKeyId();
+                final QuestionDto questionDto = questionsById.get(questionId);
+                SortedMap<Long, String> iterationsMap = instanceData.responseMap.get(questionId);
+                if (iterationsMap == null) {
+                    continue; //No answers on this sheet
+                }
+                String val = iterationsMap.get(new Long(i));
```
muloem: It feels weird to have a `new Long()` here.  Maybe its worth it converting that and related code to an `int`?

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
stellanl: No

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
stellanl: Might be. We have not had any feedback on the form version column.

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
stellanl: Then we cannot use a Map, it depends on the mapping being Object->Object

GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -529,80 +532,96 @@ private void analyseData(final List<InstanceData> allData) {
...
-                writeAnswer(iterationRow, columnIndexMap.get(questionId.toString()),
-                        questionDto, val);
-                if (!digestRows.contains(iterationRow)) { //A Set would be neater
-                    digestRows.add(iterationRow);
+        for (int i = 0; i < maxIter; i++) {
+            Row iterationRow = getRow(startRow + i, sheet);
+            writeMetadataRow(iterationRow, instanceData, i + 1, groupSheet);
+            for (QuestionDto qd : whichQuestions) {
+                final Long questionId = qd.getKeyId();
+                final QuestionDto questionDto = questionsById.get(questionId);
```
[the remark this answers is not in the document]
stellanl: Ah, cut-and-paste consequence

stellanl:
The starting row problem was that when there were NO questions on the base sheet, there was nothing entered for that instance. Now the code always puts in a row for each surveyInstance, even if it will be only metadata.

PR #3072, opened 2019-04-17
Issue/3015 fix linting

finnfiddle:
Client-side code was not being linted - ie no styleguide enforced, syntax errors slipping through.
#### The solution
Enforce eslint in the build and in a pre-commit git hook.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/controllers/data-controllers.js
```diff
@@ -293,85 +311,90 @@ FLOW.questionAnswerControl = Ember.ArrayController.create({
...
-			}
-			allResponses.push(groupResponses);
-			groupResponses = [];
-			answersInGroup = [];
-		}
-		return Ember.A(allResponses);
-	  }
-	return content;
-  }),
+    'FLOW.questionContol.content.isLoaded',
```
muloem: I think we can leave this one out since its duplicated.
```suggestion
```

Dashboard/app/js/lib/controllers/permissions.js
```diff
@@ -36,106 +37,106 @@ FLOW.permControl = Ember.Controller.create({
...
     }
-    return permissions && permissions.indexOf("DATA_UPDATE") > -1;
+    return permissions && permissions.indexOf('DATA_UPDATE') > -1;
   },
 
-  canManageDevices: Ember.computed(function () {
-    var currentUserPermissions = FLOW.currentUser.get('pathPermissions');
-    for (var perms in currentUserPermissions) {
-      if (currentUserPermissions[perms].indexOf("DEVICE_MANAGE") > -1) {
+  canManageDevices: Ember.computed(() => {
```
muloem: In some cases I see `function () {}` replaced like here and in other cases not...

finnfiddle: @muloem can you show me an example?

Dashboard/app/js/lib/controllers/survey-controllers.js
```diff
@@ -224,180 +218,170 @@ FLOW.projectControl = Ember.ArrayController.create({
...
+    });
+
     return false;
   },
 
   /* Computed properties */
-  breadCrumbs: Ember.computed(function() {
-    var result = [];
-    var currentProject = this.get('currentProject');
+  breadCrumbs: Ember.computed(function () {
```
muloem: @finnfiddle like here..

finnfiddle: @muloem so the answer is that it depends whether `this` has been used inside the function. Inside normal es5 functions `this` refers to the function itself, but in arrow functions `this` refers to the context that the function was defined in. Eg:
```
var self = this;
function oldSchool() {
  assert(this === self, false);
}
const newSchool = () => {
  assert(this === self, true);
}
```
Its called lexical scoping if you want to read up on it: https://hackernoon.com/javascript-es6-arrow-functions-and-lexical-this-f2a3e2a5e8c4

So Eslint has automatically converted the ones it can to arrow functions and the other contain `this` it has left as is.

finnfiddle: can I resolve @muloem?

Issue #2995, opened 2019-03-01
Data cleaning sheets with only repeat question groups do not update the responses


muloem:
When uploading a data cleaning sheet that only contains data from a repeat question group and no responses for the normal group(s) or a file with with only repeat question group sheet, the responses edited for data cleaning in the repeating group are never updated on the backend. 

**First sheet is empty**
<img width="1052" alt="screenshot 2019-03-01 at 12 14 09" src="https://user-images.githubusercontent.com/533336/53634849-9d28ef80-3c1b-11e9-8eb2-1a1f7a17359f.png">

**Data is only present for repeat question group**
<img width="1379" alt="screenshot 2019-03-01 at 12 14 25" src="https://user-images.githubusercontent.com/533336/53634863-ae71fc00-3c1b-11e9-9d18-d842242053bc.png">

stellanl:
Test plan:

1. Import data to a form with only one group which is repeatable.
2. Export it again and see that it was entered correctly.

PR #3074, opened 2019-04-24
[#2995]Fix header validation for all-repeatable forms.

stellanl:
Forms with the first or all groups repeatable were not imported correctly
#### The solution
Handle that, chiefly by changing the verify() method.
## Checklist
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
@@ -68,11 +68,6 @@ public class RawDataSpreadsheetImporter implements DataImporter {
     private BlockingQueue<Runnable> jobQueue;
     private List<String> errorIds;
     private static final FlowJsonObjectWriter JSON_OBJECT_WRITER = new FlowJsonObjectWriter();
-    private static final int LEGACY_MONITORING_FORMAT = 6;
-    private static final int MONITORING_FORMAT_WITH_DEVICE_ID_COLUMN = 7;
-    private static final int MONITORING_FORMAT_WITH_REPEAT_COLUMN = 8;
-    private static final int MONITORING_FORMAT_WITH_APPROVAL_COLUMN = 9;
-    private static final int MONITORING_FORMAT_WITH_FORM_VERSION = 10;
 
     private boolean otherValuesInSeparateColumns = false; //until we find one
 
@@ -139,7 +134,7 @@ public class RawDataSpreadsheetImporter implements DataImporter {
             for (int i = 0; i < wb.getNumberOfSheets(); i++) {
                 Sheet sheet = wb.getSheetAt(i);
                 String sn = sheet.getSheetName();
-                if (i == 0 || sn.startsWith("Group ")) {
+                if (i == 0 || sn != null && sn.matches(GROUP_DATA_SHEET_PATTERN)) {
                     sheetMap.put(sheet, processHeader(sheet, headerRowIndex));
                     otherValuesInSeparateColumns |= separatedOtherValues(sheet, headerRowIndex);
                 }
@@ -241,7 +236,7 @@ public class RawDataSpreadsheetImporter implements DataImporter {
 
         //these are all for the base sheet
         Map<Integer, Long> columnIndexToQuestionId = sheetMap.get(baseSheet);
-        int firstQuestionColumnIndex = 0;
+        int firstQuestionColumnIndex = -1;
         if (columnIndexToQuestionId.isEmpty()) { //Nothing but metadata
             firstQuestionColumnIndex = md5Column;
         } else {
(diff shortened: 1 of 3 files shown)


GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -1028,13 +1034,14 @@ protected void invokeUrl(String serverBase, String urlString,
 
             Row headerRow = sheet.getRow(headerRowIndex);
             boolean firstQuestionFound = false;
-            int firstQuestionColumnIndex = 0;
+            int firstNonheaderColumnIndex = 0;
```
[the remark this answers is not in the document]
stellanl: Yes, but since there may be 0 question columns,  I thought the original name was misleading.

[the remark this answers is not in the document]
stellanl: But sure, *header* might be a little misleading too.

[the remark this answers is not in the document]
muloem: More than _header_ I found it a little weird processing the negative i.e. the _first none header_ vs _first question_

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -240,8 +241,13 @@ public void executeImport(File file, String serverBase, Map<String, String> crit
 
         //these are all for the base sheet
         Map<Integer, Long> columnIndexToQuestionId = sheetMap.get(baseSheet);
-        int firstQuestionColumnIndex = Collections.min(columnIndexToQuestionId.keySet());
-        Map<String, Integer> metadataColumnHeaderIndex = getMetadataColumnIndex(baseSheet, firstQuestionColumnIndex, headerRowIndex, false);
+        int firstQuestionColumnIndex = 0;
```
muloem: I propose we initialize it to `-1` to indicate that no question column has been found rather than defaulting to the first column. 

```suggestion
        int firstQuestionColumnIndex = -1;
```

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
muloem: I propose we initialize it to `-1` to indicate that no question column has been found rather than defaulting to the first column

```suggestion
            int firstNonheaderColumnIndex = -1;
```

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -1028,13 +1034,14 @@ protected void invokeUrl(String serverBase, String urlString,
 
             Row headerRow = sheet.getRow(headerRowIndex);
             boolean firstQuestionFound = false;
-            int firstQuestionColumnIndex = 0;
+            int firstNonheaderColumnIndex = 0;
+            int lastNonemptyHeaderColumnIndex = 0;
```
muloem: I propose we initialize it to `-1` to indicate that no column has been found rather than defaulting to the first column. 

```suggestion
            int lastNonemptyHeaderColumnIndex = -1;
```

stellanl: So do you then want to get rid of the firstQuestionFound variable? Because that is doing that job right now.

muloem: Good observation and yes. it would not be needed anymore actually! 👍

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -1006,91 +1007,121 @@ protected void invokeUrl(String serverBase, String urlString,
...
-                errorMap.put(0, "First header cell must contain '" + METADATA_LABEL + "'");
-                return errorMap;
+                break;
             }
 
-            int headerRowIndex = splitSheets ? 1 : 0;
+            if (!isEmptyCell(cell)) {
+                lastNonemptyHeaderColumnIndex = cell.getColumnIndex();
+            }
+            if (cellValue.matches("[0-9]+\\|.+")) {
```
muloem: I saw you extracted the regex pattern for the sheet, maybe we can extract this one as well to make it more explicit..?

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
```diff
@@ -1006,91 +1007,121 @@ protected void invokeUrl(String serverBase, String urlString,
...
-                }
-                if (!checkCol(index, DURATION_COLUMN_KEY)) {
-                    errorMap.put(-1, "Column header '" + DURATION_LABEL + "' missing on sheet " + i);
-                }
-                if (i > 0 && !checkCol(index, REPEAT_LABEL)) {
-                    errorMap.put(-1, "Column header '" + REPEAT_LABEL + "' missing on sheet " + i);
+                Sheet sheet = wb.getSheetAt(i);
+                String name = sheet.getSheetName();
+                if (i == 0) { //TODO OR just I==0??
+                //TODO possible variant:   if (RAW_DATA_SHEET_LABEL.equals(name)) {
```
muloem: In fact I think here it should be enough to have something like
```java
int isBaseSheet = false;
if (label == "Raw data") isBaseSheet = true;
else if (label matches "Group ...") isBaseSheet ...
```

And a sheet which doesnt match either of the two doesnt necessarily need to be an error but we ignore it and dont consider for the import.

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
stellanl: Sure

GAE/src/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporter.java
stellanl: The Clojure validation only validates question ID's on sheet 0, right now, and the import also assumes base sheet to be 0, so I want to match that until we redo all this.
Unknown-name sheets now ignored for verification.

GAE/src/test/java/org/waterforpeople/mapping/dataexport/RawDataSpreadsheetImporterTests.java
```diff
@@ -0,0 +1,117 @@
...
+    @Test
+    void testNoQuestionColumns() throws IOException {
+        DataImporter dimp = new RawDataSpreadsheetImporter();
+
+        //Check that a file without any questions fail on "row" -11
+        File file = createNoQuestionsSpreadsheet("/tmp/invalid1.xlsx");
+        Map<Integer,String> errors = dimp.validate(file);
+        assertEquals(1, errors.size());
+        String err1 = errors.get(-11);
+        assertNotEquals(err1, null);
```
muloem: May be more accurate to `assertEquals()` and actually copy the error text here?

Issue #3040, opened 2019-04-02
Apk download url needs to be HTTPS


valllllll2000:
We currently have a json rest api the app queries: 
``` /deviceapprest```
This service returns a json object with amongst other contains the url to download the apk using the DeviceApplication entity of the datastore. The url field is always http. When updating the android tools with Android 28, there is a new feature enabled by default which is to prohibit http calls. For now I have disabled that setting (due to the support url and device time) and also I am replacing http by https when downloading the apk but it would be nice to start using https for these urls as both urls work.
see: https://developer.android.com/training/articles/security-config#CleartextTrafficPermitted for more info

stellanl:
Two approaches: 
1. Change ApkRedirectServlet to modify the URL from the DeviceApplication table.
2. Change the (script?) code that inserts new URLs in DeviceApplication during deployment. We could also make a script that modifies all the old URLs.

muloem:
I would vote for option 2, i.e. what is put in is what comes out. No surprises and having to always remember that there is somewhere that the URL gets modified.

PR #3100, opened 2019-05-20
[#3040]Add script to make all app URLs https.

stellanl:
S3 will stop supporting unencrypted fetches.
#### The solution
Iterate over app URLs and change http: to https:
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/scripts/data/my_lets-encrypt-app-url.sh
+++ b/scripts/data/my_lets-encrypt-app-url.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# USAGE: ./lets-encrypt-app-url.sh akvoflowsandbox [--doit]
+
+APP_ID=$1
+SERVICE_ACCOUNT="sa-$APP_ID@$APP_ID.iam.gserviceaccount.com"
+P12_FILE_PATH="/home/stellan/dev/repos/akvo-flow-server-config/$1/$1.p12"
+
+java -cp bin:"lib/*" \
+     org.akvo.gae.remoteapi.RemoteAPI \
+     LetsEncryptAppUrl \
+     $APP_ID \
+     $SERVICE_ACCOUNT \
+     $P12_FILE_PATH \
+     $2 $3
--- a/scripts/data/src/org/akvo/gae/remoteapi/LetsEncryptAppUrl.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/LetsEncryptAppUrl.java
@@ -22,6 +22,7 @@ import java.util.List;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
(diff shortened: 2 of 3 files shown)


scripts/data/src/org/akvo/gae/remoteapi/LetsEncryptAppUrl.java
```diff
@@ -0,0 +1,62 @@
...
+
+        System.out.println("Fetching device apps.");
+
+        PreparedQuery pq = ds.prepare(new Query(DEVAPPKIND));
+        for (Entity da : pq.asIterable(FetchOptions.Builder.withChunkSize(500))) {
+            String oldurl = (String) da.getProperty(DEVURL_PROP);
+            String newurl = oldurl.replaceFirst("http:", "https:");
+            String appcode  = (String) da.getProperty(APPCODE_PROP);
+            if (!newurl.equals(oldurl)
+//                    && appcode.equals(APPCODE_VALUE) //Only modern apps?
```
valllllll2000: can we maybe remove this commented out code or is it necessary?

scripts/data/my_lets-encrypt-app-url.sh
```diff
@@ -0,0 +1,15 @@
...
+SERVICE_ACCOUNT="sa-$APP_ID@$APP_ID.iam.gserviceaccount.com"
+P12_FILE_PATH="/home/stellan/dev/repos/akvo-flow-server-config/$1/$1.p12"
+
+java -cp bin:"lib/*" \
+     org.akvo.gae.remoteapi.RemoteAPI \
+     LetsEncryptAppUrl \
+     $APP_ID \
+     $SERVICE_ACCOUNT \
+     $P12_FILE_PATH \
+     $2 $3
```
valllllll2000: is it necessary to pass $3? if I understand correctly $1 is the instance name, $2 is the [--doit] param, right?

scripts/data/src/org/akvo/gae/remoteapi/LetsEncryptAppUrl.java
stellanl: That was something I wanted feedback on; should we change only the URLs with the modern code?

scripts/data/my_lets-encrypt-app-url.sh
stellanl: Cut-and-paste, obviously.

scripts/data/src/org/akvo/gae/remoteapi/LetsEncryptAppUrl.java
valllllll2000: Right, I have seen there is also fieldSurvey... no idea, who uses those anyway?

valllllll2000: satStat is no longer used in the latest versions of the app

stellanl: I think fieldSurvey used to be the name of the app. But since the old versions will become unreachable if we do not modify all URLs, I propose to simply change all of them.

Issue #2796, opened 2018-07-25
Monitoring: No more need to send all data points to a device


janagombitova:
# Context 

One of our goals is to ensure users can continuously monitor their data points on large scale. 

## Current monitoring in Flow

In Flow you can define a registration form and its connected monitoring forms in one survey. 

The registration form defines the object of your data collection. If you are interested in learning about the status of a water point in communities in your country, your registration form will hold the basic information about a water point - the information that defines it, that does not change (location, ID, photo, type). If you are interested in learning about the quality of education in schools in your programme, your data point will be a school and in the registration form you will ask questions as location, photo, when was it built, by who, etc. It you are interested in learning about farmers in your programme, then your data point will be a farmer and you will ask him questions in the registration form as her name, date of birth, photo, etc. 

The monitoring forms (one or many) will hold questions that you want to ask your subject of your data collection repetitively over the course of time to see change. In our case of water points one monitoring form can hold questions about the water quality, another monitoring form about the overall status of the water point as if it is functioning, clean, how many people are there, etc, all depending on what you want to analyse. In the school case your monitoring form can be an internal inspection form, another form can be about the satisfaction of the pupils, etc. In the farmer case a monitoring form can hold questions about his crops and yield, where another form on his nutrition. 

When in the field you first submit the registration form, this creates your data point. Now you can submit the monitoring form to that data point. Your first monitoring form submission will create your baseline and the following submissions will help you see change over time. The next time you come back to your data point, you will select the data point from the list of points on your device and submit another monitoring form submission. 

## Current issue
If you are monitoring all water points in a region or country wide we are imagining thousands of data points to monitor. Currently in Flow once a registration form is submitted and a data point created any enumerator who is assigned the survey, can download the created data points to his device. This makes it possible that if Al registered point A, Isha can download also his point A and monitor it, not having the data points only available to Al. 

But if there are thousand points registered, this means once Al and Isha hit download on their Flow app **all thousand points will start downloading to their devices**. 
- takes a lot of time, bandwidth and storage on the device to download all data points (tested to download 6k data points on Amsterdam internet, took me over 3 hours with multiple retries) 
- in areas with low connectivity is it impossible 
- Al does not need to have all the data points as he is responsible to monitor only a certain group of them
- having so many data points, finding the one to monitor becomes painful for Al and Isha and possibly resulting in them selecting the wrong data point causing the submission to link to the wrong previous data creating an analysis headache and risk 
[...]

PR #3103, opened 2019-05-21
Re-implement assignment list using react (connect #2796)

kymni:
Assignment list based on ember
#### The solution
Re-implement using react as a step towards the entire assignment redisign
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/components/devices/AssignmentsList.jsx
```diff
@@ -0,0 +1,110 @@
...
+        className="btnOutline"
+        onClick={() => onEdit('new')}
+        onKeyDown={() => onEdit('new')}
+        style={{ float: 'right' }}
+      >
+        {Ember.String.loc('_create_new_assignment')}
+      </a>
+    </div>
+    <div
+      id="devicesListTable_length"
```
finnfiddle: generally id's are not needed in react. You can remove this if youre not using the id elsewhere in the codebase

Dashboard/app/js/lib/components/devices/AssignmentsList.jsx
```diff
@@ -0,0 +1,110 @@
...
+  onDelete,
+  onSort,
+}) => (
+  <div>
+    <div className="deviceControls">
+      <a
+        className="btnOutline"
+        onClick={() => onEdit('new')}
+        onKeyDown={() => onEdit('new')}
+        style={{ float: 'right' }}
```
finnfiddle: this is fine to be here but better to be in the css

Dashboard/app/js/lib/components/devices/AssignmentsList.jsx
```diff
@@ -0,0 +1,110 @@
...
+    {!assignments.get('length') && (
+      <div style={{ marginTop: '40px' }}>No assignments</div>
+    )}
+    {assignments.get('length') > 0 && (
+      <table className="dataTable">
+        {/* TABLE HEADER */}
+        <thead>
+          <tr>
+            <th
+              className={FLOW.tableColumnControl.get('selected') == 'name' ? (FLOW.tableColumnControl.get('sortAscending') ? 'sorting_asc' : 'sorting_desc') : ''}
```
finnfiddle: nested ternary is generally bad practice because its difficult to read but its not a biggie

Dashboard/app/js/lib/components/devices/AssignmentsList.jsx
```diff
@@ -0,0 +1,110 @@
...
+            >
+              <a
+                onClick={() => onSort('name')}
+                onKeyDown={() => onSort('name')}
+              >
+                {Ember.String.loc('_name')}
```
finnfiddle: NB: try to remove all references to Ember from inside these react components because that way we can decouple the components from ember and they can be reused elsewhere - for example in Lumen which is not an Ember app.
Pass the end value to be rendered it as a prop instead

Dashboard/app/js/lib/components/devices/AssignmentsList.jsx
```diff
@@ -0,0 +1,110 @@
...
+              </a>
+            </th>
+          </tr>
+        </thead>
+        {/* TABLE BODY: MAIN CONTENT */}
+        <tbody>
+          {assignments.map(assignment => (
+            <tr key={assignment.get('keyId')}>
+              <td className="name">{assignment.get('name')}</td>
+              <td>{FLOW.date3(assignment.get('startDate'))}</td>
```
finnfiddle: NB: same as Ember - also try remove all references to FLOW from inside these react components. We want them to be totally decoupled.

Dashboard/app/js/lib/components/devices/AssignmentsList.jsx
```diff
@@ -0,0 +1,110 @@
...
+                {Ember.String.loc('_action')}
+              </a>
+            </th>
+          </tr>
+        </thead>
+        {/* TABLE BODY: MAIN CONTENT */}
+        <tbody>
+          {assignments.map(assignment => (
+            <tr key={assignment.get('keyId')}>
+              <td className="name">{assignment.get('name')}</td>
```
finnfiddle: I know this is probably annoying but is it not possible for assignments to be an array of plain objects so we can avoid having to say `assignment.get('name')` and instead say `assignment.name`? This is because it is a more standard format of data/props and we can more easily use this component later elsewhere

Dashboard/app/js/lib/views/devices/assignments-list-view.jsx
```diff
@@ -1,17 +1,113 @@
...
+    this._super();
+    this.assignmentsList = this.assignmentsList.bind(this);
+    this.assignmentEdit = this.assignmentEdit.bind(this);
+    this.assignmentDeleteConfirm = this.assignmentDeleteConfirm.bind(this);
+    this.assignmentSort = this.assignmentSort.bind(this);
+  },
+
   didInsertElement(...args) {
     this._super(...args);
+    const self = this;
```
finnfiddle: with arrow functions you dont need to do this anymore - see here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_separate_this

Issue #3125, opened 2019-06-25
Auth0 Integration


dlebrero:
Instead of moving to Keycloak, we decided to use Auth0 as our OIDC provider.

See #1265 for more background.

valllllll2000:
There is not test plan but I tested login and logout and it works

PR #3129, opened 2019-06-27
Auth0 login support

dlebrero:

Related to #3125.

Auth0 will allow for SingleSignOn with Lumen (and other Flow instances),
plus users will not need to have a Google email account.

#### The solution

Implements the OIDC Authorization Code Flow. 

All code is behind a feature flag. By default the login/logout should work as before, requiring no config changes in any instance.

To enable Auth0, the config values for oidcDomain, oidcClientId, oidcClientSecret need to be provided (see uat2 config for the required values for the Auth0 test environment)

After these are configured, the user can **manually** hit the /altlogin endpoint to start the Auth0 login process. The logout button should work for both Auth0 and Google. This should allow us to dark launch the feature.

If Auth0 is configured, there is an additional feature flag `showAltAuthButton` that will show an additional Login button (bigger and besides the Google one) that will start the Auth0 login process. Both login processes should work.

The last feature flag is `defaultAuthProvider` that when it has the value `auth0` will make the old login button to use the Auth0 login process.

UAT2 should be configured to show the Auth0 login button.

## Testing

* [ ] In UAT1, check that the user can login/logout with Google as before, without changes in the uat1 configuration
* [ ] In UAT2, check that the user can login with Google or with Auth0

## TODO

* [x] Is the implicit or Hybrid flow the most appropriate?
      Code Flow is more secure but Single Logout more difficult. Hybrid requires changes in Ember. Given that Single Logout is not initially going to be implemented, Code Flow is more appropriate.
* [x] Single Logout
      If a user logins with Google, a logout from Auth0/Flow/Lumen will be pointless unless we also log him out of Google. Probably it will be very confusing for a user to be log out from Gmail when he logs out of Flow. Given this scenario, Single Logout for non-google users can be delayed as it is complex given that each Flow backend would need to be informed of the logout.
* [x] Do we need one Client per tenant?
     Irrelevant for this PR
* [x] Are we doing enough checks in the tokens?
     Code Flow does not work with tokens.
* [x] What to do with the local dev env? Use GAE? Use KeyCloak? 
     Given that we still need to support GAE, we will keep using GAE in the local dev.
* [x] Production configuration
     Irrelevant for this PR
* [x] Session expiration 
    Code Flow sessions are held in the Flow backend, so the behaviour will be the same as the current one.

## Checklist
* [x] Connect the issue
* [x] Test plan
* [x] Copyright header
* [x] Code formatting
* [x] Documentation


GAE/src/org/waterforpeople/mapping/app/web/TestHarnessServlet.java
```diff
@@ -373,12 +374,12 @@ public void doGet(HttpServletRequest req, HttpServletResponse resp) {
         }
     }
 
-    private void setupTestUser() {
+    private void setupTestUser(String email) {
```
muloem: I think we can remove this entire method. It was used initially when setting up test accounts for the local server but now we have sample data that includes test users so this is no longer necessary.

GAE/src/org/waterforpeople/mapping/app/web/rest/security/LogoutController.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private void logoutFromGAE(HttpServletResponse res) throws IOException {
+        res.sendRedirect(UserServiceFactory.getUserService().createLogoutURL(
+                "/"));
+    }
+
+    private void logoutFromOIDC(HttpServletRequest req, HttpServletResponse res) throws IOException {
+        String returnTo = req.getScheme() + "://" + req.getServerName();
+        if ((req.getScheme().equals("http") && req.getServerPort() != 80) || (req.getScheme().equals("https") && req.getServerPort() != 443)) {
```
muloem: The `http` is to take care of the local (server) setup?

GAE/src/org/waterforpeople/mapping/app/web/rest/security/featureflag/EntryPoint.java
```diff
@@ -0,0 +1,34 @@
+package org.waterforpeople.mapping.app.web.rest.security.featureflag;
```
muloem: FYI To differentiate these we have been trying to move as many of the newer classes to new packages under the `org.akvo` domain whenever possible.

muloem: Plus not forgetting the licensing info in general. 👍

GAE/src/org/waterforpeople/mapping/app/web/TestHarnessServlet.java
dlebrero: For testing the Google integration, I am creating a new user that matches my Gmail account.

None of the tests accounts are admin accounts, so none can create new users.

dlebrero:
Ready to merge

GAE/src/org/akvo/flow/rest/security/oidc/CallbackController.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private static final Logger log = Logger.getLogger(CallbackController.class.getName());
+
+    private final AuthenticationController controller;
+    private final String redirectOnFail;
+    private final String redirectOnSuccess;
+
+    public CallbackController(AppConfig appConfig) {
+        this.controller = appConfig.authenticationController();
+        this.redirectOnFail = "/"; // Where to redirect?
```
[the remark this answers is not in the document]
dlebrero: In fact, when a user registers for the very first time, he needs to verify his email before being able to login to Flow, which Auth0 treats as an error.

We should handle also this particular case better so the user knows what he must do.

GAE/src/org/akvo/flow/rest/security/google/GoogleAuthenticationFilter.java
```diff
@@ -95,13 +93,17 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
     private boolean loggedInUserMatchesGaeUser(Authentication authentication, User googleUser) {
         assert authentication != null;
 
+        GaeUser gaeUser = (GaeUser) authentication.getPrincipal();
+
+        if (!gaeUser.isAuthByGAE()) {
+            return true;
```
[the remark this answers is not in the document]
dlebrero: Reworked the logic to make it less confusing. Added test to make it even less confusing.

PR #3131, opened 2019-07-02
Folder survey selector in react

kymni:
Folder/survey selector implemented in ember
#### The solution
Re-implement logic in react
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/views/folder-survey-selector.jsx
+++ b/Dashboard/app/js/lib/views/folder-survey-selector.jsx
@@ -40,16 +40,16 @@ FLOW.FolderSurveySelectorView = FLOW.ReactComponentView.extend(observe({
     let surveySelected = false;
     // if not root level, generate ancestor selectors
     if (parentId !== 0) {
-      const parent = this.surveyGroups.find(sg => sg.keyId === parentId);
+      const parent = this.surveyGroups.find(sg => sg.keyId == parentId);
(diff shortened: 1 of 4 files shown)


Dashboard/app/js/lib/views/folder-survey-selector.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+  folderSurveySelector(parentId) {
+    if (!FLOW.projectControl.content.isLoaded) return;
+    const levels = [];
+    let surveySelected = false;
+    // if not root level, generate ancestor selectors
+    if (parentId !== 0) {
+      const parent = this.surveyGroups.find(sg => sg.keyId == parentId);
+      for (let i = 0; i < parent.ancestorIds.length; i++) {
+        const level = [{
```
muloem: Is it possible to use a placeholder like we have always been using, instead of an object whose function mimics the placeholder?

kymni: It is possible but the idea being to move away from Ember means we currently mimic the Ember objects we've been using to avoid calling functions such as get when accessing data. This way the components such as the `ChildOption` I've created can be reused going forward by simply passing an object to them

Issue #3132, opened 2019-07-04
Disable saving assignment if name not set


kymni:
It's currently possible to save an assignment without specifying a name resulting in empty rows on the assignments list

PR #3133, opened 2019-07-08
Disable saving nameless assignment (connect #3132)

kymni:
Save option available even when no assignment name is set resulting in a sometimes empty assignments 
#### The solution
Add validation for assignment name
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/Dashboard/app/js/lib/views/devices/assignments-list-view.jsx
+++ b/Dashboard/app/js/lib/views/devices/assignments-list-view.jsx
@@ -64,7 +64,9 @@ FLOW.AssignmentsListView = FLOW.ReactComponentView.extend(observe({
   assignmentEdit(action, assignmentId) {
     switch (action) {
       case 'new': {
-        const newAssignment = FLOW.store.createRecord(FLOW.SurveyAssignment, {});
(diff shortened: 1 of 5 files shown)


Dashboard/app/js/lib/views/devices/assignments-list-view.jsx
```diff
@@ -64,7 +64,9 @@ FLOW.AssignmentsListView = FLOW.ReactComponentView.extend(observe({
   assignmentEdit(action, assignmentId) {
     switch (action) {
       case 'new': {
-        const newAssignment = FLOW.store.createRecord(FLOW.SurveyAssignment, {});
+        const newAssignment = FLOW.store.createRecord(FLOW.SurveyAssignment, {
+          name: Ember.String.loc('_new_question_please_change_name'),
```
valllllll2000: why question? is the string identical? if so maybe we should rename it something more meaningful?

kymni: changed

Issue #3142, opened 2019-07-24
Optimise batch fetching of entities by their IDs


muloem:
When we have the IDs of multiple entities, we use a `listByKeys(List)` method to retrieve them.  However, internally this method iterates of `listByKey()` this hitting the datastore with multiple requests for entities in a short span.  We can improve this by making a single query with all the IDs.

muloem:
## Test Plan

Covered by regression testing

PR #3144, opened 2019-07-25
[#3142] Retrieve entities with known IDs in a single batch request

muloem:

* Entities with known IDs retrieved one by one by repeatedly making requests to the datastore for each entity.

#### The solution

* We make a single request to the datastore with the list of IDs

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/com/gallatinsystems/framework/dao/BaseDAO.java
+++ b/GAE/src/com/gallatinsystems/framework/dao/BaseDAO.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2010-2018 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2010-2019 Stichting Akvo (Akvo Foundation)
  *
  *  This file is part of Akvo FLOW.
  *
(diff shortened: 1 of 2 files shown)


GAE/src/com/gallatinsystems/framework/dao/BaseDAO.java
```diff
@@ -436,17 +431,34 @@ protected T findByProperty(String propertyName, Object propertyValue,
...
+            Key key = KeyFactory.createKey(clazz.getSimpleName(), id);
+            Object objectId = pm.newObjectIdInstance(clazz, key);
+            datastoreKeysList.add(objectId);
         }
-        return list;
+
+        final List<T> resultsList = new ArrayList<>();
+
+        try {
+            resultsList.addAll(pm.getObjectsById(datastoreKeysList));
```
valllllll2000: What would happen if some of the ids are found and others are not, would it still work and returns the ones that were actually found or throw an exception?

muloem: So I made a fallback method incase there is any wrong ID, the fall back retrieves the items one by one

Issue #3031, opened 2019-03-27
Push Auth information to Unilog


dlebrero:
To build the new "grouping things app" (GTA), we need the user, user roles and user authorization data in the Unilog so that the GTA can implement its logic

PR #3163, opened 2019-08-16
Exporting SurveryGroups

dlebrero:
Adding SurveyGroups to Unilog after the clean up. 
Script already run in test and production.

Part of #3031

### DIFF
--- a/scripts/data/src/org/akvo/gae/remoteapi/ExportAuthDataToEventLog.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/ExportAuthDataToEventLog.java
@@ -42,7 +42,7 @@ public class ExportAuthDataToEventLog implements Process {
                     System.out.println("progress " + amount);
                 }
                 Map<String, Object> event = createEvent(entity, eventTypes, orgId);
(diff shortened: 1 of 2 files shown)


scripts/data/src/org/akvo/gae/remoteapi/ExportAuthDataToEventLog.java
```diff
@@ -38,11 +42,13 @@ public void execute(DatastoreService ds, String[] args) throws Exception {
                     System.out.println("progress " + amount);
                 }
                 Map<String, Object> event = createEvent(entity, eventTypes, orgId);
-                Date timestamp = (Date) entity.getProperty(EventUtils.Prop.LAST_UPDATE_DATE_TIME);
+                Date timestamp = new Date();
                 Entity eventLogEntity = EventUtils.createEventLogEntity(event, timestamp);
                 ds.put(eventLogEntity);
             }
 
+            System.out.println("amount = " + amount);
```
muloem: Was this for personal debugging?  We usually use the Logger if we want this in the GAE logs.

dlebrero: This code is not going to be run in GAE but in the local dev box.

The information is useful to know how big the dataset was exported.

Issue #3176, opened 2019-09-10
Data scripts should keep Unilog updated


dlebrero:
Right now, any change made in an entity by a data script is not reflected in the Unilog, which means that there are 
some inconsistencies between the DataStore and the Unilog.

Entities created or updated can be synchronized again by doing a new data export for the Unilog, but to sync entities deleted we need to 
clean up all the Unilog events and do a reimport.

All of this is cumbersome, time consuming and inefficient, plus difficult to detect.

Proposed change is to make the data scripts aware of Unilog and for them to push the required event entities to the EventQueue kind

PR #3177, opened 2019-09-10
Data scripts to update Unilog

dlebrero:
See #3176

Wrapping the default datastore into one that is Unilog aware

Requires no change in any of the current or future data scripts.

Any change in an entity will imply x2 writes, one for the entity itself and
one for the event.

Events are added after changes in the entities.

Notify Unilog of new events, using the unilog url found in the appengine-web.xml, if exists.

### DIFF
--- a/scripts/data/src/org/akvo/gae/remoteapi/DataStoreWithUnilog.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/DataStoreWithUnilog.java
@@ -118,7 +118,9 @@ public class DataStoreWithUnilog implements DatastoreService {
         final EventTypes eventTypes = EventUtils
                 .getEventAndActionType(kind);
 
-        if (eventTypes != null) {
+        if (eventTypes == null) {
+            return null;
+        } else {
             Map<String, Object> event = updateOrCreateEvent(entity, eventTypes);
             try {
                 return EventUtils.createEventLogEntity(event, new Date());
@@ -126,7 +128,6 @@ public class DataStoreWithUnilog implements DatastoreService {
                 throw new RuntimeException(e);
             }
         }
-        return entity;
     }
 
     private Map<String, Object> updateOrCreateEvent(Entity entity,
--- a/scripts/data/src/org/akvo/gae/remoteapi/DataStoreWithUnilog.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/DataStoreWithUnilog.java
@@ -31,6 +31,7 @@ public class DataStoreWithUnilog implements DatastoreService {
     private static final Date MANUAL_FIX_START_TIME = new Date();
     private DatastoreService delegate;
     private String orgId;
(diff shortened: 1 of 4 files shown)


scripts/data/src/org/akvo/gae/remoteapi/DataStoreWithUnilog.java
```diff
@@ -0,0 +1,265 @@
...
+    @Override
+    public Key put(Entity entity) {
+        Key put = delegate.put(entity);
+        storeEntityUpdate(Stream.of(entity));
+        return put;
+    }
+
+    @Override
+    public Key put(Transaction transaction, Entity entity) {
+        Key put = delegate.put(transaction, entity);
```
muloem: Minor comment: its a funny name for the variable. 😄

scripts/data/src/org/akvo/gae/remoteapi/DataStoreWithUnilog.java
```diff
@@ -0,0 +1,265 @@
...
+
+        if (eventTypes != null) {
+            Map<String, Object> event = updateOrCreateEvent(entity, eventTypes);
+            try {
+                return EventUtils.createEventLogEntity(event, new Date());
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+        return entity;
```
muloem: Does this mean if no event type is found then the entity would get stored as is in the event queue?

scripts/data/src/org/akvo/gae/remoteapi/RemoteAPI.java
```diff
@@ -60,16 +76,21 @@ public static void main(String[] args) {
 
         final RemoteApiInstaller installer = new RemoteApiInstaller();
 
+        String orgId = args[1];
```
muloem: rather than add this as an argument that has to be entered, how about we get it from the `appengine-web.xml` as well (like you do with the `eventNotification` property?

scripts/data/src/org/akvo/gae/remoteapi/DataStoreWithUnilog.java
dlebrero: Intellij default name! Didn't even notice :P. Fixed

scripts/data/src/org/akvo/gae/remoteapi/DataStoreWithUnilog.java
dlebrero: Correct. It is a bug. Fixed. 

Good catch!

scripts/data/src/org/akvo/gae/remoteapi/RemoteAPI.java
dlebrero: There is the special case for local development, which does not require any files. See a further up the method how args[1] is used.

In any case, changed the method to make this more obvious. `orgId` is now used everywhere instead of `args[1]`

Issue #3167, opened 2019-08-21
Update /surveyedlocale endpoint to enable datapoint filtering


muloem:
As part of the updates to the assignments (see #2796) that enable selecting specific datapoints to assign them to devices (#3166), we need to modify the endpoint used by devices to download datapoints, to enable it filter out and only return datapoints assigned to a specific device.

Some issues to consider:
* If we return such a filtered list, would it be processed without any problems by devices with current versions of the app?

stellanl:
To avoid backwards-compatibility problems we decided to add a new endpoint, /assigned_datapoints that will return only the datapoints (surveyedLocales) the device has been assigned. It also needs info about the device so it can identify it.

Old apps continue to use the old endpoint and gets all data point as before.

janagombitova:
I guess there is nothing user-facing to test for this specific issue for now, right? Is this change on UAT1? Can we close the issue?

janagombitova:
All linked PRs are merged and released. If this still needs work - please reopen

PR #3196, opened 2019-09-23
Issue/3167 filter app datapoints (Connect #3167)

stellanl:
No way for app to get only assigned data points.
#### The solution
Add a new endpoint with modern names.
Moved most of the old endpoint (SurveyedLocaleServlet) methods to the new service (DataPointServlet) and share them.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


GAE/src/org/akvo/flow/api/app/DataPointServlet.java
```diff
@@ -120,14 +135,13 @@ private SurveyedLocaleResponse convertToResponse(List<SurveyedLocale> slList,
...
+        List<SurveyedLocaleDto> dtoList = getSurveyedLocaleDtosList(slList, surveyId);
 
-        resp.setSurveyedLocaleData(dtoList);
+        resp.setDataPointData(dtoList);
         return resp;
     }
 
-    private List<SurveyedLocaleDto> getSurveyedLocaleDtosList(List<SurveyedLocale> slList,
-            Long surveyGroupId) {
+    protected static List<SurveyedLocaleDto> getSurveyedLocaleDtosList(List<SurveyedLocale> slList, Long surveyId) {
```
valllllll2000: Just out of curiosity, why is this method static?

GAE/src/org/akvo/flow/api/app/DataPointRequest.java
```diff
@@ -0,0 +1,79 @@
...
+ *  Akvo FLOW is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *  See the GNU Affero General Public License included below for more details.
+ *
+ *  The full license text can also be seen at <http://www.gnu.org/licenses/agpl.html>.
+ */
+
+package org.akvo.flow.api.app;
+
+import java.util.Date;
```
valllllll2000: This seems to be unused

GAE/src/org/akvo/flow/api/app/SurveyedLocaleServlet.java
```diff
@@ -0,0 +1,136 @@
...
+        if (slList == null) {
+            resp.setCode(String.valueOf(HttpServletResponse.SC_INTERNAL_SERVER_ERROR));
+            resp.setMessage("Internal Server Error");
+            return resp;
+        }
+        // set meta data
+        resp.setCode(String.valueOf(HttpServletResponse.SC_OK));
+        resp.setResultCount(slList.size());
+
+        List<SurveyedLocaleDto> dtoList = DataPointServlet.getSurveyedLocaleDtosList(slList, surveyId);
```
valllllll2000: In my opinion, there should be another class/classes containing methods used by both servlets. This will help have smaller classes and avoid static methods which are harder to test.

GAE/src/org/akvo/flow/api/app/DataPointServlet.java
valllllll2000: Ok I understood why, see my comment below why I may suggest to do it differently.

GAE/src/org/akvo/flow/api/app/SurveyedLocaleServlet.java
stellanl: Sure. Just hoped the old servlet would be removed soon. And why are static methods harder to test?

valllllll2000: When writing tests one usually initialize to class to test with a constructor but of course there are tools like powermock and then static classes methods can also be tested. So there could be a static utility class with utility methods but it seems unusual to me that a servlet calls another one. If you decide to remove the other servlet, then the methods would not need to be static any more. Also what disturbs me a bit is that DataPointServlet is used both with static methods and instantiated via constructor.

valllllll2000: ok to sum up: if they are only utility methods -> we can move them to a static utility class, I guess in this case this would require less rework and makes sense. _I still think they are not exactly utility classes since they access the database and read data and in this case if we wanted to mock some responses for them, they should be made into an object with constructor but it would require a change in architecture so maybe not needed right now._

Issue #3197, opened 2019-09-25
Issues with new assignments survey selector


janagombitova:
When testing UAT2 the new assignments changes on v1.9.49-107-g656c14d3 I found the following issues that need to be addressed. 

@marvinkome I gave you access to https://uat2.akvoflow.org

# Switching folders 

Steps to take to reproduce the issue: 
1. Select a folder and then the subfolder or survey. Then I realise I am in the wrong main folder. 
2. Select the other main folder. I assume the 2nd dropdown will update with subfolders and surveys from this new main folder but it does not. 
3. If I select the 1st item in the dropdown - 'Choose folder or survey' I then get another dropdown instead of the first one updating and the second one disappearing. 

My expectation was that once I select another main folder in the 1st dropdown, the following ones will update accordingly. 

![Assignments folder swith](https://user-images.githubusercontent.com/12456965/65580666-64b37f00-df7a-11e9-9751-9c48aa629c68.gif)

# 2nd dropdown does not show subfolders 
I have a folder within a folder and then another folder and only then my published survey. I want to be able to select up to that last folder, but the 2nd dropdown does not show after selecting my main folder showing the other subfolders. 

![Assignments 2nd dropdown does not show folders](https://user-images.githubusercontent.com/12456965/65580885-c8d64300-df7a-11e9-90c8-e1ab66bb3162.gif)

# Published forms do not show 
After I select my survey, the list of published forms for that survey should show. After selecting a survey, with a published form, it does not show under forms. 
![Assignments pubished forms do not show](https://user-images.githubusercontent.com/12456965/65581486-d50ed000-df7b-11e9-9046-8df2350d8b73.gif)

# Existing assignments miss data 
When viewing existing assignments that had forms selected and devices, they are empty now. 
![Assignments existing one](https://user-images.githubusercontent.com/12456965/65581432-bdcfe280-df7b-11e9-8645-248d21059c93.gif)

janagombitova:
All seems to be resolved. I will create a separate issue if I find any more problems.

PR #3199, opened 2019-09-26
Handle changed properties

kymni:
`surveys` and `devices` properties from the survey_assignments endpoint changed but not handled on front-end causing forms and devices selection not to work
#### The solution
handle both
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


kymni:
should fix #3197

PR #3206, opened 2019-09-27
Issue/3192 move assignments edit to react

marvinkome:
Move whole assignment edit view to react

## Checklist
* [x] Connect the issue
* [ ] Test plan
* [x] Copyright header
* [x] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
```diff
@@ -0,0 +1,201 @@
...
+
+  onSubmit = () => {
+    const { assignmentName, startDate, expireDate } = this.state;
+    if (this.assignmentNotComplete()) return;
+
+    this.props.actions.onSubmit({ assignmentName, startDate, expireDate });
+  }
+
+  // helpers
+  validateAssignment = (assignmentName) => {
```
muloem: As its validating only the assignment name maybe it makes sense to specify this?

```suggestion
  validateAssignmentName = (assignmentName) => {
```

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
```diff
@@ -0,0 +1,201 @@
...
+      startDate: this.state.startDate,
+      expireDate: this.state.expireDate,
+    };
+
+    return {
+      assignmentDetailsState,
+    };
+  }
+
+  assignmentNotComplete = () => {
```
muloem: This seems to be doing the same thing as the `validateAssignment` function above except it shows a dialog popup. Although the date validation below is not included but maybe we can implement that in a similar way to the validation before and then skip using the `FLOW.dialogControl`. One less dependency on the global `FLOW` variable. What do you think?

Dashboard/.eslintrc
```diff
@@ -1,77 +1,77 @@
...
+    "browser": true,
+    "node": true,
+    "jest": true
+  },
+  "plugins": ["react", "babel"],
+  "parserOptions": {
+    "ecmaFeatures": {
+      "jsx": true
     }
+  },
```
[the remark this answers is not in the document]
marvinkome: Yup, mainly formatting.

janagombitova:
@muloem I agree. All mandatory items should be selected before an assignment can be saved

muloem:
An assignment is created even when the save assignment button is supposed to be disabled.
However, when the page is reloaded, it disappears from the list.

assignment-with-disbaled-save

Kiarii:
> @muloem I agree. All mandatory items should be selected before an assignment can be saved, thus also at least one form and one device
> 
> I prefer disabling the button rather than having popups. But indeed this is @Kiarii's call

I think requiring the selection of all mandatory items is fine, we might need to indicate to the user of what is still un-selected if they hit the [save assignment] before completing all fields..

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
```diff
@@ -0,0 +1,210 @@
...
+    };
+  }
+
+  // render
+  render() {
+    const { nameValidationMsg, startDateMsg, expireDateMsg } = this.state;
+    const { strings, actions, data } = this.props;
+    const { assignmentDetailsState } = this.formatStateForComponents();
+    const showSubmitBtn =
+      !nameValidationMsg.length && !startDateMsg.length && !expireDateMsg.length;
```
muloem: This was a little confusing. Does that mean we only show the submit button when the fields are empty?  I thought it would be the reverse.

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
```diff
@@ -0,0 +1,210 @@
...
+                  </button>
+                </li>
+              ) : (
+                  <li>
+                    <button type="button" className="button noChanges" id="standardBtn">
+                      {strings.saveAssignment}
+                    </button>
+                  </li>
+                )}
+              <li>
```
muloem: This is the part that wasn't clear. The `onClick` and other actions are only for when the name, and dates values are empty?

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
marvinkome: The `showSubmitBtn` is `true` if all validationMsg is empty.

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
```diff
@@ -0,0 +1,210 @@
...
+    if (key === 'assignmentName') this.validateAssignmentName(value);
+    if (key === 'startDate' || key === 'expireDate') this.validateDate(`${key}Msg`, value);
+    this.setState({ [key]: value });
+  }
+
+  onSubmit = () => {
+    const { assignmentName, startDate, expireDate } = this.state;
+
+    this.props.actions.onSubmit({ assignmentName, startDate, expireDate });
+  }
```
muloem: Perhaps we can also add the validation just before submitting?

marvinkome: Well, the problem here is that we don't want to add the FLOW variable to the react side

muloem: I mean the same validation that you are performing in the `onChangeState` function. Does that require the FLOW variable?

Dashboard/app/js/lib/views/devices/assignment-edit-views.jsx
```diff
@@ -172,18 +171,35 @@ FLOW.AssignmentEditView = FLOW.ReactComponentView.extend(
...
+        sa.set('surveyId', FLOW.selectedControl.get('selectedSurveyGroup').get('keyId'));
+      }
+
 
       FLOW.selectedControl.get('selectedDevices').forEach((item) => {
         devices.push(item.get('keyId'));
       });
+
+      if (!devices.length) {
+        alert('Please select a device to continue');
```
muloem: We need to use a standardised way to do this. So for now lets not use any alerts.  Lets take this out and perhaps push the notifications/validation to the next release.

Dashboard/app/js/lib/views/devices/assignment-edit-views.jsx
```diff
@@ -172,18 +171,35 @@ FLOW.AssignmentEditView = FLOW.ReactComponentView.extend(
...
+      }
+
       sa.set('deviceIds', devices);
 
       FLOW.selectedControl.get('selectedSurveys').forEach((item) => {
         surveys.push(item.get('keyId'));
       });
+
+      if (!surveys || !surveys.length) {
+        alert('Please select a form to continue');
```
muloem: Same.. no alerts

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
```diff
@@ -20,6 +20,8 @@ export default class AssignmentsEditView extends React.Component {
   // lifecycle methods
   componentDidMount() {
     this.validateAssignmentName(this.props.inputValues.assignmentName);
+    this.validateDate('startDateMsg', this.props.inputValues.startDate || '');
+    this.validateDate('expireDateMsg', this.props.inputValues.toDate || '');
```
muloem: This causes the red text to show up next to the dates as soon as you click on _Create new assignment_

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
marvinkome: Oh, so rather than it being in the `onChange` method. It should just validate when the save button is clicked. I get it now?

Dashboard/app/js/lib/views/devices/assignment-edit-views.jsx
marvinkome: So what do we put right now?

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
muloem: Will be addressed under https://github.com/akvo/akvo-flow/issues/3221

Dashboard/app/js/lib/views/devices/assignment-edit-views.jsx
muloem: We've used dialog boxes and will do the rest in this issue https://github.com/akvo/akvo-flow/issues/3221

Issue #2559, opened 2018-03-27
Remove Flow landing/log in page


janagombitova:
# Context 

We create separate Flow instances for each of our partners. This means they get their own Flow url. Once you open it up you see this: 
- Flow logo, 
- a small log in button, 
- a public facing map with data marked as public if any 
- a footer with: 
   - Software updates linking to the GH release notes, 
   - Help & support link  
   - Terms of services link. 
- Copyright and link to akvo.org

This is not the most informative nor inviting landing page.

The users go to this page primarily to log into Flow's online workspace. 

The idea behind the map was to showcase data set as public, but that is not used in this way in practice. The map takes up most of the screen once one opens the url. If we look at the current Flow instances and their public facing maps we have out of the 105 active partner instances only 43 showing data (of any kind) publicly on the Flow public map. For the partners who do not show data, this results in their Flow landing page to show a blank map. 

![screen shot 2017-11-16 at 12 23 41](https://user-images.githubusercontent.com/12456965/32888873-0f5cff26-cac9-11e7-9bce-ae73b2b9b3c7.png)

# Problems 

## Security risk

Currently users set surveys as private or public. The default setting is private but some surveys may be set as public. Data captured to these survey forms will show on this map. But the data can contain private data and if our partners are not fully aware of what data is publicly available, then they are facing a data security risk. 

## What is the value of the map? 

The map takes up most of the screen once one opens the url. If we look at the current Flow instances and their public facing maps we have out of the 105 active partner instances only 43 showing data (of any kind) publicly on the Flow public map. For the partners who do not show data, this results in their Flow landing page to show a blank map. 

Our users do not use Flow as a platform to share the data they have set as public. Not so many partners have their data set as public, mostly because of data privacy reasons and the granularity of the setting is missing (you want to share answers to some questions but not to all).

When one opens the Flow url and sees the map (with or without data) many users asks: Why is the map here? or they do not understand what Flow is from looking at the page or the linkage between Flow and Akvo and other Akvo tools and services. 

# Currently 

Salim enters his organisation's Flow url in the browser. Hits log in. Logs in with his Google account. Accesses Flow if he has an account. If he does not have a user account he gets a page telling him so. 

## Salim is a Flow user 
![log in1](https://user-images.githubusercontent.com/12456965/37969813-0748133c-31d2-11e8-98ca-b93ed79ddedb.gif)

## Salim is not a Flow user 
[...]

Issue #3213, opened 2019-10-02
Remove code related to the public map


muloem:
As part of the work under https://github.com/akvo/akvo-flow/issues/2559, we have added a redirect that takes a user directly to the login page when they try to access the Flow public page.  However, this is a partial step in this direction.  We need to update the entire build process to remove all the code related to the building of the public page.

PR #3216, opened 2019-10-04
[#2559] Redirect to "/altlogin"

loicsans:
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


muloem:
we will set the flag `defaultAuthProvider` and so we do not need this for now.  Instead we will fix the larger issue in #3213

Issue #3249, opened 2019-10-22
Published-form based reports cannot handle RQGs


stellanl:
Get an NPE.

janagombitova:
Anything user visible to test? If no, we can close the issue

muloem:
No issues found in regression testing 👍

PR #3250, opened 2019-10-22
Issue/3249 Handle RQGs (connect #3249)

stellanl:
An RQG on a published form caused an NPE in report generator.
#### The solution
Use correct key for map lookup.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
@@ -186,7 +186,6 @@ public class GraphicalSurveySummaryExporter extends SurveySummaryExporter {
     private Map<String, Integer> columnIndexMap = new HashMap<>();
 
     // maps from a (repeatable) question group dto to the sheet that contains the raw data for it (if split)
-    //private Map<Long, Sheet> qgSheetMap = new HashMap<>();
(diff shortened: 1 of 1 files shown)


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -514,9 +512,8 @@ public void run() {
 
                 //For each group, write the repeats from top to bottom
                 for (Entry<QuestionGroupDto, List<QuestionDto>> groupEntry : questionMap.entrySet()) {
-                    Long gid = groupEntry.getKey().getKeyId();
                     if (safeTrue(groupEntry.getKey().getRepeatable())) {
-                        writeInstanceDataSplit(qgSheetMap.get(gid),
+                        writeInstanceDataSplit(qgSheetMap.get(groupEntry.getKey()),
```
iperdomo: The previous version of this code, `gid` was a `Long` and now we use `Key` as value. Is this correct?

stellanl: Yes. It tests ok. That Long was the only remnant of the old qgSheetMap keys, which were the QuestionGroup key ids. But there are none for a published form. So qgSheetMap was changed to heve QuestionGroupDto's as keys. In that one place it was not changed.

PR #3264, opened 2019-10-29
[#3260]Fix data exporter bugs.

stellanl:
Option read from published XML were not split into columns.
Image filenames required a slash to be put into a report.
#### The solution
Copy option flags around.
Better code for stripping filename paths.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
+++ b/GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
@@ -911,9 +911,9 @@ public class GraphicalSurveySummaryExporter extends SurveySummaryExporter {
         Media media = MediaResponse.parse(value);
         String filename = media.getFilename();
         if (filename != null) {
-            final int filenameIndex = filename.lastIndexOf("/") + 1;
-            if (filenameIndex > 0 && filenameIndex < filename.length()) {
-                cells.add(imagePrefix + filename.substring(filenameIndex));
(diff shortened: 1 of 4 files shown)


GAE/src/org/waterforpeople/mapping/dataexport/GraphicalSurveySummaryExporter.java
```diff
@@ -911,9 +911,9 @@ private static String dateCellValue(String value) {
         Media media = MediaResponse.parse(value);
         String filename = media.getFilename();
         if (filename != null) {
-            final int filenameIndex = filename.lastIndexOf("/") + 1;
-            if (filenameIndex > 0 && filenameIndex < filename.length()) {
-                cells.add(imagePrefix + filename.substring(filenameIndex));
+            filename = filename.substring(filename.lastIndexOf("/") + 1); //strip path, if any
```
iperdomo: I would suggest to use `Paths.get(filename).getFileName()` instead of doing _substring_ manipulation.
`java.nio.file.Paths` was introduced in Java 7, so is safe to use.
```java
Paths.get("/this/is/a/long/path.txt").getFileName().equals(Paths.get("path.txt").getFileName());
// true
Paths.get("/this/is/a////long/path.txt").getFileName().equals(Paths.get("path.txt").getFileName())
// true
```

stellanl: Done

Issue #3180, opened 2019-09-12
Geotag not showing


janagombitova:
@janagombitova moved this issue from Flow app repo [Thu Sep 12 2019](https://github.com/akvo/akvo-flow-mobile/issues/1479)

## Issue 
If a user adds a photo as an answer to the question from the Gallery and this photo has location data in the EXIF, the data is saved well but not shown in the Flow maps or Data analysis exports.  In the app you see the confirmation that the geolocation is saved, but in the map on the online workspace or in the Data analysis export the geotag does not show. So the impression is that the geotag is not saved. 

![Screenshot_2019-09-12-13-16-55](https://user-images.githubusercontent.com/12456965/64781591-83ba1600-d563-11e9-93e6-c4c3185a73fc.png)
![Screen Shot 2019-09-12 at 13 22 14](https://user-images.githubusercontent.com/12456965/64781593-861c7000-d563-11e9-99dd-9ecc96cee982.png)

## Details
-  I saved an image taken in Vietnam and submitted in from the Amsterdam office 
- https://uat1.akvoflow.org/admin/
- _Lumen tests > Example dataset (for testing and demos) > New form
- Submission made today 12-9-2019

Valeria checked on the app's end everything works as expected: the image is sent with the correct exif data and the correct exif data is present on the image that is on s3, but the dashboard does not display the data

janagombitova:
Just an update: a partner reported the issue with photos captured via the camera. They showed the geotag in the app but no longer in Flow. When downloading the photo to geotag is in the exif file. 

The expectation is to read the exif tag in Flow at all times (in the online workspace and in the exports)

stellanl:
Interestingly, when I analyse the exif in that image, I get an error.
I have to include the -d (debug) flag to the exif command to see all of the tags.

EXIF tags in '67369622-a5061d00-f579-11e9-89e4-7e2eb7a806c4.jpg' ('Motorola' byte order):
--------------------+----------------------------------------------------------
Tag                 |Value
--------------------+----------------------------------------------------------
Image Length        |230
Orientation         |Top-left
Image Width         |408
X-Resolution        |72
Y-Resolution        |72
Resolution Unit     |Inch
Corrupt data
The data provided does not follow the specification.
ExifEntry: The tag 'LightSource' contains data of an invalid format ('Long', expected 'Short').
Light Source        |
Exif Version        |Exif Version 2.1
FlashPixVersion     |FlashPix Version 1.0
[...]

PR #3267, opened 2019-10-30
[#3180]Extract-missing-geotags script.

stellanl:
Geotags in image Exif were not extracted by the app for a while, and thus do not show up in the UI.
#### The solution
Retroactively extract them via a datascript.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation

### DIFF
--- a/GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
+++ b/GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
@@ -28,6 +28,7 @@ public class MediaResponse {
     private static final Logger log = Logger.getLogger(MediaResponse.class.getName());
     private static FlowJsonObjectReader jsonObjectReader = new FlowJsonObjectReader();
     private static FlowJsonObjectWriter jsonObjectWriter = new FlowJsonObjectWriter();
+    private static FlowJsonObjectWriter jsonObjectWriterNoNulls = new FlowJsonObjectWriter().withExcludeNullValues();
 
     public static final int VERSION_STRING = 0;
     public static final int VERSION_GEOTAGGING = 1;
@@ -86,9 +87,8 @@ public class MediaResponse {
      */
     public static String formatWithoutGeotag(Media media) {
             media.setLocation(null);
-            FlowJsonObjectWriter nonullWriter = new FlowJsonObjectWriter().withExcludeNullValues();
             try {
-                return nonullWriter.writeAsString(media);
+                return jsonObjectWriterNoNulls.writeAsString(media);
             } catch (IOException e) {
                 log.warning(e.getMessage());
                 return "";
--- a/GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
+++ b/GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
@@ -67,6 +67,35 @@ public class MediaResponse {
         return media.getFilename();
     }
 
+
+    /**
+     * Format media value as JSON with a geotag location, whether null or not.
+     */
+    public static String formatWithGeotag(Media media) {
+            try {
+                return jsonObjectWriter.writeAsString(media);
+            } catch (IOException e) {
+                log.warning(e.getMessage());
+                return "";
+            }
+        }
+
+
+    /**
+     * Format media value as JSON without a geotag location.
+     */
+    public static String formatWithoutGeotag(Media media) {
+            media.setLocation(null);
+            FlowJsonObjectWriter nonullWriter = new FlowJsonObjectWriter().withExcludeNullValues();
+            try {
+                return nonullWriter.writeAsString(media);
+            } catch (IOException e) {
+                log.warning(e.getMessage());
+                return "";
+            }
+        }
+
+
     public static Media parse(String value) {
         try {
             return jsonObjectReader.readObject(value, new TypeReference<Media>() {});
--- a/scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
@@ -64,7 +64,7 @@ import com.drew.metadata.iptc.IptcReader;
 
 
 /*
- * - Bring questions up to date
+ * - Bring question answers up to date
  */
 public class ExtractImageGeotag implements Process {
 
@@ -75,7 +75,6 @@ public class ExtractImageGeotag implements Process {
 
     @Override
     public void execute(DatastoreService ds, String[] args) throws Exception {
-
 
         if (args.length >= 3 ) {
             S3bucket = args[0];
@@ -89,9 +88,7 @@ public class ExtractImageGeotag implements Process {
             return;
         }
         //System.out.printf("#S3-bucketname %s, S3-id %s, S3-secret %s\n", S3bucket, S3id, S3secret);
-
         processQuestions(ds);
-
     }
 
     private void processQuestions(DatastoreService ds) throws ParseException {
@@ -139,13 +136,10 @@ public class ExtractImageGeotag implements Process {
                     tagged++;
                     continue; //Skip
                 }
-
-
             } else {
                 System.out.printf("#ERROR null or empty value for IMAGE %d: '%s'\n", q.getKey().getId(), v);
                 continue; //TODO convert this to JSON too?
             }
-
             //No location known; must read the file
             Location loc = new Location();
             Boolean tagFound = fetchLocationFromJpegInS3(media.getFilename(), loc);
@@ -171,7 +165,6 @@ public class ExtractImageGeotag implements Process {
                     questionsToFix.clear();
                 }
             }
-
         }
         System.out.println("Found " + total + " images.");
         System.out.println("  Non-JSON " + nonjson + " answers.");
@@ -212,8 +205,6 @@ public class ExtractImageGeotag implements Process {
                 Rational[] altTag = directory.getRationalArray(GpsDirectory.TAG_ALTITUDE);
                 Integer altRefTag = directory.getInteger(GpsDirectory.TAG_ALTITUDE_REF);
                 Rational[] accTag = directory.getRationalArray(GpsDirectory.TAG_H_POSITIONING_ERROR);
-                //Rational[] timTag = directory.getRationalArray(GpsDirectory.TAG_TIME_STAMP);
-                //Rational[] datTag = directory.getRationalArray(GpsDirectory.TAG_DATE_STAMP);
                 if (latTag == null || lonTag == null) {
                     f.delete();
                     return false; //Bad GPS tag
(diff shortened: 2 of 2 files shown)


scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,292 @@
...
+import java.util.logging.Level;
+
+import org.akvo.flow.util.FlowJsonObjectReader;
+import org.apache.commons.io.IOUtils;
+import org.waterforpeople.mapping.domain.SurveyInstance;
+import org.waterforpeople.mapping.domain.response.value.Location;
+import org.waterforpeople.mapping.domain.response.value.Media;
+import org.waterforpeople.mapping.serialization.response.MediaResponse;
+
+//import com.fasterxml.jackson.core.type.TypeReference;
```
valllllll2000: Why can't we remove it instead of commenting out?

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,292 @@
...
+import com.drew.imaging.jpeg.JpegSegmentMetadataReader;
+import com.drew.lang.Rational;
+import com.drew.metadata.Metadata;
+import com.drew.metadata.Directory;
+import com.drew.metadata.exif.GpsDirectory;
+import com.drew.metadata.Tag;
+import com.drew.metadata.exif.ExifReader;
+import com.drew.metadata.iptc.IptcReader;
+
+
```
valllllll2000: Maybe too many spaces? Maybe use the IDE automatic formatting shortcuts?

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,292 @@
...
+    private boolean doIt = false;
+    private String S3bucket;
+    private String S3id;
+    private String S3secret;
+
+    @Override
+    public void execute(DatastoreService ds, String[] args) throws Exception {
+
+
```
valllllll2000: Maybe too many spaces? Maybe use the IDE automatic formatting shortcuts?

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,292 @@
...
+            S3id = args[1];
+            S3secret = args[2];
+            if (args.length == 4 && args[3].equals("--doit")) {
+                doIt = true;
+            }
+        } else {
+            System.out.printf("#Arguments: S3-bucketname S3-id S3-secret [--doit to execute]\n");
+            return;
+        }
+        //System.out.printf("#S3-bucketname %s, S3-id %s, S3-secret %s\n", S3bucket, S3id, S3secret);
```
valllllll2000: How about removing that, we do not want to log the secrets right?

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
```diff
@@ -67,6 +68,34 @@ public static String format(String value, int version) {
...
+    public static String formatWithGeotag(Media media) {
+            try {
+                return jsonObjectWriter.writeAsString(media);
+            } catch (IOException e) {
+                log.warning(e.getMessage());
+                return "";
+            }
+        }
+
+
```
valllllll2000: Maybe too many spaces? Maybe use the IDE automatic formatting shortcuts?

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
```diff
@@ -67,6 +68,34 @@ public static String format(String value, int version) {
...
+            media.setLocation(null);
+            try {
+                return jsonObjectWriterNoNulls.writeAsString(media);
+            } catch (IOException e) {
+                log.warning(e.getMessage());
+                return "";
+            }
+        }
+
+
```
valllllll2000: Maybe too many spaces? Maybe use the IDE automatic formatting shortcuts?

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
```diff
@@ -28,6 +28,7 @@
     private static final Logger log = Logger.getLogger(MediaResponse.class.getName());
     private static FlowJsonObjectReader jsonObjectReader = new FlowJsonObjectReader();
     private static FlowJsonObjectWriter jsonObjectWriter = new FlowJsonObjectWriter();
+    private static FlowJsonObjectWriter jsonObjectWriterNoNulls = new FlowJsonObjectWriter().withExcludeNullValues();
```
valllllll2000: Does it mean we write null in a json file? I had issues in Android as it skipped null values so wanted to make sure, the method name does not seem clear.

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
stellanl: This is run from my command line, so nothing is logged

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,292 @@
...
+                    tagged++;
+                    continue; //Skip
+                }
+                //also want to skip if location is present, but null, to avoid re-evaluation
+                if (v.matches("\"location\":null")) {
+                    System.out.printf("Null location in IMAGE %d: '%s'\n", q.getKey().getId(), v);
+                    tagged++;
+                    continue; //Skip
+                }
+
```
valllllll2000: Maybe too many spaces? Maybe use the IDE automatic formatting shortcuts?

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,292 @@
...
+                    if (tagFound.equals(Boolean.FALSE)) {
+                        loc = null; //There is no tag!
+                    }
+                    media.setLocation(loc);
+                    v = MediaResponse.formatWithGeotag(media);
+                }
+                System.out.printf(" New IMAGE value '%s'\n", v);
+                q.setProperty("value", v);
+                questionsToFix.add(q);
+                if (doIt && questionsToFix.size() >= 250) {
```
valllllll2000: I am not 100% sure but when you retrieve 500 and only save 250, would it maybe make sense to just retrieve them in batches of 250 in line 111?

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,292 @@
...
+                }
+
+                Rational[] latTag = directory.getRationalArray(GpsDirectory.TAG_LATITUDE);
+                String latRefTag = directory.getString(GpsDirectory.TAG_LATITUDE_REF);
+                Rational[] lonTag = directory.getRationalArray(GpsDirectory.TAG_LONGITUDE);
+                String lonRefTag = directory.getString(GpsDirectory.TAG_LONGITUDE_REF);
+                Rational[] altTag = directory.getRationalArray(GpsDirectory.TAG_ALTITUDE);
+                Integer altRefTag = directory.getInteger(GpsDirectory.TAG_ALTITUDE_REF);
+                Rational[] accTag = directory.getRationalArray(GpsDirectory.TAG_H_POSITIONING_ERROR);
+                //Rational[] timTag = directory.getRationalArray(GpsDirectory.TAG_TIME_STAMP);
```
valllllll2000: If those are not used, I would remove them

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
stellanl: The batches are not equal. Only those values that make it past the filters end up in the toFix list.

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,292 @@
...
+
+            //No location known; must read the file
+            Location loc = new Location();
+            Boolean tagFound = fetchLocationFromJpegInS3(media.getFilename(), loc);
+            if (tagFound != null || forceSave) {
+
+                if (tagFound == null) { // We cannot know (right now - file may arrive later)
+                    v = MediaResponse.formatWithoutGeotag(media);
+                } else { //We do know!
+                    if (tagFound.equals(Boolean.FALSE)) {
```
valllllll2000: Would maybe !tagFound work?

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
```diff
@@ -0,0 +1,282 @@
...
+                System.out.printf("#ERROR null or empty value for IMAGE %d: '%s'\n", q.getKey().getId(), v);
+                continue; //TODO convert this to JSON too?
+            }
+            //No location known; must read the file
+            Location loc = new Location();
+            Boolean tagFound = fetchLocationFromJpegInS3(media.getFilename(), loc);
+            if (tagFound != null || forceSave) {
+
+                if (tagFound == null) { // We cannot know (right now - file may arrive later)
+                    v = MediaResponse.formatWithoutGeotag(media);
```
valllllll2000: Wouldn't this save the location value as null and the when file arrives it will be skipped?

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
stellanl: It would in this case since it is protected by the null check. But Mulo makes me change almost every == to and .equals() call, so I guess I stopped trying to be clever.

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
stellanl: No, this will save it with just the "filename": tag

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
stellanl: The noNull writer writes nothing in the json for null java values.

scripts/data/src/org/akvo/gae/remoteapi/ExtractImageGeotag.java
stellanl: We can.

GAE/src/org/waterforpeople/mapping/serialization/response/MediaResponse.java
stellanl: I like two empty lines between methods.

PR #3273, opened 2019-10-31
Issue 3235/add stats tab

marvinkome:
*Stats Lists*
image

*New stats*
image

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


stellanl:
The filename URL is not a link. Ideally, it should work like those in the Exports list.
The Stats reports still also show up in the Exports list.
The little floating "Loading" notification never goes away after running a report.

marvinkome:
The loading notification is due to the continuous polling when the export is queued

Dashboard/app/js/lib/components/stats/StatsLists.jsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class StatsList extends React.Component {
+  renderNoStats = () => {
+    return (
+      <div className="no-stats">
+        <p>No stats generated yet</p>
+        <p>Click {'"Export stats"'} to get started</p>
```
muloem: We need to use translation strings here

Dashboard/app/js/lib/components/stats/StatsLists.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+  renderStat = stat => {
+    return (
+      <a href={stat.url} key={stat.id}>
+        <div className="generated-stat">
+          <div className="stats-details">
+            <span className="stat-icon" />
+            <div>
+              <span className="date">
+                {stat.startDate} - {stat.endDate} Submissions
```
muloem: Ditto re translation string

muloem: What about the cases where one of either start or end date or even both are left out?  For now all we show is:

dates-not-provided

Dashboard/app/js/lib/components/stats/NewStats.jsx
```diff
@@ -0,0 +1,71 @@
...
+              onChange={this.onChange}
+              className="datePicker"
+              style={{ width: '95%' }}
+              min={this.state.startDate}
+            />
+          </label>
+        </div>
+
+        <button onClick={this.onSubmit} className="standardBtn" type="button">
+          Download Stats
```
muloem: Translation missing

Dashboard/app/js/lib/components/stats/StatsLists.jsx
marvinkome: @janagombitova and @Kiarii any help here?

Dashboard/app/js/lib/components/stats/StatsLists.jsx
```diff
@@ -0,0 +1,65 @@
...
+    return (
+      <div id="stats-listing-page">
+        <div className="page-header">
+          <h2>{this.props.strings.generatedStats}</h2>
+          <button
+            className="standardBtn newStats"
+            type="button"
+            onClick={this.props.goToExport}
+          >
+            {this.props.strings.exportStats}
```
muloem: Since you are already destructuring why not the rest of them?

Dashboard/app/js/lib/components/stats/StatsLists.jsx
janagombitova: Answering this not seeing the implementation yet: 

Defining the dates is optional so there is a high chance users will want to get the Stats from the beginning of time till today. I would suggest that: 
* if the start date is not defined we simply use 'from the beginning of time' that keeps it clear, generic and is a term used by other household tools so users know. 
* if the end date is not defined, then we should always use the current date (today) in the date format

Kiarii: we had mockups illustrating how the date picking would work by default; without the user having to enter anything; https://invis.io/9NUGWCM82MK#/389609091_0-_Stats_-_HOME_-_ES

marvinkome: @Kiarii can't find the mock-up that points to when no date is added

PR #3318, opened 2019-11-22
Issue 3297/redesign devices section

marvinkome:
#### Screenshots
1: When there's no device
image

2: Adding devices to an assignment: 
image

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/components/devices/AssignmentsEditView/DevicesSection.jsx
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable import/no-unresolved */
+import React from 'react';
+// import DeviceGroupSelectorView from 'akvo-flow/components/selectors/DeviceSelector';
```
muloem: Do we need this? if not remove instead of leaving commented out

Dashboard/app/js/lib/components/devices/AssignmentsEditView/__partials/DeviceEmpty.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class DeviceEmpty extends React.Component {
+  render() {
+    return (
+      <div className="device-empty">
+        <i className="fa fa-mobile" />
+        <p>{this.props.warningText}</p>
```
muloem:
```suggestion
        <p>{this.props.strings.warningText}</p>
```

Is it part of translated strings?

Dashboard/app/js/lib/components/selectors/DeviceSelector/old-index.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
muloem: if we dont need it anymore we should remove.

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
```diff
@@ -38,115 +32,61 @@ export default class AssignmentsEdit extends React.Component {
...
-          <div className="assignment-main">
-            <FormSection
-              actions={actions}
-              strings={strings}
-              data={data}
-              inputValues={this.state.data}
-              onInputChange={this.onChangeState}
-            />
-          </div>
+          <AssignmentContext.Provider value={contextData}>
```
muloem: I would have though that the context here has two parts.. the SideBar and the MainSection

Dashboard/app/js/lib/components/devices/AssignmentsEditView/Sidebar.jsx
```diff
@@ -0,0 +1,56 @@
...
+    currentTab: 'ADD_DEVICE',
+  };
+
+  changeTab = tab => {
+    this.setState({ currentTab: tab });
+  };
+
+  render() {
+    return (
+      <React.Fragment>
```
muloem: In this case I don't think we should use fragments since these are two sections that are related to each other but not exactly peers.

Dashboard/app/js/lib/components/devices/AssignmentsEditView/__partials/DeviceEmpty.jsx
marvinkome: No not a translation. A string will be passed into the component, so it can display whatever info you want.

Dashboard/app/js/lib/components/devices/AssignmentsEditView/index.jsx
marvinkome: The data is not for the sidebar. It's for the whole component. So it can be accessed from any component since they're lots of deeply nested components.

PR #3320, opened 2019-11-22
Issue 3297/list devices in sidebar

marvinkome:
Add UI to show devices in the sidebar

image

## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


Dashboard/app/js/lib/components/devices/AssignmentsEditView/__partials/SidebarDropdown.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class SidebarDropdow extends React.Component {
```
muloem:
```suggestion
export default class SidebarDropdown extends React.Component {
```

Dashboard/app/js/lib/components/devices/AssignmentsEditView/__partials/SidebarDropdown.jsx
```diff
@@ -0,0 +1,53 @@
...
+          {devices.map(device => (
+            <button type="button">{device.name}</button>
+          ))}
+        </div>
+      </div>
+    );
+  }
+}
+
+SidebarDropdow.propTypes = {
```
muloem:
```suggestion
SidebarDropdown.propTypes = {
```

Dashboard/app/js/lib/components/devices/AssignmentsEditView/__partials/SidebarDropdown.jsx
```diff
@@ -40,14 +40,16 @@ export default class SidebarDropdow extends React.Component {
 
         <div style={panelStyle} className="sidebar-panel">
           {devices.map(device => (
-            <button type="button">{device.name}</button>
+            <button key={device.id} type="button">
```
muloem: One more thought here is that the list of devices is probably links rather than buttons no?

marvinkome: Well yes. But links need `href` links. And we don't have them

muloem: Maybe we can implement for now without the href but with associated actions. So whenever we replace the router we fill in the links?

marvinkome: That works, but `eslint` will complain about it

muloem: lets disable the complaint for now with a note to reenable it when links are in place.

muloem:
Also the issue regarding the alignment of devices is back. i.e. they dont appear one after the other in a list., they all appear on the same line.

muloem:
Approved. Ignore the above comments as they were based on an outdated branch.

Issue #3317, opened 2019-11-21
Data export missing rows


janagombitova:
Reported via support: https://flowhelp.reamaze.com/admin/conversations/urgent-error-uploading-data-from-mobile-devices

The dataset (see all details below) in Flow exports 327 rows (in the Data cleaning export) but in Lumen it shows 644 rows. This brings up the question: Is Flow exports exporting all the data?

Details:
* https://nlrc.akvoflow.org/admin/
* Cote d'Ivoire > Survey_Pneumonie/IRA > IRA_CAP_FINALE_MENAGE > CAP_MENAGE
* Form ID: 458910091

stellanl:
Running that export in the debugger, it fetches a list of 644 instance ids. But when it goes to fetch the data for them, half of them fail with a 500.

stellanl:
Looking in the logs, the error is

2019-11-25 10:39:50.286 CET
com.gallatinsystems.framework.rest.AbstractRestApiServlet executeRequest: Could not execute rest request (AbstractRestApiServlet.java:95)
java.lang.IllegalArgumentException: displayName: String properties must be 1500 bytes or less.  Instead, use com.google.appengine.api.datastore.Text, which can store strings of any length.

so displayname is apparently overflowing 1500 chars.

stellanl:
Looking in the form, 5 of the 6 questions in the 1st group are used to form the display name. So it gets very long.

stellanl:
Fixing the crash is relatively easy. I suggest truncating the displayName at 1500 chars when necessary, possibly adding "..." to the end.

janagombitova:
Sounds like a great solution to me @stellanl this way we do not need users to make any changes on their end.

valllllll2000:
I am not sure what is supposed to happen so the DATA cleaning report is generated but the name is wrong:
Here is the correct name in the app:
![Screenshot_20191217-155145](https://user-images.githubusercontent.com/923280/71006284-65753080-20e5-11ea-9f26-5774e0448483.png)
[DATA_CLEANING-579479115.xlsx](https://github.com/akvo/akvo-flow/files/3973854/DATA_CLEANING-579479115.xlsx)

I am attaching the data cleaning report as you can see the last line the name is -5 as if the text question has been skipped

stellanl:
The text is in the **valueText** field since it is too large for **value** (>500 chars). Only short values get assembled into the displayName.

valllllll2000:
[...]

PR #3323, opened 2019-11-25
[#3317]Go via setter for truncation check. Use modern 1500-char limit.

stellanl:
Too much content for the display name made it overflow the 1500 char storage limit for String. Length check was bypassed by assembleDsiplayName() method.
#### The solution
Fix that, and bump the check from old 500-char limit.
## Checklist
* [ ] Connect the issue
* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting
* [ ] Documentation


stellanl:
We now have a bit of a mismatch between answers and displayName. Answers > 500 chars go in the valueText field, which is not considered when assembling the display name.

Issue #3343, opened 2019-12-06
Allow filtering data points on prefix of displayName


stellanl:
This is to support #3297.
Currently, the /rest/surveyed_locales endpoint allows filtering by 
surveyGroupId, identifier or displayName. These must match exactly.
To allow practical searching by displayName while assigning data points to devices, we could add a parameter that allows getting all data points whose display name **begins** with _string_.
So if the parameter is "ABC" we would return all DPs with displayName >= "ABC" and displayName <"ABD".

stellanl:
Test plan:

1. Run Flow locally.
2. From a browser, connect to http://akvoflow.local:8888/_ah/admin/datastore, look in the surveyedLocales entities to see what displayNames exist.
3. From a browser, connect to http://akvoflow.local:8888/rest/surveyed_locales?displayNamePrefix= (and a prefix of a displayName you know from 2. to exist.)
4. You should get a JSON list of data points.

stellanl:
In a conversation w @muloem and @marvinkome, agreed to add further "search" parameter to try to filter data points by _either_ identifier or display name.

stellanl:
Also, the search needs to filter by (UI)survey.

muloem:
@stellanl I just tried testing on a branch where the UI is implemented and I get back the entire list of datapoints from the backend instead of a filtered list

![search-datapoints](https://user-images.githubusercontent.com/533336/73756549-0d16f280-4768-11ea-863b-eacf10eec8ca.gif)

PR #3345, opened 2019-12-09
Issue/3343 search dps by displayname prefix (Connect #3343)

stellanl:
There was no way to search for data points with a prefix of the displayName matching a string. 
#### The solution
Add one.
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/com/gallatinsystems/surveyal/dao/SurveyedLocaleDao.java
+++ b/GAE/src/com/gallatinsystems/surveyal/dao/SurveyedLocaleDao.java
@@ -40,8 +40,6 @@ import com.gallatinsystems.surveyal.domain.SurveyedLocale;
  */
 public class SurveyedLocaleDao extends BaseDAO<SurveyedLocale> {
 
-    private static final String LT_OP = "<";
-
     public SurveyedLocaleDao() {
         super(SurveyedLocale.class);
     }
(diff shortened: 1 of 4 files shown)


GAE/src/com/gallatinsystems/surveyal/dao/SurveyedLocaleDao.java
```diff
@@ -40,6 +40,8 @@
  */
 public class SurveyedLocaleDao extends BaseDAO<SurveyedLocale> {
 
+    private static final String LT_OP = "<";
```
iperdomo: What do you think if we move this constant to `BaseDAO` similar to the rest of them?

GAE/src/com/gallatinsystems/surveyal/dao/SurveyedLocaleDao.java
```diff
@@ -289,8 +291,16 @@ public SurveyedLocaleDao() {
                 "String", identifier, paramMap);
         appendNonNullParam("displayName", filterString, paramString,
                 "String", displayName, paramMap);
+        if (displayNamePrefix != null && !"".equals(displayNamePrefix)) {
+            String endString = displayNamePrefix + Character.MAX_VALUE;
+            appendNonNullParam("displayName", filterString, paramString, "String", displayNamePrefix, paramMap, GTE_OP);
+            appendNonNullParam("displayName", filterString, paramString, "String", endString, paramMap, LT_OP);
```
iperdomo: If we're filtering by a display name _prefix_, I fail to understand why we need to add this extra filter clause :sweat: 
What is the relationship between the operators `<` `>=` and the content of the string? Have we tried this with displayName containing non ASCII characters?

GAE/src/com/gallatinsystems/surveyal/dao/SurveyedLocaleDao.java
stellanl: Sure

GAE/src/com/gallatinsystems/surveyal/dao/SurveyedLocaleDao.java
stellanl: Otherwise we do not filter by prefix. Just `>= 'ABC'` will hit ABC*, ABD*...ABZ* and beyond, whereas `>= ABC && < ABC&#xffffffff;` willl hit exactly the strings starting with ABC. And there are hyphens in the test datastore and I tried filtering by those...

stellanl: I could add some äöñ in a test datastore if you think it is weird.

stellanl: Tests ok with accented and cyrillic characters.

PR #3367, opened 2019-12-19
Issue/3343 search DPs by name and identifier

stellanl:
Data points could only be searched for by displayname prefix.
All surveys were searched.
#### The solution
Filter by survey.
Also try to match searchstring to identifier if correct format.

#### Reviewer Checklist
* [x] Added an explanation about the work done
* [x] Connected the PR and the issue on Zenhub
* [x] Added a test plan to the issue
* [ ] Formatted the code
* [ ] Added documentation (if relevant)


GAE/src/com/gallatinsystems/surveyal/dao/SurveyedLocaleDao.java
```diff
@@ -296,6 +296,8 @@ public SurveyedLocaleDao() {
             appendNonNullParam("displayName", filterString, paramString,
                     "String", endString, paramMap, LT_OP);
             query.setOrdering("displayName asc"); //necessary if inequality ops present
+            //Alternative custom filterString:
+            // surveyGroupId == surveyGroupIdParam0 && displayName.startsWith(displayNameParam1)
```
valllllll2000: it's commented out, will this work?

stellanl: Yes, this shows an alternative implementation. Something that would also work.

Issue #3283, opened 2019-11-08
Remove Auth0 feature flag


janagombitova:
As we released Auth0 a few weeks ago and all is going well, we should have it as the default log in implementation and remove all the code and other objects that we do not need or use anymore. Plus we should ensure that also all dev Flow instances are running on Auth0 (uat2, for example, isn't)

PR #3389, opened 2020-01-14
[#3283] Remove featureflag package

iperdomo:
We want to keep 2 authentication methods for now. ODIC based (Auth0)
in Production, and GAE accounts for local development


dlebrero:
Is this a subset of the work to remove the feature flag?

iperdomo:
We're not going to remove the 2 auth methods, so I removed the _misleading_ `featureflag` package.

muloem:
Can you also switch to make the auth0 the default method. Right now it is the alternative entry point?

iperdomo:
I've tested the changes. In development we use GAE accounts, and OIDC in production. - Right now this branch is deployed in `akvoflowsandbox`

GAE/src/org/akvo/flow/rest/security/EntryPoint.java
```diff
@@ -25,26 +25,24 @@
...
+    public EntryPoint(AuthenticationEntryPoint prodEntryPoint, AuthenticationEntryPoint devEntryPoint) {
+        this.devEntryPoint = devEntryPoint;
+        this.prodEntryPoint = prodEntryPoint;
     }
 
     @Override
     public void commence(HttpServletRequest request, HttpServletResponse response, AuthenticationException authException) throws IOException, ServletException {
-        if ("auth0".equals(defaultAuthProvider)) {
-            alternativeEntryPoint.commence(request, response, authException);
+        if (SystemProperty.environment.value() == SystemProperty.Environment.Value.Production) {
```
dlebrero: Explicit configuration please, so that developers can chose which auth method to use locally.

GAE/war/WEB-INF/webapp-security.xml
```diff
@@ -24,13 +24,6 @@
...
    </b:bean>
 
-   <http pattern="/altlogin" use-expressions="true" entry-point-ref="alternativeEntryPoint" authentication-manager-ref="authenticationManager"
-         access-decision-manager-ref="accessDecisionManager">
-      <intercept-url pattern="/**" access="hasRole('ROLE_USER')"/>
-      <custom-filter position="PRE_AUTH_FILTER" ref="gaeFilter"/>
-      <csrf disabled="true" />
-   </http>
-
     <http use-expressions="true" entry-point-ref="defaultEntryPoint" authentication-manager-ref="authenticationManager"
```
dlebrero: Maybe we can remove the defaultEntryPoint and use Spring's PropertyPlaceholderBlaBlabla... to configure the bean to use here, using as default value the auth0 bean (https://stackoverflow.com/questions/2513484/is-there-a-way-to-specify-a-default-property-value-in-spring-xml).

AFAIK, there is not PropertyPlaceholderBlaBlabla... yet in the Spring config, so it should be added if we opt for this.

Issue #3380, opened 2020-01-07
Use pre-built Docker image for development


iperdomo:
Instead of building the image locally we can reuse the pre-built image
See https://github.com/akvo/akvo-dockerfiles/issues/21

PR #3392, opened 2020-01-15
[#3380] Update dev image + code clean up

iperdomo:
- Update to the latest akvo/akvo-flow-builder image that is based
  on a published google/cloud-sdk
- Remove references to ruby build process, e.g. RAKEP_MODE BUNDLE etc
  We're now using `node` to bundle our client application
- Use `npm ci` instead of `npm install` as part of the build
  https://docs.npmjs.com/cli/ci.html
- Cache `$HOME/.npm` to speed up builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ services:
 - docker
 cache:
   directories:
-    - $HOME/.m2
+    - "$HOME/.m2"
+    - "$HOME/.npm"
 
 script:
   - "./ci/bootstrap-build.sh /app/src/ci/build.sh && ./ci/deploy.sh"
--- a/ci/bootstrap-build.sh
+++ b/ci/bootstrap-build.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 #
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -5,7 +5,8 @@ services:
 - docker
 cache:
   directories:
-    - $HOME/.m2
+    - "$HOME/.m2"
+    - "$HOME/.npm"
```
dlebrero: For my information, what is stored here? I thought npm does not use a global path, but just messes around with the node_modules dir

iperdomo: From the documentation page
````
# keep the npm cache around to speed up installs
cache:
  directories:
  - "$HOME/.npm"
````
https://docs.npmjs.com/cli/ci.html

In a fresh container:
````
cd $HOME/.npm/
~/.npm # du -sh
43.6M	.
````
The actual content can be found here:
https://gist.github.com/iperdomo/042af3dafe282012e3cb66230a51da2b

`npm ci` deletes the local `node_modules` if exists, before installing the versions used in `package-lock.json`

Issue #3168, opened 2019-08-21
Selection of datapoints to assign to a device


muloem:
As part of #2796, we will enable selection of datapoints to assign to a specific device in an assignment.  This is the UI work related to #3166 that will communciate with the backend to create and update the list of datapoints assigned to a device.

Initial MVP simply shows a list of datapoints selected if present or none if they are not selected. 

<img width="989" alt="Screenshot 2019-08-21 at 16 47 16" src="https://user-images.githubusercontent.com/533336/63442419-5f22c080-c433-11e9-9fd0-aa8970c549fb.png">

And a layover? when _Assign datapoints_ is selected to list all the datapoints to make a selection of the ones to be assigned to a device.

janagombitova:
For all the details check: #2796 

With 1.10 we changed how Assignments work to ensure one assignment can only hold forms from one survey and we redesigned the form and device selectors. These changes were needed to implement **the MVP of allowing to assign data points to devices**. 

The expected user behaviour that we will support is: 
1. Salim creates a new assignment - gives it a name, dates and selects the survey and forms and which devices will collect data for this/these forms. 
2. IMPORTANT: If Salim selects a monitoring survey (with monitoring enabled) he can also assign data points to the selected devices. If he selected a survey that does not have monitoring enables, assigning data points is not available to him. 
3. He selects a device or an entire device group. It is highlighted in the UI. 
4. He clicks to Assign data points (for the selected device or device group). 
5. He sees two options: By data point name or ID, Assign all data points. 
6. He selects how he will assign his data points. @Kiarii how will we handle if there are no data points yet to assign in this MVP? 
6a. If he selects to Assign all data points, all the data points will be sent to that selected device. 
6b. If he selects By data point name or id, he will see a search bar when he types the name or ID and matching data points show in the list with the option to assign to the device.

For all the details please check the mock ups. You can find the full mockups here: https://projects.invisionapp.com/share/NHQGVLB2EKX#/screens/346093699_0-0_Assignments_ES

Kiarii:
> He selects how he will assign his data points. @Kiarii how will we handle if there are no data points yet to assign in this MVP?

@janagombitova am thinking that we had storymapped this or specified somewhere else what would take place, am looking Mural but cant seem to find what we had decided/put down,

janagombitova:
[...]

PR #3396, opened 2020-01-17
[#3168] - save datapoints in api

marvinkome:
Datapoints were only stored in memory and not saved in the data store.

#### The solution
Store datapoints with the API and fetch back datapoint assignments when editing assignments

#### Reviewer Checklist
* [ ] Formatted the code


Dashboard/app/js/lib/components/devices/AssignmentsEditView/screens/AddDevices.jsx
```diff
@@ -16,9 +16,7 @@ export default class AddDevice extends React.Component {
     // filter out selected devices
     const { devices, selectedDeviceIds } = this.context.data;
 
-    const filteredDevices = devices.filter(
-      device => !selectedDeviceIds.includes(device.id)
-    );
+    const filteredDevices = devices.filter(device => !selectedDeviceIds.includes(device.id));
```
muloem:
```suggestion
    const unSelectedDevices = devices.filter(device => !selectedDeviceIds.includes(device.id));
```

muloem: or something more descriptive of why the filtering is happening

Dashboard/app/js/lib/components/devices/AssignmentsEditView/screens/EditDevices.jsx
```diff
@@ -16,9 +16,7 @@ export default class EditDevices extends React.Component {
     // filter out selected devices
     const { devices, selectedDeviceIds } = this.context.data;
 
-    const filteredDevices = devices.filter(device =>
-      selectedDeviceIds.includes(device.id)
-    );
+    const filteredDevices = devices.filter(device => selectedDeviceIds.includes(device.id));
```
muloem: Did you merge the code from the previous PR? I thought this one had changed...
```suggestion
    const selectedDevices = devices.filter(device => selectedDeviceIds.includes(device.id));
```

Dashboard/app/js/lib/components/devices/AssignmentsEditView/screens/EditDevices.jsx
```diff
@@ -43,9 +41,7 @@ export default class EditDevices extends React.Component {
     if (checked) {
       newSelectedDevices = selectedDevices.concat(ids);
     } else {
-      newSelectedDevices = selectedDevices.filter(
-        device => !ids.includes(device)
-      );
+      newSelectedDevices = selectedDevices.filter(device => !ids.includes(device));
```
muloem:
```suggestion
      newSelectedDeviceIds = selectedDevicesIds.filter(deviceId => !ids.includes(deviceId));
```

Dashboard/app/js/lib/components/devices/AssignmentsEditView/screens/EditDevices.jsx
marvinkome: This is different

marvinkome:
There's a bug where Datapoints isn't being saved because `surveyAssignment` keyId isn't yet available when creating new `surveyAssignment`, this bug was fixed in this PR #3404

Dashboard/app/js/lib/router/router.js
```diff
@@ -231,9 +227,15 @@ FLOW.Router = Ember.Router.extend({
...
-          router
-            .get('navDevicesController')
-            .connectOutlet('editSurveyAssignment');
+          router.get('navDevicesController').connectOutlet('editSurveyAssignment');
+
+          // if editing fetch datapoints assignments,
+          // else well, no need for datapoints assignments in a new assignments
+          const assignmentId = FLOW.selectedControl.selectedSurveyAssignment.get('keyId');
+          if (assignmentId) {
+            FLOW.dataPointAssignmentControl.populate(assignmentId);
```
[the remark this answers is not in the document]
marvinkome: I'm loading them here to be able to get each data point assignment when it's needed rather than making another API request when in the assignment. And also other checks like, when changing survey folder the datapoints assignments need to be removed, If they're not loaded it wouldn't know what to remove.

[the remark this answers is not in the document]
muloem: Then maybe its enough to skip them if there is nothing to remove. and remove if there is something.  So right now you go to the assignments page, select an assignment to edit and there you see a form and a list of devices that are part of that assignment.  however, its only relevant to have information on the datapoint assignment if you actually select a specific device to look at. if you never select a device then we have spent resources and time loading up datapoint assignments which will not be used.  So I can understand the reasoning but I think it will slow down the performance unnecessarily.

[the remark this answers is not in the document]
marvinkome: I'm loading them here to be able to get each data point assignment when it's needed rather than making another API request when in the assignment. And also other checks like, when changing survey folder the datapoints assignments need to be removed, If they're not loaded it wouldn't know what to remove.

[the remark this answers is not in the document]
marvinkome: True, I see your point. So that means, only load datapoint related to a device when the user clicks on it.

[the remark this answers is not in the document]
muloem: Yes exactly 👍

Dashboard/app/js/lib/components/devices/AssignmentsEditView/screens/AssignDatapoints/index.jsx
```diff
@@ -20,7 +20,9 @@ export default class AssignDatapoints extends React.Component {
     const { devices, datapointAssignments } = this.context.data;
 
     const deviceData = devices.find(device => device.id === selectedDeviceId);
-    const datapointAssignment = datapointAssignments.find(dp => dp.deviceId === selectedDeviceId);
+    const datapointAssignment = datapointAssignments.find(
+      dp => dp.deviceId === parseInt(selectedDeviceId, 10)
```
muloem: Instead of doing this I think its much clearer to set the property to a number. i.e. 

```
AssignDatapoints.propTypes = {
  selectedDeviceId: PropTypes.number.isRequired,
};
```

Dashboard/app/js/lib/views/devices/assignment-edit-views.jsx
```diff
@@ -706,7 +706,10 @@ FLOW.AssignmentEditView = FLOW.ReactComponentView.extend(
       this.renderReactSide();
     },
 
-    assignDataPointsToDevice(datapoints, deviceId) {
+    assignDataPointsToDevice(datapoints, deviceIdInString) {
+      // convert devieId to number
+      const deviceId = parseInt(deviceIdInString, 10);
```
muloem: Same here regarding setting the PropType

Dashboard/app/js/lib/components/devices/AssignmentsEditView/screens/AssignDatapoints/index.jsx
marvinkome: Yeah, But in #3404 it's already converted to a number, this is like a minor hack to make sure this PR works

PR #3439, opened 2020-02-24
Issue/3274 filter report list by type

stellanl:
You had to fetch all report items
#### The solution
Let UI filter by type
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/org/akvo/flow/rest/ReportRestService.java
+++ b/GAE/src/org/akvo/flow/rest/ReportRestService.java
@@ -31,6 +31,8 @@ import org.akvo.flow.rest.dto.ReportPayload;
 import org.akvo.flow.rest.security.AppRole;
 import org.akvo.flow.servlet.ReportServlet;
 import org.springframework.beans.BeanUtils;
+import org.springframework.security.core.Authentication;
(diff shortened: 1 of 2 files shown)


GAE/src/org/akvo/flow/rest/ReportRestService.java
```diff
@@ -123,10 +123,17 @@ private boolean isAdminRole() {
...
-    public Map<String, Object> listMyReports() {
+    public Map<String, Object> listMyReports(
+            @RequestParam(value = "reportType", defaultValue = "") String reportType
+            ) {
         final Map<String, Object> response = new HashMap<String, Object>();
         List<ReportDto> results = new ArrayList<ReportDto>();
-        List<Report> reports = reportDao.listAllByCurrentUser();
+        List<Report> reports;
+        if (reportType.equals("")) { //Cannot have null as the defaultValue
+            reportType = null;
```
valllllll2000: Why don't we just pass empty string and then check for empty in the method that uses it?

stellanl: Just for consistency. In most places we pass null to the dao methods to indicate a param should be ignored.

Issue #3451, opened 2020-03-03
Implement a flag to enable running servlet requests as tasks


muloem:
# Context
In some cases we have requests to servlets that timeout because they hit the one minute deadline by GAE.  We typically resolve this by using the frontend request to create a task that then performs the real work.

# Idea
Instead of writing new code for tasks everytime one is required, an idea is to include a flag in a request that when included, redirects the request to be executed as a task.  This has to be for tasks that do not require results of the task execution to be returned immediately, e.g. generating reports or importing data to the backend.

PR #3452, opened 2020-03-04
[#3451] Enable redirecting a servlet request to run as a task

muloem:
### Before the PR (what is the issue or what needed to be done)
Some requests time out due to hitting the one minute deadline on GAE

#### The solution
We add the possibility to create a task by simply adding a parameter to the request.  This works for servlets that are extended from com.gallatinsystems.framework.rest.AbstractRestApiServlet.  Also right now we support only the ones that are secured.

#### Reviewer Checklist
* [ ] Formatted the code


muloem:
> Why the method is always POST?

Since the GAE tasks typically run in the background and their purpose is usually to update or generate something, I chose to use `POST`. Tasks never actually return something to the original requester.

PR #3464, opened 2020-03-11
[#3451] Implement request to task filter

muloem:
* Request to task conversion handled in a servlet

#### The solution
* We intercept the request in a filter so that we don't process / parse it unnecessarily since it will go to a task.

#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/pom.xml
+++ b/GAE/pom.xml
@@ -227,12 +227,6 @@
         <version>${spring.version}</version>
         <scope>test</scope>
     </dependency>
-    <dependency>
-        <groupId>org.mockito</groupId>
-        <artifactId>mockito-core</artifactId>
-        <version>3.3.0</version>
-        <scope>test</scope>
-    </dependency>
 
       <!-- datastore unit tests -->
     <dependency>
--- a/GAE/pom.xml
+++ b/GAE/pom.xml
@@ -221,7 +221,20 @@
       <scope>test</scope>
     </dependency>
 
-    <!-- datastore unit tests -->
+    <dependency>
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-test</artifactId>
+        <version>${spring.version}</version>
+        <scope>test</scope>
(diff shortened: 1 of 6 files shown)


muloem:
> * I don't think we need a filter, and the way to _break_ a potential infinite loop is by adding a HTTP header `X-Akvo-Run-As-Task=1` that will be added by the first request. The Servlet checks the existence of it: If present, just execute the logic, if not present forward request the a Task and **add** the header.

Regardless of whether we take your proposed approach or the one currently implemented, I think it actually does make sense to use a filter so that we don't go through unnecessary processing before deciding that a request will be redirected to a task.

> * We're changing the _semantics_ from `GET` to `POST` for some reason that I don't really understand :confused: 

The semantics of GET is expecting to return something but a task is a type of fire and forget and usually its operations are to do with updating or creating a resource so I believe it fits with the semantics of POST/PUT.

> * We don't need to re-implement the hashing calculation again. Just _forward_ the request to the Task as it comes

Fair point, however, we still need to unpack the request on order to forward it. We get either POST or GET requests and only in the latter case do we have access to the parameters via `HttpServletRequest.getQueryString()`

iperdomo:
I didn't know that we could get `POST` requests, only `GET`. 

What about _honor_ the initial request method? for `GET` `getQueryString()` for `POST` read the body and pass it via `payload`

* https://stackoverflow.com/questions/8100634/get-the-post-request-body-from-httpservletrequest/31848289#31848289
* https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/taskqueue/TaskOptions.html#payload-java.lang.String-

muloem:
> What about honor the initial request method? for GET getQueryString() for POST read the body and pass it via payload

I think its not necessary.  We are modifying the request and the semantics of that modification are POST semantics. Otherwise what reason do we have for maintaining a GET request in these cases.

PR #3474, opened 2020-03-17
Issue/3469 all datapoints

iperdomo:
Restore behavior of serving all data points for those assignments that doesn't have a `DataPointAssignment` entities

### DIFF
--- a/GAE/src/org/akvo/flow/api/app/DataPointServlet.java
+++ b/GAE/src/org/akvo/flow/api/app/DataPointServlet.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2019 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2019,2020 Stichting Akvo (Akvo Foundation)
  *
  *  This file is part of Akvo FLOW.
  *
@@ -23,23 +23,25 @@ import com.gallatinsystems.framework.rest.RestRequest;
 import com.gallatinsystems.framework.rest.RestResponse;
 import com.gallatinsystems.surveyal.dao.SurveyedLocaleDao;
 import com.gallatinsystems.surveyal.domain.SurveyedLocale;
-import org.akvo.flow.api.app.DataPointServlet;
 import org.akvo.flow.dao.DataPointAssignmentDao;
+import org.akvo.flow.dao.SurveyAssignmentDao;
 import org.akvo.flow.domain.persistent.DataPointAssignment;
+import org.akvo.flow.domain.persistent.SurveyAssignment;
 import org.akvo.flow.util.FlowJsonObjectWriter;
 import org.waterforpeople.mapping.app.web.dto.SurveyedLocaleDto;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-
 import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 
 /**
@@ -48,9 +50,9 @@ import java.util.logging.Logger;
 @SuppressWarnings("serial")
 public class DataPointServlet extends AbstractRestApiServlet {
     private static final Logger log = Logger.getLogger(DataPointServlet.class.getName());
+    private static Set<Long> ALL_DATAPOINTS = new HashSet<>(Arrays.asList(0L));
     private SurveyedLocaleDao surveyedLocaleDao;
     private DataPointAssignmentDao dataPointAssignmentDao;
-    private static Set<Long> ALL_DATAPOINTS = new HashSet<>(Arrays.asList(0L));
 
     public DataPointServlet() {
         setMode(JSON_MODE);
@@ -74,7 +76,6 @@ public class DataPointServlet extends AbstractRestApiServlet {
     @Override
     protected RestResponse handleRequest(RestRequest req) throws Exception {
         DataPointRequest dpReq = (DataPointRequest) req;
-        List<SurveyedLocale> dpList;
         RestResponse res = new RestResponse();
         if (dpReq.getSurveyId() != null) {
             //Find the device (if any)
@@ -83,22 +84,7 @@ public class DataPointServlet extends AbstractRestApiServlet {
             if (device != null) {
                 log.info("Found device id: " + device.getKey().getId());
                 log.fine("Found device: " + device);
-                //Find which assignments we are part of
-                List<DataPointAssignment> assList =
-                        dataPointAssignmentDao.listByDeviceAndSurvey(device.getKey().getId(), dpReq.getSurveyId());
-                //Combine their point lists
-                Set<Long> pointSet = new HashSet<>();
-                for (DataPointAssignment ass : assList) {
-                    pointSet.addAll(ass.getDataPointIds());
-                }
-                //Fetch the data points
-                if (ALL_DATAPOINTS.equals(pointSet)) {
-                    dpList = surveyedLocaleDao.listLocalesBySurveyGroupId(dpReq.getSurveyId());
-                } else {
-                    List<Long> pointList = new ArrayList<>();
-                    pointList.addAll(pointSet);
-                    dpList = surveyedLocaleDao.listByKeys(pointList);
-                }
+                List<SurveyedLocale> dpList = getDataPointList(device.getKey().getId(), dpReq.getSurveyId());
                 res = convertToResponse(dpList, dpReq.getSurveyId());
                 return res;
             }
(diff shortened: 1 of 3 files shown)


GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        dpa.setDeviceId(deviceId);
+        dpa.setDataPointIds(dataPointIds);
+        dpa.setSurveyAssignmentId(assignmentId);
+        dpa.setSurveyId(surveyId);
+
+        return dao.save(dpa);
+    }
+
+    private SurveyAssignment createAssingment(Long surveyId, List<Long> deviceIds, List<Long> formIds) {
```
muloem: small typo.

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
```diff
@@ -0,0 +1,205 @@
...
+        final Long surveyId = 1L;
+
+        createDataPointAssignment(assignmentId, deviceId, dataPointIds, surveyId);
+
+        final DataPointAssignmentDao dataPointAssignmentDao = new DataPointAssignmentDao();
+        final List<DataPointAssignment> dataPointAssignments = dataPointAssignmentDao.listByDeviceAndSurvey(deviceId, surveyId);
+        assertFalse(dataPointAssignments.isEmpty());
+
+        final DataPointAssignment assignment = dataPointAssignments.get(0);
+        assertEquals(assignment.getDataPointIds(), dataPointIds);
```
muloem: The `assertEquals` params should be reversed I believe. `assertEquals(expected, actual)`

GAE/src/org/akvo/flow/api/app/DataPointServlet.java
```diff
@@ -111,6 +97,39 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
...
+        if (assList.isEmpty()) {
+            // Mimic old behavior
+            SurveyAssignmentDao saDao = new SurveyAssignmentDao();
+
+            List<SurveyAssignment> allAssigmentsToDevice = saDao.listAllContainingDevice(deviceId);
+            List<SurveyAssignment> filteredAssignments = allAssigmentsToDevice.stream().filter(assignment ->
+                    assignment.getSurveyId().equals(surveyId)).collect(Collectors.toList());
+
+            if (filteredAssignments.isEmpty()) {
+                return Collections.emptyList();
```
muloem: This branch should be an exceptional one.. maybe we should log some kind of error here.  There is a survey on a device but when attempting to retrieve its assignment we find nothing

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
```diff
@@ -0,0 +1,214 @@
...
+        createAssignment(surveyId, deviceIds, formIds);
+
+        final SurveyAssignmentDao saDao = new SurveyAssignmentDao();
+        final List<SurveyAssignment> surveyAssignments = saDao.listAllContainingDevice(2L);
+        assertFalse(surveyAssignments.isEmpty());
+
+        final SurveyAssignment sa = surveyAssignments.get(0);
+        assertEquals(surveyId, sa.getSurveyId());
+        assertTrue(sa.getDeviceIds().contains(2L));
+        assertEquals(formIds, sa.getFormIds());
```
dlebrero: Does this part of the test has anything to do with the previous part about `dataPointAssignmentDao`? If not, move to another tests

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @Test
+    public void someDataPointsTest() {
+
+        final Long surveyId = 100L;
+        final List<SurveyedLocale> dataPoints = createDataPoints(surveyId, 10);
+
+        final SurveyedLocale dataPoint = dataPoints.get(7);
+        assertEquals("7", dataPoint.getIdentifier());
```
dlebrero: Why is this important for this test?

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        final List<Long> selectedDataPointIds = Arrays.asList(dataPoints.get(5).getKey().getId(), dataPoints.get(6).getKey().getId());
+        final Long assignmentId = 20L;
+        final Long deviceId = 1L;
+
+        createDataPointAssignment(assignmentId, deviceId, selectedDataPointIds, surveyId);
+
+        final DataPointServlet servlet = new DataPointServlet();
```
dlebrero: This use of `servlet` was really unexpected. I was expecting all the tests to be testing at the same layer, so either the DAO layer or the servlet layer.

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        final Set<Long> selectedIds = new HashSet<>(selectedDataPointIds);
+        final Set<Long> foundIds = getEntityIds(someDataPoints);
+
+        assertEquals(selectedIds, foundIds);
+    }
+
+
+    /**
+     * New schema: `[0]` as data point list represent all data points
```
dlebrero: This special value is not obvious where it is used in the tests. 

Also, try to remove comment by making the test say the same.

dlebrero: I really dislike special values like [0]. They always cause problems. 

Did you discard the option of adding a flag in the survey assignment objects?

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        final SurveyAssignment sa = createAssignment(surveyId, deviceIds, formIds);
+
+        final DataPointServlet servlet = new DataPointServlet();
+        final List<SurveyedLocale> foundDataPoints = servlet.getDataPointList(surveyId, 6L);
+
+        final Set<Long> allDataPointIds = getEntityIds(allDataPoints);
+        final Set<Long> foundDataPointIds = getEntityIds(foundDataPoints);
+
+        assertEquals(allDataPointIds, foundDataPointIds);
```
dlebrero: This is duplicated several times.

GAE/src/org/akvo/flow/api/app/DataPointServlet.java
```diff
@@ -83,21 +84,11 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
...
-                //Fetch the data points
-                if (ALL_DATAPOINTS.equals(pointSet)) {
-                    dpList = surveyedLocaleDao.listLocalesBySurveyGroupId(dpReq.getSurveyId());
-                } else {
-                    List<Long> pointList = new ArrayList<>();
-                    pointList.addAll(pointSet);
-                    dpList = surveyedLocaleDao.listByKeys(pointList);
+                List<SurveyedLocale> dpList = getDataPointList(device.getKey().getId(), dpReq.getSurveyId());
+                if (dpList == null) {
+                    res.setCode(String.valueOf(HttpServletResponse.SC_BAD_REQUEST));
```
muloem: Perhaps we should return a `404` instead? @valllllll2000 how would the device behave when we return one of these error codes?

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
iperdomo: We're storing `[0]` for _All data points_ as in the JDO/Dao writes/reads a List of ids (longs). Instead of storing a potential large List of Ids, we have a _constant_ that represent all _current_ and _future_ datapoints for a Survey.

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
iperdomo: _"Duplicated code is better than a wrong abstraction"_ , we only have 3?

dlebrero: _""Duplication is the root of all evil"_. Even with 1, it is going to make the tests more readable.

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
dlebrero: I feel that my question was not understood. Rephrasing it:

Did you discard the option of adding an explicit flag/field in SurveyAssignment object to mean "ALL DATAPOINTS" or "PARTIAL DATAPOINTS"?

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
dlebrero: These are two different tests. Please split in two.

PR #3512, opened 2020-04-10
Issue/3496 datapoint counts

iperdomo:
* A new endpoint `/surveyed_locale_counts/{surveyGroupId}` returns the number of _Datapoints_ (`SurveyedLocale`) for that particular _Survey_ (`SurveyGroup`)
* We do a `KeysOnly` query to be as efficient as possible
  * https://cloud.google.com/appengine/docs/standard/java/datastore/queries#keys-only_queries
* We only use this count request in the Assignments tab, after selecting a Survey
* Translation strings need to be updated (We need to wait for `en.properties` to be in `master`)


iperdomo:
The UI changes look
Peek 2020-04-10 07-43
 like this

Issue #3498, opened 2020-04-09
Minimum conditions to be able to save assignment 


janagombitova:
# Context
To make users lives as easy as possible when working with Flow and thus mandatory actions as clear as possible. 

# Problem or idea
To have a working assignment the assignment needs a date, form and at least one device. Today we show these different pop up messages when these three conditions are not meant. This means every time Salim wants to save he has to read and add something new to his assignment. It is an extra hassle. 

![Screen Shot 2020-04-09 at 09.08.20.png](https://images.zenhubusercontent.com/5c8f6c3ea7f08c1a3af97cc1/37e266a2-9bd8-42f0-9c32-b31f92bb3c60)

# Solution or next step

Instead of having the pop-up messages, how about we disable `save` until the date, form and one device are selected?

janagombitova:
works well on uat2

PR #3513, opened 2020-04-10
[#3498] - disable save assignment button until all required fields are set

marvinkome:
To have a working assignment the assignment needs a date, form and at least one device. Today we show these different pop-up messages when these three conditions are not meant. This means every time Salim wants to save he has to read and add something new to his assignment. It is an extra hassle.

#### The solution
Disable the Save button until all required fields are set

#### Reviewer Checklist
* [ ] Formatted the code


marvinkome:
@iperdomo Yes we're disabling it not hiding

Issue #1223, opened 2015-04-01
Navigating away from Assignment Screen always sends a /survey_assignments/ request


rumca:
When the following steps are performed:
- Visit `Devices > Assignments List`
- Click `Create New Assignment`
- Don't fill out any details
- Navigate away from screen (via any method)

A request is sent to /rest/survey_assignments regardless of whether Save is clicked or not (and thus there's no validation on the details). This results in a null pointer at:

```
java.lang.NullPointerException
    at java.util.ArrayList.<init>(ArrayList.java:164)
    at org.waterforpeople.mapping.app.web.rest.SurveyAssignmentRestService.generateDeviceJobQueueItems(SurveyAssignmentRestService.java:195)
    at org.waterforpeople.mapping.app.web.rest.SurveyAssignmentRestService.newSurveyAssignment(SurveyAssignmentRestService.java:161)
```

PR #3525, opened 2020-04-16
[#1223] Use a 'transaction' for new Assignment

iperdomo:
The new `Assignment` behavior is different than the rest of the
application, e.g.  When creating a new `Survey`, we create and commit
the record, we get a the new `keyId` from the server and then we ask
the user to "edit" it.

In the case of `Assignment` we create a new record in the client and
we don't commit it until the user clicks the "Save" button or the user
navigates away triggering an implicit `FLOW.store.commit()`.

Since we don't want to save partially filled `Assignment` if the user
navigates away, we want to discard that record.

Ember Data has a concept of "transaction" and we use it to deal with
this exception.  When creating a new `Assignment` we do it within a
transaction and we discard if the user navigates away.

>  Because newly created records are dirty from the time they are created,
>  and because dirty records can not be added to a transaction, you must
>  use the `createRecord()` method to assign new records to a transaction.
>  For example, instead of this:
>    var transaction = store.transaction();
>    var person = App.Person.createRecord({ name: "Steve" });
>    // won't work because person is dirty
>    transaction.add(person);
>  Call `createRecord()` on the transaction directly:
>    var transaction = store.transaction();
>    transaction.createRecord(App.Person, { name: "Steve" });

https://github.com/akvo/akvo-flow/blob/20200403-170810/Dashboard/app/js/vendor/ember-data-rev10.js#L516-L555

### DIFF
--- a/Dashboard/app/js/lib/router/router.js
+++ b/Dashboard/app/js/lib/router/router.js
@@ -20,6 +20,7 @@ FLOW.Router = Ember.Router.extend({
     FLOW.selectedControl.set('selectedCascadeResource', null);
     FLOW.selectedControl.set('cascadeImportNumLevels', null);
     FLOW.selectedControl.set('cascadeImportIncludeCodes', null);
+    FLOW.selectedControl.set('surveyAssignmentTransaction', null);
     FLOW.surveyControl.set('content', null);
     FLOW.questionControl.set('OPTIONcontent', null);
     FLOW.metaControl.set('since', null);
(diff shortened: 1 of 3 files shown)


iperdomo:
@tangrammer the PR message is a _copy_ of the commit message. Github pre-fills the PR message with the content of the commit message: https://github.com/akvo/akvo-flow/pull/3525/commits/6ea6bd1d68e68f48afc61c0b391e1aff5abc4448

tangrammer:
even better! :)

Issue #3503, opened 2020-04-09
Assignments side-panel UI improvements


janagombitova:
# Context
**Why do we add this issue?**
Small improvement related to our current Data point assignment MVP (#2796) with the goal to make the UI as simple and clean as possible.

# Problem or idea - n1
When navigating the assignments side panel it is confusing on which section is selected and what is not selected and thus clickable. 

![Assignments - side panel](https://user-images.githubusercontent.com/12456965/78883060-4403d000-7a59-11ea-9270-095dced8b489.gif)

It should be that when I am in the Forms section, the text Forms is not longer clickable but highlighted. See mock ups below.  
<img width="1101" alt="Screen Shot 2020-04-09 at 11 51 34" src="https://user-images.githubusercontent.com/12456965/78883216-7ad9e600-7a59-11ea-82a1-cf234453aa8b.png">

# Problem or idea - n2
The overall alignment, use or vertical space and font is a bit clunky in the side panel. It hard to distinguish what is a section title what is a device group, what is a device as the font sides are the same.

<img width="1097" alt="Screen Shot 2020-04-09 at 11 55 30" src="https://user-images.githubusercontent.com/12456965/78883715-284cf980-7a5a-11ea-9e82-16c44f3ecfa9.png"> 

# Solution or next step
* [ ] If Salim is in a section, make the section title NOT clickable, only keep the highlight of that section 
* [ ] Make the vertical space between the device groups smaller 
* [ ] Make the device group and device list font a tad smaller than the section titles, Form and Devices

marvinkome:
**A note:**
The "Devices" tab isn't clickable when they're already devices available

janagombitova:
Looking good on uat2

PR #3527, opened 2020-04-16
[#3503] - clean up assignment sidebar to look like mock-up

marvinkome:
image

#### The solution
When in devices/datapoints page
image

When in forms page:
image

#### Reviewer Checklist
* [ ] Formatted the code


iperdomo:
@janagombitova can you comment on this, from the user point of view?

The original mockups have UI look that the left side is somehow _part_ (the same color) of the right/main one.

Something like this 
2020-04-16_1115x498

marvinkome:
I tried to do this, but if I remembered @loicsans  said it should match with the current styling and variables of flow

iperdomo:
> I tried to do this, but if I remembered @loicsans said it should match with the current css styling and variables of FLOW

OK

Issue #3562, opened 2020-05-19
Use random url for webforms


tangrammer:
# Context
Relates #3145 

# Problems
- We don't want to expose flow form SurveyId in the webform shareable url
- Surveys have to be validated as Webforms (constraints)
- Surveys need another boolean field `webform` that will indicate that could be used as webform

# Solution or next step
- Persist in Survey table a new field with `webform`=true/false
- Maybe we need to run a consistency scritp so all surveys starts with `webform`=false
- Extend SurveyRestService to create/store/return the `webform` boolean value
- implement function to encrypt `surveyId` with `restPrivateKey` and returns `webformId` that should be used for url shareable webform xxx.akvoflow.org/webform/webformId

- Double-check web form constraints

janagombitova:
https://flowsupport.akvo.org/article/show/111363-webforms Webforms with their random URLs are up and running in production.

PR #3564, opened 2020-05-20
Use random url for webforms

tangrammer:
Relates #3562 

Besides setting up a new endpoint  in `SurveyRestService`  `GET, value = "/{id}/webform_id"` and the logic to ofuscate the SurveyId value using an OneTimePad encrypted url encoded algorithm, there are changes to take care of the new Survey db boolean column `webform` that depends on survey column types support (currently don't supported `CASCADE, GEOSHAPE, SIGNATURE, CADDISFLY`)

Changes once integrated with UI branch
Screenshot 2020-05-21 at 08 40 42

New boolean column `webform` in Survey Table

Screenshot 2020-05-21 at 08 56 03

#### Reviewer Checklist
* [x] Added an explanation about the work done
* [x] Connected the PR and the issue on Zenhub
* [x] Added a test plan to the issue
* [x] Updated the copyright header (when relevant)
* [x] Formatted the code
* [x] Added a documentation (if relevant)
* [x] Added some unit tests (if relevant)

### DIFF
--- a/GAE/src/com/gallatinsystems/common/util/MD5Util.java
+++ b/GAE/src/com/gallatinsystems/common/util/MD5Util.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2010-2018 Stichting Akvo (Akvo Foundation)
+ *  Copyright (C) 2010-2018,2020 Stichting Akvo (Akvo Foundation)
  *
  *  This file is part of Akvo FLOW.
  *
--- a/GAE/src/com/gallatinsystems/common/util/MD5Util.java
+++ b/GAE/src/com/gallatinsystems/common/util/MD5Util.java
@@ -21,7 +21,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import javax.xml.bind.DatatypeConverter;
+import org.apache.commons.codec.binary.Base64;
 
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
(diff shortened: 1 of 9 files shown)


tangrammer:
@muloem could you point me the standard flow approach to load a config value? So we can use it with the current encoding secret-key https://github.com/akvo/akvo-flow/blob/issue-3562/GAE/src/org/waterforpeople/mapping/app/web/rest/SurveyRestService.java#L160

tangrammer:
@muloem BTW: not sure if there are more places where we need to take care of webform column types constraints besides current changes on QuestionRestService.java

muloem:
@tangrammer the properties are loaded a static field in the `PropertyUtil` class. So you can simply get the property by `PropertyUtil.getProperty()`

tangrammer:
> @tangrammer the properties are loaded a static field in the `PropertyUtil` class. So you can simply get the property by `PropertyUtil.getProperty()`
then ... these should be the changes needed
https://github.com/akvo/akvo-flow/pull/3564/commits/746d3ca5f6019ff85a05c99c613ed6ff619f47a3

https://github.com/akvo/akvo-flow-server-config/pull/599

tangrammer:
> In the PR we have resolved the status of webform=true for creating, and updating questions. But this should also be checked for when deleting a question although maybe its better handled in the UI..? 
webform `true` value couldn't change on removing questions, only on adding or changing questions

tangrammer:
@muloem I didn't take care of repeatable groups or monitoring ones as client side is checking..... should backend be extended on these scenarios?

muloem:
> @muloem I didn't take care of repeatable groups or monitoring ones as client side is checking..... should backend be extended on these scenarios?

@tangrammer  yes please. also slipped by...

tangrammer:
ok!

PR #3567, opened 2020-05-21
Fix webform validation conditions and refactor logic application contexts

tangrammer:

- Remove Cascade question type as not allowed
- Validate survey (and update webform boolean column db) only on `surveys/{id}/webform_id` and [publishing survey logic](https://github.com/akvo/akvo-flow/blob/issue-3562/GAE/src/org/waterforpeople/mapping/app/web/SurveyAssemblyServlet.java#L197) (thus webform python backend only uses s3 survey published definitions)

#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/com/gallatinsystems/survey/domain/WebForm.java
+++ b/GAE/src/com/gallatinsystems/survey/domain/WebForm.java
@@ -31,11 +31,19 @@ public class WebForm {
         return unsupportedTypes;
     }
 
+    public static boolean validQuestionGroups(final Survey survey){
+        return survey.getQuestionGroupMap().values().stream().filter(i -> i.getRepeatable().booleanValue()).collect(Collectors.toList()).size() == 0;
+    }
+
+    public static boolean validSurveyGroup(final SurveyGroup surveyGroup){
+        return !surveyGroup.getMonitoringGroup().booleanValue();
+    }
+
     public static boolean validWebForm(final SurveyGroup surveyGroup, final Survey survey, final List<Question> questions){
-        boolean hasRepeatableGroups = survey.getQuestionGroupMap().values().stream().filter(i -> i.getRepeatable().booleanValue()).collect(Collectors.toList()).size() > 0;
-        if (hasRepeatableGroups) return false;
-        boolean monitoring = surveyGroup.getMonitoringGroup().booleanValue();
-        if (monitoring) return false;
+        boolean validQuestionGroups = validQuestionGroups(survey);
+        if (!validQuestionGroups) return false;
+        boolean validSurveyGroup = validSurveyGroup(surveyGroup);
+        if (!validSurveyGroup) return false;
         List<Question> validQuestions = questions.stream().filter(i -> !unsupportedQuestionTypes().contains(i.getType().toString())).collect(Collectors.toList());
         return validQuestions.size() == questions.size();
     }
--- a/GAE/src/com/gallatinsystems/survey/domain/WebForm.java
+++ b/GAE/src/com/gallatinsystems/survey/domain/WebForm.java
@@ -23,7 +23,7 @@ import java.util.stream.Collectors;
 
 public class WebForm {
 
-    public static Set<String> unsupportedQuestionTypes(){
+    public static Set<String> unsupportedQuestionTypes() {
         Set<String> unsupportedTypes = new HashSet<String>();
         unsupportedTypes.add(Question.Type.GEOSHAPE.toString());
         unsupportedTypes.add(Question.Type.SIGNATURE.toString());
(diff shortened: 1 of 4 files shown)


GAE/src/com/gallatinsystems/survey/domain/WebForm.java
```diff
@@ -25,17 +25,26 @@
...
         unsupportedTypes.add(Question.Type.SIGNATURE.toString());
         unsupportedTypes.add(Question.Type.CADDISFLY.toString());
         return unsupportedTypes;
     }
 
-    public static boolean validWebForm(final List<Question> questions){
-        
-        List<Question> validQuestions = questions.stream().filter(i -> !unsupportedQuestionTypes().contains(i.getType().toString())).collect(Collectors.toList());
+    public static boolean validQuestionGroups(final Survey survey){
+        return survey.getQuestionGroupMap().values().stream().filter(i -> i.getRepeatable().booleanValue()).collect(Collectors.toList()).size() == 0;
```
iperdomo, line 35: Why do we need an explicit `boolean` primitive? .... using `i.getRepeatable()` does the trick

GAE/src/com/gallatinsystems/survey/domain/WebForm.java
```diff
@@ -25,17 +25,26 @@
...
+    }
 
+    public static boolean validSurveyGroup(final SurveyGroup surveyGroup){
+        return !surveyGroup.getMonitoringGroup().booleanValue();
```
iperdomo, line 39: Idem, that extra `.booleanValue()` is not required, I think

GAE/src/com/gallatinsystems/survey/domain/WebForm.java
```diff
@@ -25,17 +25,26 @@
...
     }
 
-    public static boolean validWebForm(final List<Question> questions){
-        
-        List<Question> validQuestions = questions.stream().filter(i -> !unsupportedQuestionTypes().contains(i.getType().toString())).collect(Collectors.toList());
+    public static boolean validQuestionGroups(final Survey survey){
+        return survey.getQuestionGroupMap().values().stream().filter(i -> i.getRepeatable()).collect(Collectors.toList()).size() == 0;
+    }
 
+    public static Boolean validSurveyGroup(final SurveyGroup surveyGroup){
```
muloem, line 38: So this one is a bit more subtle. It should be that webforms doesn't support monitoring forms. I.e. in a survey you have registration form and monitoring forms.  We should allow a registration form to be filled in.  Registration form id is === `SurveyGroup.newLocaleSurveyId` @marvinkome I guess this is already done on UI side?

marvinkome: Yeah on the UI, monitoring surveys aren't supported

muloem: Monitoring surveys or monitoring _forms_?

marvinkome: 👀 monitoring survey groups

muloem: Should be the same... in a monitoring Survey(Group) you have two kinds of forms, registration and monitoring forms. Registration forms can be shared via webforms, monitoring forms cannot be shared. Registration form id is found by looking at the `SurveyGroup.newLocaleSurveyId` property

GAE/src/com/gallatinsystems/survey/domain/WebForm.java
```diff
@@ -25,17 +25,26 @@
...
+        return !surveyGroup.getMonitoringGroup();
+    }
+
+    public static boolean validWebForm(final SurveyGroup surveyGroup, final Survey survey, final List<Question> questions){
+        boolean validQuestionGroups = validQuestionGroups(survey);
+        if (!validQuestionGroups) return false;
```
iperdomo, line 44: Let's always use `{ }` in `if` conditions.

GAE/src/com/gallatinsystems/survey/domain/WebForm.java
```diff
@@ -23,19 +23,32 @@
...
     }
 
-    public static boolean validWebForm(final List<Question> questions){
-        
-        List<Question> validQuestions = questions.stream().filter(i -> !unsupportedQuestionTypes().contains(i.getType().toString())).collect(Collectors.toList());
+    public static boolean validQuestionGroups(final Survey survey) {
+        return survey.getQuestionGroupMap().values().stream().filter(i -> i.getRepeatable()).collect(Collectors.toList()).size() == 0;
+    }
 
+    public static boolean validSurveyGroup(final Survey survey, final SurveyGroup surveyGroup) {
```
muloem, line 38:
```suggestion
    public static boolean validForm(final Survey survey, final SurveyGroup surveyGroup) {
```

Issue #3571, opened 2020-05-22
Software Delivery metrics


dlebrero:
As per other projects, we want to record the deploy frequency, change fail rate and cycle time

PR #3574, opened 2020-05-22
Issue/3571 delivery metrics

dlebrero:
Scripts to capture the software delivery metrics.

1. Promote script will ask if the release is a hot fix and give instructions for the developer to run if the developer wants to proceed with the deploy.
1. Flip script will ask about the release to switch the live traffic to, guessing which one is the most likely.

Changes:

1. "Promote" tags will deploy to dark production.
1. "Flip" tags will flip the production traffic.
1. UAT1 deploys do not rebuild the binaries, but reuse the ones in UAT2.

All scripts can be tested as we are using dev2 as uat2, dev1 as uat1 and dev3 as akvoflow-23.

TODO:
1. Move flip and promote script to the docker images repo
2. Use the correct names for the different environments (dev1-> uat1, dev2->uat2, ...)
3. Agree on what is the best version to show in the Dashboard and which one in Clojars. 
4. Agree on skip logic inside the scripts.

Part of #3571


ci/flip-production-traffic.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -u
+
+function log {
+   echo "$(date +"%T") - INFO - $*"
+}
+
+function read_version () {
+    CLUSTER=$1
```
iperdomo, line 10: Why `CLUSTER` ? In App Engine you work with _projects_

ci/flip-production-traffic.sh
```diff
@@ -0,0 +1,61 @@
...
+    log "Reading ${CLUSTER} version"
+    VERSION=$(gcloud app versions list --project="${CLUSTER}" --hide-no-traffic --service=default | grep "default" | tr -s " " | cut -f 2  -d\ )
```
iperdomo, line 12: I've seen those pipe _tricks_ elsewhere. If you want to continue working with _text_:
```bash
gcloud app versions list --project=akvoflow-uat1 --hide-no-traffic --service=default | awk '/^default/ {print $2}'
```
Or `gcloud` has `--format=json` for structured output
```bash
gcloud app versions list --project=akvoflow-uat1 --hide-no-traffic --service=default --format=json | jq -M -r '.[].version.id'
```

ci/promote-test-to-prod.sh
```diff
@@ -0,0 +1,85 @@
...
+log "Commits to be deployed:"
+echo ""
+
+git log --oneline "$NEWEST_VERSION_IN_PROD".."$TEST_LIVE_VERSION" | grep -v "Merge pull request" | grep -v "Merge branch"
+
+TAG_NAME="promote-$(TZ=UTC date +"%Y%m%d-%H%M%S")"
+
+echo ""
+read -r -e -p "Does this deployment contain a hotfix, rollback or fix-forward for a previous deployment? [Y/n] " FIX
+if [ "${FIX}" != "n" ] || [ "${FIX}" != "N" ]; then
```
iperdomo, line 72:
```
In promote-test-to-prod.sh line 72:
if [ "${FIX}" != "n" ] || [ "${FIX}" != "N" ]; then
                       ^-- SC2252: You probably wanted && here, otherwise it's always true.

For more information:
  https://www.shellcheck.net/wiki/SC2252 -- You probably wanted && here, othe...
```

ci/generate-slack-notification.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+OLDER_GIT_VERSION=$1
+NEWEST_GIT_VERSION=$2
+MSG=$3
+COLOR=$4
+WRAP_SLACK=$5
+GITHUB_PROJECT=$6
+
+cat << EOF > notify.team.sh
```
iperdomo, line 10: Why do we need a script to generate another script!? :exploding_head:

ci/promote-test-to-prod.sh
```diff
@@ -0,0 +1,85 @@
...
+
+function log {
+   echo "$(date +"%T") - INFO - $*"
+}
+
+GITHUB_PROJECT=akvo-flow
+NOTIFICATION=${4:-slack}
+
+function read_version () {
+    CLUSTER=$1
```
iperdomo, line 13: Ideam. Projects no cluster.

ci/deploy.sh
```diff
@@ -2,36 +2,15 @@
...
-
-if [[ "${CI_TAG:0:8}" == "promote-" ]]; then
-    echo "Skipping deployment"
-    exit 0
-fi
-
-develop_project_id="${DEVELOP_PROJECT_ID:=akvoflow-uat2}"
-release_project_id="${RELEASE_PROJECT_ID:=akvoflow-uat1}"
+develop_project_id="${DEVELOP_PROJECT_ID:=akvoflow-dev2}"
+FLOW_VERSION=${FLOW_VERSION:-${SEMAPHORE_GIT_SHA}}
```
iperdomo, line 10: Let's use `CI_GIT_SHA` or similar

dlebrero:
> * Agree on what is the best version to show in the Dashboard and which one in Clojars.
>   
>   * The `flip` and `promote` prefix are just operations on a _main_ version. We should strip out those _internal_ details? 🤔

I am guessing here that your suggestion is to use the date part of the tags as the version. Note that tags do not rebuild artefacts, so I will go for both versions to be $date-$git_sha, which I think is in line with other conversations about versioning that we have had in the past.

ci/flip-production-traffic.sh
dlebrero: Copy and paste legacy ...

ci/flip-production-traffic.sh
dlebrero: opted for --format="value(VERSION.id)"

ci/promote-test-to-prod.sh
dlebrero: I copy and pasted this error 3 times at least

ci/generate-slack-notification.sh
dlebrero: This is because as a confirmation for the promote or flip, we expect the developer to manually run some commands.

dlebrero:
I decided to not duplicate the skip logic inside the scripts.

Issue #3579, opened 2020-05-25
Add usage tracking for webform on flow forms


marvinkome:
# Context
Why do we add this issue?
To track and understand how users use the webform in Flow

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?
Add tracking to when a user clicks on the copy URL button

janagombitova:
Webform URL copied logged 11 times in the last 3 days. Tracking works well. Thanks Marvin

PR #3580, opened 2020-05-25
[#3579] - add usage tracking to webforms

marvinkome:
Relates #3579 

#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/Dashboard/app/js/lib/components/forms/form-share/index.jsx
+++ b/Dashboard/app/js/lib/components/forms/form-share/index.jsx
@@ -2,6 +2,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Modal from 'akvo-flow/components/reusable/Modal';
+import { trackEvent } from 'akvo-flow/analytics';
 import './style.scss';
 
 export default class WebFormShare extends React.Component {
@@ -18,6 +19,7 @@ export default class WebFormShare extends React.Component {
   copyToClipboard = () => {
     navigator.clipboard.writeText(this.props.data.shareUrl);
     this.setState({ copyToClipboard: true });
+    trackEvent('Webform URL copied');
   };
 
   openModal = () => {


Dashboard/app/js/lib/components/forms/form-share/index.jsx
```diff
@@ -18,6 +19,7 @@ export default class WebFormShare extends React.Component {
   copyToClipboard = () => {
     navigator.clipboard.writeText(this.props.data.shareUrl);
     this.setState({ copyToClipboard: true });
+    trackEvent('Webform URL copied');
```
tangrammer, line 22: don't we want to send any meaningful data inside this event?

marvinkome: I can't think of anything else that can/should be tracked, do you have any ideas?

tangrammer: @janagombitova  maybe could know better... I'm thinking on formId, or instanceId 🤔

marvinkome: I asked her, she said just this text is good enough

janagombitova: @tangrammer and @marvinkome would **instanceId** mean the organisation - so icco.akvoflow.org? That would be interesting to see, which partner uses the webforms the most

marvinkome: @janagombitova would that mean the current url?

janagombitova: I thought about this over the night and decided it is good enough just to have the count. We do not need more for now and should keep things to the minimum. Once we want to know more, we can add the properties later. So no need for extra data

Issue #3591, opened 2020-05-28
Data script to delete iterations and update the iteration numbers in RQG


muloem:
# Context
Some RQG have loads of iterations that it is not possible to delete/update them through the report import.  We need a temporary solution to clean up the RQG

# Solution or next step
We create a datascript to delete iterations

PR #3592, opened 2020-05-28
[#3591] Datascript to delete specified iterations for a survey instance

muloem:
#### The solution
* Delete one or more iterations by specifying at the command line
* Update the iteration index in the responses if needed (not strictly necessary)
* Update the dependency libraries

#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/scripts/data/delete-repeat-iterations.sh
+++ b/scripts/data/delete-repeat-iterations.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+# USAGE: ./delete-repeat-iterations.sh akvoflow-uat1 [surveyInstanceId] [group name] [1,2,4] [d | u]
+# d to delete the iterations
+# u to update the itertion number on remaining iterations
+APP_ID=$1
+SERVICE_ACCOUNT="sa-$APP_ID@$APP_ID.iam.gserviceaccount.com"
+REPOS_HOME="$(cd $(dirname "$THIS_SCRIPT")/../../.. && pwd)"
+P12_FILE_PATH="$REPOS_HOME/akvo-flow-server-config/$1/$1.p12"
+
+java -cp bin:"lib/*" \
+     org.akvo.gae.remoteapi.RemoteAPI \
+     DeleteRepeatQuestionGroupIteration \
+     $APP_ID \
+     $SERVICE_ACCOUNT \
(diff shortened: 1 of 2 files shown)


scripts/data/delete-repeat-iterations.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+# USAGE: ./delete-repeat-iterations.sh akvoflow-uat1 [surveyInstanceId] [group name] [1,2,4] [d | u]
```
valllllll2000, line 3: So in order to delete and update, you need to run the script twice? you cannot have both d and u?

scripts/data/src/org/akvo/gae/remoteapi/DeleteRepeatQuestionGroupIteration.java
```diff
@@ -0,0 +1,315 @@
...
+    private void parseArgs(String[] args, DatastoreService ds) {
+        Long surveyInstanceId = parseSurveyInstanceId(args);
+        if (surveyInstanceId.equals(-1L)) {
+            System.out.println("No survey instance id found. Exiting");
+            System.exit(1);
+        } else {
+            this.surveyInstanceId = surveyInstanceId;
+        }
+
+        this.formId = retrieveFormId(surveyInstanceId, ds);
```
valllllll2000, line 94: what happens is formId is null?

scripts/data/delete-repeat-iterations.sh
muloem: yes. thats the idea. I figured that if anything happens and there is a crash midway, then it will cause inconsistencies and delete unwanted data. e.g. if we delete iteration 3 and as we are updating the remaining ones i.e. 4 and 5 then something goes wrong, if we run the script again, we will then delete the updated ones from 4 and 5.

scripts/data/src/org/akvo/gae/remoteapi/DeleteRepeatQuestionGroupIteration.java
muloem: if we have only one question group matching then the data is deleted correctly but if for example we have two forms with the same question group, then we get an exception and cannot perform any updates.

Issue #3593, opened 2020-06-03
Add tooltip for personal data flag


janagombitova:
# Context
See more here #3576 

# Problem or idea
The more options Salim has to decide on when designing the form the more effort he needs to take. We want to keep things simple and one way to do that is to explain what different settings mean or do. For this, we use tooltips in the survey form editor. 

We recently added a new setting `Holds personal data`. The idea is to add a tooltip to explain what this setting does. 

# Solution or next step
Proposed text: Define that answers to this question are personal data (personally identifiable information - any data that can be used to identify a specific individual). When exporting the data via the Flow API, you can filter out the values based on this flag.

janagombitova:
Plus we should resolve the bug tooltips have today: When you click on one tooltip it works, but if you come back to it (without refreshing the page) then it does not.

janagombitova:
Released today

PR #3602, opened 2020-06-12
[#3593] Fix tooltip text

muloem:

* The `eventManager` is a shared instance therefore the tooltip text was being overwritten. i.e. all tooltips contained text of the last tooltip loaded on a page. 

#### The solution

* Save tooltip text in a new attribute and retrieve it on `mouseover` a tooltip element
* Replace handlebars comment with HTML comment to fix wrong rendering

#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/Dashboard/app/js/templates/navSurveys/nav-surveys-main.handlebars
+++ b/Dashboard/app/js/templates/navSurveys/nav-surveys-main.handlebars
@@ -50,7 +50,8 @@
                 <li><a class="addSurvey noChanges">{{t _create_new_survey}}</a></li>
               {{else}}
                 {{#if view.disableAddSurveyButtonInRoot}}
-                <!--  {{t _survey_only_in_folder}} -->
+                <!--  Needed in the comment in order for translation parser to identify translation key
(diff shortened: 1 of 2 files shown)


Dashboard/app/js/templates/navSurveys/nav-surveys-main.handlebars
```diff
@@ -50,7 +50,7 @@
                 <li><a class="addSurvey noChanges">{{t _create_new_survey}}</a></li>
               {{else}}
                 {{#if view.disableAddSurveyButtonInRoot}}
-                {{!--  {{t _survey_only_in_folder}} --}}
+                <!--  {{t _survey_only_in_folder}} -->
```
iperdomo, line 53: If not in use, let's remove it

muloem: Its in use. but because I pass it in as a key below, the translation parser doesnt recognize it. I'll add comment to clarify that.

Issue #3608, opened 2020-06-23
Cannot download many data points to device 


janagombitova:
For all the details check: https://github.com/akvo/akvo-flow-mobile/issues/1670

janagombitova:
Resolved and ready for release

PR #3611, opened 2020-07-01
[#3608] Download datapoints

muloem:
* Frequent timeouts while downloading datapoints from the device
* Not all datapoints downloaded

#### The solution
* Reduce the number of datapoints downloaded in a single batch to 30
* Use the `cursor` query parameter to enable selective/incremental downloading

#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/org/akvo/flow/api/app/DataPointServlet.java
+++ b/GAE/src/org/akvo/flow/api/app/DataPointServlet.java
@@ -18,6 +18,7 @@ package org.akvo.flow.api.app;
 
 import com.gallatinsystems.device.dao.DeviceDAO;
 import com.gallatinsystems.device.domain.Device;
+import com.gallatinsystems.framework.dao.BaseDAO;
 import com.gallatinsystems.framework.rest.AbstractRestApiServlet;
 import com.gallatinsystems.framework.rest.RestRequest;
 import com.gallatinsystems.framework.rest.RestResponse;
@@ -77,7 +78,7 @@ public class DataPointServlet extends AbstractRestApiServlet {
     @Override
     protected RestResponse handleRequest(RestRequest req) throws Exception {
         DataPointRequest dpReq = (DataPointRequest) req;
-        RestResponse res = new RestResponse();
+        DataPointResponse res = new DataPointResponse();
         if (dpReq.getSurveyId() == null) {
             res.setCode(String.valueOf(HttpServletResponse.SC_FORBIDDEN));
             res.setMessage("Invalid Survey");
@@ -99,10 +100,12 @@ public class DataPointServlet extends AbstractRestApiServlet {
 
         if (dpList.isEmpty()) {
             res.setCode(String.valueOf(HttpServletResponse.SC_NOT_FOUND));
-            res.setMessage("No assignment was found");
+            res.setMessage("No datapoint assignment was found");
             return res;
         }
         res = convertToResponse(dpList, dpReq.getSurveyId());
+        res.setCursor(BaseDAO.getCursor(dpList));
+
         return res;
 
     }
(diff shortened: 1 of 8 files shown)


GAE/src/org/akvo/flow/api/app/DataPointServlet.java
```diff
@@ -77,63 +80,89 @@ protected RestRequest convertRequest() throws Exception {
...
-                    return res;
-                }
-                res = convertToResponse(dpList, dpReq.getSurveyId());
-                return res;
-            }
-            res.setCode(String.valueOf(HttpServletResponse.SC_NOT_FOUND));
-            res.setMessage("Unknown device");
-        } else {
+        if (dpReq.getSurveyId() == null) {
             res.setCode(String.valueOf(HttpServletResponse.SC_FORBIDDEN));
```
iperdomo, line 84: We're changing the status code from `HTTP 404` to `HTTP 403`. If we don't find the *device* or the *datapoint list* is empty, we keep using `HTTP 404`. Why this case is different?

GAE/test/org/akvo/flow/api/app/DataPointServletTest.java
```diff
@@ -219,6 +220,44 @@ public void noAssignmentTest() {
...
-        assertNull(foundDataPoints);
+        assertTrue(foundDataPoints.isEmpty());
+    }
+
+    @Test
+    public void testRetrieveDataPointsByLastUpdateTime() throws InterruptedException {
+        final Long surveyId = randomId();
+        final List<SurveyedLocale> oldDataPoints = createDataPoints(surveyId, 5);
+        final Date lastUpdateTime = new Date();
+        Thread.sleep(1500); // only datapoints created after this point should be retrieved
```
iperdomo, line 231: I don't think we should filter by `lastUpdatedDateTime`, but if we need to. We can create a few datapoints "in the future" directly instead of dealing with `Thread.sleep` ? :thinking:

valllllll2000:
> I think we can't generate the `nextSyncUrl` directly, because our internal URLs have authentication baked-in (via the `ts` and `h` parameters), but I like that the server is in _control_ and can update/change the semantic of the parameter to fetch the next batch without disturbing the clients. Clients must consider the URL _opaque_ and just use it.
> 
> My feeling that is the Mobile App (a client) shouldn't need to know anything about `lastUpdateTime`. It should request the datapoints, we return X number of entities plus with a `baseURL` to get the next batch. When the server has no more datapoints (we reached the end of the index) we return `HTTP 204`.
> 
I am not against it, we just need to take backward compatibility into account

GAE/src/org/akvo/flow/api/app/DataPointServlet.java
muloem: The `403` was always there in the `else` branch so I have just made it more explicit. I am not sure how the older versions process the returned status so I did not want to change it. Otherwise I actually wanted to return a `401` if there is no `surveyId`

muloem:
@iperdomo So we went the route of the `lastUpdateTime` because thats how its done with the old API but the next block of work is to add a cursor parameter that is sort of the equivalent of what you are talking about.  We return the results to the app with a cursor and when the endpoint is called with that cursor, then we fetch the next batch.

iperdomo:
> @iperdomo So we went the route of the lastUpdateTime because thats how its done with the old API but the next block of work is to add a cursor parameter that is sort of the equivalent of what you are talking about. We return the results to the app with a cursor and when the endpoint is called with that cursor, then we fetch the next batch.

I'm OK with sorting `lastUpdateTime` but it should be a internal detail, not exposed to the Mobile App. It shouldn't be a parameter.

muloem:
> I'm OK with sorting lastUpdateTime but it should be a internal detail, not exposed to the Mobile App. It shouldn't be a parameter.
Sure. makes sense. I'll confirm first that the cursor is one that we can rely on.

PR #3647, opened 2020-10-07
[#3646] Use java.util.Base64 instead of org.apache.commons.codec

### LINKED ISSUE #3646: Use java.util.Base64 instead of org.apache.commons.codec.binary.base64
# Context
Java 8 already includes support for encoding

# Problem or idea
What is the cause?

# Solution or next step
How could we solve it?

tangrammer:
Fix #3646

### DIFF
--- a/GAE/src/com/gallatinsystems/common/util/S3Util.java
+++ b/GAE/src/com/gallatinsystems/common/util/S3Util.java
@@ -40,8 +40,8 @@ import java.util.TimeZone;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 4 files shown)


iperdomo:
Looks good, it's the *padding* behaviour the same? - https://en.wikipedia.org/wiki/Base64#Decoding_Base64_with_padding

iperdomo:
> Looks good, it's the padding behaviour the same?

I ask because we're touching a couple of classes used in Flow services for spreadsheet export/import

iperdomo (approved):

Issue #3645, opened 2020-09-28
Limit the number of monitoring submissions download with data points 


janagombitova:
# Context
Currently, when data points are assigned for monitoring and downloaded to the Flow app, previous submissions for the assigned monitoring forms and the data are downloaded with the data points. This functionality was created to give the enumerator background information on the data point and context to what data was previously submitted. It is also used for the feature to `prefill data` that add previously submitted answers to a new monitoring form submission. 

# Problem
However, with monitoring scaling, more and more data points are being assigned and all the previous submissions are being downloaded causing timeout issues. The more data users collect the more this issue will grow.

It also brings up the question, does Al need to have all the old data on his device? Does he need all of it? For what reason? Is it secure? Does it not cause biased entries? 

# Opportunity 
We need to ensure Flow can work on a large scale what however may mean we need to limit what data users get on their device, especially if it is not used/needed. 

## Research 
We asked a few users the following questions: 
* Does the enumerator need to have all previous submissions? 
* What do they use this data for (if they use it)? 

**User A**
>  Actually, I was not aware of this. So good that you mention it. Well, maybe it is even an advantage if the enumerator does not know the previous submission values, as this makes it more sure that the new submission is not influenced by previous submissions. So, I would limit it indeed to the last one. Then at least no trends can be seen and the enumerator is fully unbiased.

**User B**
> I think there are different use cases. For example, if it's quarterly monitoring, there might still be a need to see what happened in q1. But then for other partners, they might not need it at all. So it's a bit tricky and case-specific. And usually cases I have seen it is to see a value from the past, for example on agricultural products, or on types of crops they were cultivating at y1. Or just to see how if a point has been monitored already. But the data isn't been used. 

**User C**
> It could be good to have the previous submissions for reference, but I haven't particularly found a real case as of now. But can think of something like checking the functionality of water point for the enumerator reference and probe specific questions based on this information. But then the survey might need to actually be set up differently and have questions in the monitoring form depend on previous answers. 

## Idea 
As of now, it seems the historical data is not actively used. 

We could decide to limit the number of submissions downloaded per assigned form with the data points. For example, we, by default, only download the last and the first and indicate in the app in History that there is a total of XYZ forms submitted. Or only the last submissions. Or none. Later on, we could provide the option for Salim when assigning data points to decide if for this form all previous submissions need to be sent too. Or other options...

Kiarii:
My 2 cents and thought for food:
[...]

PR #3648, opened 2020-10-07
[#3645] Create new endpoints for the app to retrieve datapoints in smaller chunks

muloem:
* The backend would Time out when there are lots of form instances associated with a specific data point

#### The solution
* Split the retrieval of datapoints into smaller chunks. One endpoint returns only the datapoints and a second one returns all form instances (paginated) for a single datapoint.

#### Reviewer Checklist
* [ ] Formatted the code


GAE/src/org/waterforpeople/mapping/app/web/SurveyInstanceServlet.java
```diff
@@ -85,14 +90,21 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
...
+            try {
+                List<SurveyInstance> formInstances = formInstanceUtil.getFormInstances(siReq.getAndroidId(), siReq.getDataPointIdentifier(), LIMIT_FORM_INSTANCES_30, siReq.getCursor());
+                response.setSurveyInstances(formInstanceUtil.getFormInstancesDtoList(formInstances));
+                response.setCursor(BaseDAO.getCursor(formInstances));
+                response.setResultCount(formInstances.size());
+
+                return response;
+            } catch (Exception e) {
+                log.warning("Exception accessing endpoint: " + e.getMessage());
+                response.setCode(String.valueOf(HttpServletResponse.SC_NOT_FOUND));
```
tangrammer, line 105: not sure if every exception is a not found exception 🤔 ... maybe we need a couple of specific exceptions to know which kind was thrown (eg: DeviceNotFoundException) from the formInstanceUtil

GAE/src/org/waterforpeople/mapping/app/web/SurveyInstanceServlet.java
```diff
@@ -18,14 +18,17 @@
...
 
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 
 import com.gallatinsystems.framework.dao.BaseDAO;
 import org.akvo.flow.api.app.FormInstanceResponse;
 import org.akvo.flow.api.app.FormInstanceUtil;
 import org.akvo.flow.util.FlowJsonObjectWriter;
 import org.springframework.beans.BeanUtils;
+import org.springframework.http.HttpStatus;
```
iperdomo, line 31: We're not using this class. Let's remove the import.

iperdomo: Change made and pushed

GAE/src/org/waterforpeople/mapping/app/web/SurveyInstanceServlet.java
iperdomo: then you enter in the *longish conversation* of checked vs unchecked exceptions. https://www.baeldung.com/java-checked-unchecked-exceptions

iperdomo (approved):

PR #3651, opened 2020-10-12
Issue/3645 limit form instances

### LINKED ISSUE #3645: Limit the number of monitoring submissions download with data points

iperdomo:
Closes #3645

### DIFF
--- a/GAE/src/org/akvo/flow/api/app/DataPointUtil.java
+++ b/GAE/src/org/akvo/flow/api/app/DataPointUtil.java
@@ -17,18 +17,16 @@
 package org.akvo.flow.api.app;
 
 import java.util.*;
-import java.util.logging.ConsoleHandler;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import com.gallatinsystems.device.dao.DeviceDAO;
 import com.gallatinsystems.device.domain.Device;
 import com.gallatinsystems.surveyal.dao.SurveyedLocaleDao;
 import org.akvo.flow.dao.DataPointAssignmentDao;
-import org.akvo.flow.dao.SurveyAssignmentDao;
 import org.akvo.flow.domain.DataUtils;
 import org.akvo.flow.domain.persistent.DataPointAssignment;
-import org.akvo.flow.domain.persistent.SurveyAssignment;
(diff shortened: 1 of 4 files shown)


kardan (approved):
Approved after walkthrough by @iperdomo.

Issue #3652, opened 2020-10-13
Crash when trying to get assigned datapoints


valllllll2000:
# Context
Support request from Ecosia, they are unable to download datapoints even after latest release.
In the logs there is a NPE:
`com.gallatinsystems.framework.rest.AbstractRestApiServlet executeRequest: Could not execute rest request (AbstractRestApiServlet.java:119)
java.lang.NullPointerException
	at java.util.stream.Collectors.lambda$groupingBy$46(Collectors.java:953)
	at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1383)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at org.akvo.flow.api.app.DataPointUtil.getSurveyInstances(DataPointUtil.java:218)
	at org.akvo.flow.api.app.DataPointUtil.getSurveyedLocaleDtosList(DataPointUtil.java:67)
	at org.akvo.flow.api.app.DataPointServlet.convertToResponse(DataPointServlet.java:113)
	at org.akvo.flow.api.app.DataPointServlet.handleRequest(DataPointServlet.java:89)`

It seems that a datapoint has no registration form submission.

janagombitova:
Can we close this now that the changes are released?

PR #3653, opened 2020-10-13
[#3652] attempt to fix NPE

valllllll2000:
### Before the PR (what is the issue or what needed to be done)
NA
#### The solution
Fixed NPE when missing registration form submission for a given datapoint
NA
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/org/akvo/flow/api/app/DataPointUtil.java
+++ b/GAE/src/org/akvo/flow/api/app/DataPointUtil.java
@@ -16,14 +16,12 @@
 
 package org.akvo.flow.api.app;
 
-import java.util.*;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import java.util.stream.Collectors;
-
 import com.gallatinsystems.device.dao.DeviceDAO;
 import com.gallatinsystems.device.domain.Device;
+import com.gallatinsystems.survey.dao.QuestionDao;
+import com.gallatinsystems.survey.domain.Question;
 import com.gallatinsystems.surveyal.dao.SurveyedLocaleDao;
+import com.gallatinsystems.surveyal.domain.SurveyedLocale;
 import org.akvo.flow.dao.DataPointAssignmentDao;
 import org.akvo.flow.domain.DataUtils;
 import org.akvo.flow.domain.persistent.DataPointAssignment;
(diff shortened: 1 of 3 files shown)


GAE/src/org/akvo/flow/api/app/DataPointUtil.java
```diff
@@ -35,10 +33,18 @@
 import org.waterforpeople.mapping.domain.SurveyInstance;
 import org.waterforpeople.mapping.serialization.response.MediaResponse;
 
-import com.gallatinsystems.survey.dao.QuestionDao;
-import com.gallatinsystems.survey.domain.Question;
-import com.gallatinsystems.surveyal.domain.SurveyedLocale;
 import javax.annotation.Nullable;
+import java.util.ArrayList;
```
iperdomo, line 37: There are lots of changes to import *style* and *order* that are not directly to the change business logic change we want.

GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
```diff
@@ -67,6 +44,26 @@
 import com.google.appengine.api.taskqueue.Queue;
 import com.google.appengine.api.taskqueue.QueueFactory;
 import com.google.appengine.api.taskqueue.TaskOptions;
+import org.akvo.flow.domain.DataUtils;
```
iperdomo, line 47: IDEM. Import re-ordering, not directly related to business logic changes.

GAE/src/org/waterforpeople/mapping/dao/SurveyInstanceDAO.java
```diff
@@ -774,13 +771,14 @@ public SurveyInstance getRegistrationSurveyInstance(SurveyedLocale locale,
     public List<SurveyInstance> getMonitoringData(@Nonnull List<SurveyedLocale> surveyedLocales, int numberOfInstances) {
         List<SurveyInstance> result = new ArrayList<>();
 
-        surveyedLocales.forEach(surveyedLocale ->
-                result.add(getRegistrationSurveyInstance(surveyedLocale, surveyedLocale.getCreationSurveyId())));
-
-        surveyedLocales.forEach(surveyedLocale ->
+        for (SurveyedLocale s: surveyedLocales) {
```
iperdomo, line 774: We're changing the logic and introduce a *corner case*. We should add a test in [SurveyInstanceDAOTest](https://github.com/akvo/akvo-flow/blob/757155a918a44db85ec5f64a844cc6527decfec5/GAE/test/com/gallatinsystems/surveyal/dao/SurveyInstanceDAOTest.java)

iperdomo (changes requested):
Let's add a test for this case. Where we have *data points* for a given *survey* without *Registration Form Instance*

GAE/src/org/akvo/flow/api/app/DataPointUtil.java
valllllll2000: the IDE does this, we need a common code style for all IDEs so this does not happen

iperdomo: it does it with you reformat the whole file. We *had* a common style defined for Eclipse but not for IntelliJ IDEA

PR #3658, opened 2020-10-16
[#3652] using offset to return assigned datapoints

valllllll2000:
#### The solution
using offset to return assigned datapoints in batches of 30 as query by key does not allow to query with a cursor
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/org/akvo/flow/api/app/DataPointServlet.java
+++ b/GAE/src/org/akvo/flow/api/app/DataPointServlet.java
@@ -20,10 +20,7 @@ import com.gallatinsystems.framework.dao.BaseDAO;
 import com.gallatinsystems.framework.rest.AbstractRestApiServlet;
 import com.gallatinsystems.framework.rest.RestRequest;
 import com.gallatinsystems.framework.rest.RestResponse;
-import com.gallatinsystems.surveyal.dao.SurveyedLocaleDao;
 import com.gallatinsystems.surveyal.domain.SurveyedLocale;
-import org.akvo.flow.dao.DataPointAssignmentDao;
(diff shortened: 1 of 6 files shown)


GAE/src/org/akvo/flow/api/app/DataPointServlet.java
```diff
@@ -87,7 +79,11 @@ protected RestResponse handleRequest(RestRequest req) throws Exception {
         }
 
         res = convertToResponse(dpList, dpReq.getSurveyId());
-        res.setCursor(BaseDAO.getCursor(dpList));
+        String cursor = BaseDAO.getCursor(dpList);
+        if (cursor == null && dpList.size() > 0) {
+            cursor = dpList.size() + "";
+        }
+        res.setCursor(cursor);
```
muloem, line 86: If we follow the current behaviour of the app, it will send requests to the backend until the returned results == 0.

So take the case of 34 assigned datapoints. 
- First query = 30 datapoints and cursor=30
- second query = 4 datapoints and cursor=(4 or 34?)
- third query = 0 datapoints and cursor=(?)

valllllll2000: cursor would be null, I need to check if it could be an issue

valllllll2000: I checked and having a null cursor is not an issue. I think in this case it makes sense. User downloads all datapoints assigned to him and cursor is reset to null. Datapoints are assigned/modified. User pressed download -> download restarts from 0 (we have no way to tell if the datapoints have been modified recently because we don't have a "real" cursor)

muloem (approved):

Issue #3624, opened 2020-08-05
Inform the user that he/she is offline 


janagombitova:
# Context
Our tools are used in areas where stable internet connection is still not standard. But we do not want that to be a limitation and want to ensure users regardless of their location can use the products and trust them. 

## Why do we have this issue? What are we trying to solve? 
Today our products do not communicate to the user that they are offline, resulting in: 
* user losing their work 
* user thinking the products are slow (what is not true but he/she has no way to know)

# Opportunity 

## Current status quo

Flow lets you continue with the work until you hit save and Flow says something is being saved, but that is not happening and as a user you think that Flow is slow, where the problem is that you are offline. In Lumen when you create a new item at the moment you go offline the loading icon spins in screen indicating something is happening, what is as confusing as in Flow. Or Lumen informs you that it cannot save, please retry, but you cannot as you are offline. 

## The idea 
With this issue we want to start small: **How might we inform Salim that he is offline?** 

We will not change how the functionality works, disable editing, do autosaving, etc. That is for later. With this issue we just want to make sure he/she knows he is offline. 

As both Flow and Lumen do not inform users, it might be good to think of a way to solve this problem for both products. What do you think @marvinkome and @tangrammer? 

Here are some screenshots on how other products do it for inspiration

### Gmail:
<img width="1142" alt="Screen Shot 2020-08-05 at 12 49 44" src="https://user-images.githubusercontent.com/12456965/89405662-3a932780-d71c-11ea-96af-0d6ca00d42a0.png">

### Google Drive:
<img width="1131" alt="Screen Shot 2020-08-05 at 12 50 01" src="https://user-images.githubusercontent.com/12456965/89405665-3b2bbe00-d71c-11ea-9456-977e2e1da82e.png">

### Google sheets:
<img width="1130" alt="Screen Shot 2020-08-05 at 12 50 07" src="https://user-images.githubusercontent.com/12456965/89405667-3bc45480-d71c-11ea-93a4-dc463fe4bc87.png">

### Zenhub:
<img width="1143" alt="Screen Shot 2020-08-05 at 12 50 12" src="https://user-images.githubusercontent.com/12456965/89405668-3bc45480-d71c-11ea-8465-32dd584b891c.png">

### Slack:
<img width="1140" alt="Screen Shot 2020-08-05 at 12 49 12" src="https://user-images.githubusercontent.com/12456965/89405651-36670a00-d71c-11ea-8e44-46b1305c6c67.png">

[...]

PR #3625, opened 2020-08-06
Add /ping public entrypoint

tangrammer:
```
 curl -i 'http://localhost:8888/ping'
HTTP/1.1 200 OK
Content-Type: text/plain;charset=utf-8
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: 0
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Server: Development/1.0
Date: Thu, 06 Aug 2020 17:53:55 GMT
Content-Length: 5

ping
```

#### Reviewer Checklist
* [ ] Added an explanation about the work done
* [ ] Connected the PR and the issue on Zenhub
* [ ] Added a test plan to the issue
* [ ] Updated the copyright header (when relevant)
* [ ] Formatted the code
* [ ] Added a documentation (if relevant)
* [ ] Added some unit tests (if relevant)

PR #3668, opened 2020-11-17
[#3624] - add a basic text to tell user if he/she is offline

marvinkome:
Tell user is he/she is online or offline

#### The solution
This is a simple implementation of using navigator.onLine.
It wouldn't work on some older browsers, but we would handle this edge case in future modifications

image

#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/Dashboard/app/js/lib/components/reusable/OfflineIndicator/index.jsx
+++ b/Dashboard/app/js/lib/components/reusable/OfflineIndicator/index.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+
+export default class OfflineIndicator extends React.Component {
+  state = {
+    isOffline: false,
+  };
+
+  componentDidMount() {
+    window.addEventListener('online', this.onlineStatus);
+    window.addEventListener('offline', this.onlineStatus);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('online', this.onlineStatus);
+    window.removeEventListener('offline', this.onlineStatus);
+  }
+
+  onlineStatus = () => {
+    this.setState({ isOffline: !navigator.onLine });
+  };
+
+  render() {
+    return this.state.isOffline && <p>You are offline. Changes {"won't"} be saved</p>;
+  }
+}
--- a/Dashboard/app/js/lib/components/reusable/OfflineIndicator/index.jsx
+++ b/Dashboard/app/js/lib/components/reusable/OfflineIndicator/index.jsx
@@ -6,18 +6,24 @@ export default class OfflineIndicator extends React.Component {
   };
 
   componentDidMount() {
-    window.addEventListener('online', this.onlineStatus);
-    window.addEventListener('offline', this.onlineStatus);
+    this.timer = setInterval(() => {
+      fetch('https://www.google.com/generate_204', { mode: 'no-cors' })
+        .then(res => {
+          // with no-cors, res.status is 0 because the javascript ignores all response.
+          // we don't need the response, all we need to know is if the request went through or not
+          if (res.status !== 204) {
+            this.setState({ isOffline: false });
+          }
+        })
+        .catch(() => {
+          this.setState({ isOffline: true });
+        });
+    }, 5000);
   }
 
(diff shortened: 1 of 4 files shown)


iperdomo (changes requested):
The `onlineStatus` is a *new* API in the browsers. If you're connected to the local office WiFi you'll be *online* but potentially you can't reach the Flow Dashboard. See the comments on the other PR - https://github.com/akvo/akvo-flow/pull/3625#issuecomment-693384279

Some apps reach to an known service (e.g. Google) to know if you're actually online.

> Unfortunately, these events aren't fully reliable. If you need greater reliability, or if the API isn't implemented in the browser, you can use other signals to detect if you are offline including using service workers and responses from XMLHttpRequest.

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events

> "online" does not always mean connection to the internet, it can also just mean connection to some network. 

https://caniuse.com/online-status

marvinkome:
@iperdomo based on this. Do I remove the current navigator.onLine implementation and replace it with a timeout that pings google every X seconds?

iperdomo:
> @iperdomo based on this. Do I remove the current navigator.onLine implementation and replace it with a timeout that pings google every X seconds?

Yes, I think is a *safer* approach

iperdomo (changes requested):
Still waiting for the changes

Dashboard/app/js/lib/components/reusable/OfflineIndicator/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+  componentDidMount() {
+    this.timer = setInterval(() => {
+      fetch('https://www.google.com/generate_204', { mode: 'no-cors' })
+        .then(res => {
+          // with no-cors, res.status is 0 because the javascript ignores all response.
+          // we don't need the response, all we need to know is if the request went through or not
+          if (res.status !== 204) {
+            this.setState({ isOffline: false });
+          }
```
marvinkome, line 16: https://stackoverflow.com/questions/58518098/fetch-request-to-httpswww-google-com-fails-obviously-with-cors#comment103363447_58518361

This explains this line

Dashboard/app/js/lib/components/reusable/OfflineIndicator/index.jsx
```diff
@@ -6,19 +6,25 @@ export default class OfflineIndicator extends React.Component {
...
 
   componentDidMount() {
-    window.addEventListener('online', this.onlineStatus);
-    window.addEventListener('offline', this.onlineStatus);
+    this.timer = setInterval(() => {
+      fetch('https://www.google.com/generate_204', { mode: 'no-cors' })
+        .then(res => {
+          // with no-cors, res.status is 0 because the javascript ignores all response.
+          // we don't need the response, all we need to know is if the request went through or not
+          if (res.status !== 204) {
```
iperdomo, line 14: We want to check if the status is 204? `res.status === 204` right?

iperdomo: OK, i tested it, then if we know that `res.status` is `0`, we should check for that? `res.status === 0` ?

iperdomo: So either check for `res.status === 0` or remove the whole `if` block, since we don't want to check the status

iperdomo (approved):

Dashboard/app/js/lib/components/reusable/OfflineIndicator/index.jsx
```diff
@@ -0,0 +1,27 @@
...
+        });
+    }, 5000);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.timer);
+  }
+
+  render() {
+    return this.state.isOffline && <p>You are offline. Changes {"won't"} be saved</p>;
```
iperdomo, line 25: @marvinkome what about translations of this messages?

Issue #3669, opened 2020-11-17
Survey copy may duplicate/triplicate questions and groups


valllllll2000:
# Context
When making a survey copy, Jana has ended up with a Survey where some groups were triplicated (but not all questions)

# Problem or idea
We use tasks to copy questions and question groups. Tasks have to be completed in 10 minutes or they get interrupted. We have retry logic for tasks of 10 times or until successful completion. When a task is interrupted, work is resumed from the beginning.

# Solution or next step
* For long surveys we need to expect that a Task can be interrupted and make sure it completes in less than 10 minutes, we could use smaller tasks: ej one Task per group or per question?
* If a Task restarts we should check which groups/questions already exist and not to copy them.

We have decided to optimize the Translations copy. Problem: some old translations in the datastore do not have the surveyId setup so part 2 of the work requires to create a script to fix this using the parentId field.

janagombitova:
What is the status of this issue? Is there still more to be done? How have we improved the copying?

iperdomo:
We deployed a change that will prevent reaching the 10min limit when copying.

iperdomo:
Just for completion, a Survey with 1 form and 538 questions (without any translation) takes 6min

PR #3673, opened 2020-11-20
[#3669] add a script to fix the translations ids

valllllll2000:
Some translations are missing surveyId and surveyGroupId and with the new logic #3669  they would not be copied
#### The solution
Create a script to fix the data
NA
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/scripts/data/src/org/akvo/gae/remoteapi/FixOldTranslations.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/FixOldTranslations.java
@@ -2,6 +2,7 @@ package org.akvo.gae.remoteapi;
 
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.PreparedQuery;
@@ -70,10 +71,12 @@ public class FixOldTranslations implements Process {
         System.out.println("Fixing QuestionOption translation");
         Long parentId = (Long) t.getProperty("parentId");
         if (parentId != null) {
-            final Query.Filter fs = new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY,
-                    Query.FilterOperator.EQUAL, KeyFactory.createKey("QuestionOption", parentId));
-            final Query q = new Query("QuestionOption").setFilter(fs);
-            Entity option = ds.prepare(q).asSingleEntity();
+            Entity option = null;
+            try {
+                option = ds.get(KeyFactory.createKey("QuestionOption", parentId));
+            } catch (EntityNotFoundException e) {
+                e.printStackTrace();
+            }
             if (option != null) {
                 final Entity question = fetchQuestion(ds, (Long) option.getProperty("questionId"));
                 if (question != null) {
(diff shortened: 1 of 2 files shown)


scripts/data/src/org/akvo/gae/remoteapi/FixOldTranslations.java
```diff
@@ -0,0 +1,129 @@
...
+        translations.addAll(entities);
+        System.out.println("Found " + translations.size() + " broken translations");
+        return translations;
+    }
+
+    private Entity fetchGroup(DatastoreService ds, Long groupId) {
+        if (groupId == null) {
+            return null;
+        }
+        final Query.Filter fs = new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY,
```
iperdomo, line 114: If you know the `id` of the entity you want, you can just do `.get()`, you don't need to *query* for it:
You need to convert the `Long` value into ` Key`:
https://cloud.google.com/appengine/docs/standard/java/datastore/entities#Retrieving_an_entity

You can get a `Key` passing the `Long` value and the `Kind` name:
https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/KeyFactory.html#createkey_2

scripts/data/src/org/akvo/gae/remoteapi/FixOldTranslations.java
```diff
@@ -0,0 +1,129 @@
...
+            return t;
+        }
+        return null;
+    }
+
+    private List<Entity> fetchBrokenTranslations(DatastoreService ds) {
+        List<Entity> translations = new ArrayList<>();
+        final Query q = new Query("Translation");
+        final PreparedQuery pq = ds.prepare(q);
+        final Query.Filter filterNullSurveyId = new Query.FilterPredicate("surveyId", Query.FilterOperator.EQUAL, null);
```
iperdomo, line 101: I wonder if this query is *"correct"*. In the Datastore a *missing* property gets exposed as a `null` value, when you have already the object in memory, but you can't query for entities with a missing property. See: https://cloud.google.com/appengine/docs/standard/java/datastore/query-restrictions#entities_lacking_a_property_named_in_the_query_are_ignored

If we're sure that the value **is** `null` then just ignore this comment.

scripts/data/src/org/akvo/gae/remoteapi/FixOldTranslations.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+        return null;
+    }
+
+    private Entity fixQuestionOptionTranslation(Entity t, final DatastoreService ds) {
+        System.out.println("Fixing QuestionOption translation");
+        Long parentId = (Long) t.getProperty("parentId");
+        if (parentId != null) {
+            final Query.Filter fs = new Query.FilterPredicate(Entity.KEY_RESERVED_PROPERTY,
```
iperdomo, line 73: Use `ds.get()` if we know the `id` of the entity we want

iperdomo (changes requested):
I've added a couple of comments:
* Use `ds.get(<Key>)` for entities we know the `id`
* We can't search for entities with missing properties, so querying for `null` can be misleading

scripts/data/src/org/akvo/gae/remoteapi/FixOldTranslations.java
valllllll2000: yes I am sure this works, I have tested it and Translations do get found with these missing properties and loads of them

valllllll2000:
Fixed: getting entities by Key, As for null, I am sure this is the value per default as I have tried it on several dev instances and for sandbox for example it found 60 such entities which now have been fixed.

iperdomo (approved):

Issue #3685, opened 2020-11-25
Form xml does not contain QG translations


joycarpediem:
# Context
Why do we add this issue?
In webforms we will need to display the translations for question groups. The question groups contain longish texts which the user will need to understand to properly grasp the context of the questions

# Problem or idea
What is the cause?
We can enter question group translations in the form definition. However the form xml does not contain the translations 

# Solution or next step
How could we solve it?
Form xml should contain the translations for the QG

janagombitova:
When handling this issue we should look at all Form elements that should have translations but maybe do not: 
* form title
* question groups
* question tool tips 

And we need to also handle how this will affect the app

iperdomo:
@janagombitova @joycarpediem can you confirm that we only need the translation strings in the XML? I mean, we can deal with the *use* of those translations in the App later and not part of this effort?

janagombitova:
@iperdomo Joy needs to be able to show the translations in webforms, so we can leave the app changes for another time, but if it is minimal effort then I rather we handle then on both sides. I will leave this up to you and @valllllll2000 to estimate and decide.

joycarpediem:
> @janagombitova @joycarpediem can you confirm that we only need the translation strings in the XML? I mean, we can deal with the _use_ of those translations in the App later and not part of this effort?

@iperdomo  For the webforms we only need the translations to be present in  the xml

PR #3694, opened 2020-11-30
[#3685] added translations to question group

valllllll2000:
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/com/gallatinsystems/survey/dao/QuestionGroupDao.java
+++ b/GAE/src/com/gallatinsystems/survey/dao/QuestionGroupDao.java
@@ -86,8 +86,7 @@ public class QuestionGroupDao extends BaseDAO<QuestionGroup> {
             int i = 1;
             for (QuestionGroup group : groups) {
                 List<Translation> translations = translationDao.findTranslations(
-                        group.getKey().getId(), Translation.ParentType.QUESTION_GROUP_DESC,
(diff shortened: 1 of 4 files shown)


GAE/src/com/gallatinsystems/survey/dao/QuestionGroupDao.java
```diff
@@ -79,6 +85,16 @@ public void deleteGroupsForSurvey(Long surveyId) throws IllegalDeletionException
         if (groups != null) {
             int i = 1;
             for (QuestionGroup group : groups) {
+                List<Translation> translations = translationDao.findTranslations(
+                        group.getKey().getId(), Translation.ParentType.QUESTION_GROUP_DESC,
+                        Translation.ParentType.QUESTION_GROUP_NAME);
+                HashMap<String, Translation> translationMap = new HashMap<>();
+                if (translations != null) {
+                    for (Translation t: translations) {
+                        translationMap.put(t.getLanguageCode(), t);
```
muloem, line 94: Wont it happen here that if we have a list of translation for both the group description and the group name that one will be overwritten for the same language?

valllllll2000: you are right, after checking the Translations, it seem group description does not exist but I removed it from the query just in case

muloem (approved):

PR #3696, opened 2020-12-01
[#3685] Prevent empty <altText/> tag at Question level

iperdomo:


valllllll2000:
This will not work, the `@JsonInclude(JsonInclude.Include.NON_NULL)` inside question would prevent empty tags. The empty tag is inside XmlHelp

valllllll2000 (changes requested):
See previous comment

Issue #3689, opened 2020-11-26
Copying surveys does not copy the dependencies fully 


janagombitova:
When copying surveys some dependencies do not copy fully. 

Steps I took: 
1. I copied a survey with 1 form, translated, with 80 questions. 
2. I waited for the notification in the Messages tab, it came within a minute
3. I opened my copy, the questions were there, no duplications, translations too, options too but some dependencies were not set well. Some dependencies were copied over ok. 

When I say **some dependencies were not set well** I mean the checkbox dependent was checked but: 
* There was no question defined and thus no options 
* The question was defined but no option were shown/selected 

![Screen Shot 2020-11-26 at 11 28 17](https://user-images.githubusercontent.com/12456965/100347720-15396a80-2fe6-11eb-8955-c805e927850f.png)
![Screen Shot 2020-11-26 at 11 27 40](https://user-images.githubusercontent.com/12456965/100347721-17032e00-2fe6-11eb-8634-c3a572764f16.png)

PR #3701, opened 2020-12-03
[#3689] Fix different group dependencies are broken

valllllll2000:
#### The solution
We were using a map to store copied groups so the group order was lost,
using a list now
Added a test

Co-authored-by: Juan A. Ruz <juanantonioruz@gmail.com>
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/com/gallatinsystems/survey/dao/SurveyUtils.java
+++ b/GAE/src/com/gallatinsystems/survey/dao/SurveyUtils.java
@@ -113,12 +113,12 @@ public class SurveyUtils {
 
         log.log(Level.INFO, "Copying " + qgList.size() + " `QuestionGroup`");
 
-        List<Pair<Long, QuestionGroup>> sourceToCopiedGroupMap = shallowCopyQuestionGroups(copiedSurveyId, qgList);
+        List<Pair<Long, QuestionGroup>> listOfPairs = shallowCopyQuestionGroups(copiedSurveyId, qgList);
 
         // batch save question groups
-        new QuestionGroupDao().save(sourceToCopiedGroupMap.stream().map(pair -> pair.right).collect(Collectors.toList()));
+        new QuestionGroupDao().save(listOfPairs.stream().map(pair -> pair.right).collect(Collectors.toList()));
 
-        for(Pair<Long, QuestionGroup> entry : sourceToCopiedGroupMap) {
+        for(Pair<Long, QuestionGroup> entry : listOfPairs) {
             Long sourceGroupId = entry.left;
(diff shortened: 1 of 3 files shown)


GAE/src/com/gallatinsystems/survey/dao/SurveyUtils.java
```diff
@@ -16,11 +16,13 @@
 
 package com.gallatinsystems.survey.dao;
 
+import com.google.gwt.dev.util.Pair;
```
iperdomo, line 19: Why there is a need to use a `Pair` from gwt?

GAE/src/com/gallatinsystems/survey/dao/SurveyUtils.java
```diff
@@ -111,14 +113,14 @@ public static void copySurvey(Long copiedSurveyId, Long originalSurveyId) {
 
         log.log(Level.INFO, "Copying " + qgList.size() + " `QuestionGroup`");
 
-        Map<Long, QuestionGroup> sourceToCopiedGroupMap = shallowCopyQuestionGroups(copiedSurveyId, qgList);
+        List<Pair<Long, QuestionGroup>> sourceToCopiedGroupMap = shallowCopyQuestionGroups(copiedSurveyId, qgList);
 
         // batch save question groups
-        new QuestionGroupDao().save(sourceToCopiedGroupMap.values());
+        new QuestionGroupDao().save(sourceToCopiedGroupMap.stream().map(pair -> pair.right).collect(Collectors.toList()));
```
iperdomo, line 119: this complex one can be done with `new ArrayList(sourceToCopiedGroupMap.values())` ? not sure the value of stream to just get the values

GAE/src/com/gallatinsystems/survey/dao/SurveyUtils.java
valllllll2000: we were using javafx.util.Pair but that worked in my local environment but broke the build saying the class was not found, this was another option, I don't see what si the problem with this class, can you specify? I also though about creating my own object but why doing so if that class exists, unless there is a valid reason to not use it

GAE/src/com/gallatinsystems/survey/dao/SurveyUtils.java
valllllll2000: yes because this is no longer a map, I renamed the var to reflect that

valllllll2000:
as per @muloem suggestion will try using TreeMap

tangrammer:
finally choosing Pair instead TreeMap to avoid complexity

Issue #3711, opened 2020-12-13
Check Java properties encoding files as part of CI


kardan:
# Context
https://github.com/akvo/akvo-flow/issues/3686

# Problem or idea
It's easy to by mistake swap encoding to UTF-8 while adding translations (if you use a non Java IDE as Code). Nothing will break until it shows up in the UI in production. 

# Solution or next step
We should check in builds to make sure we have not corrupted translations.

PR #3712, opened 2020-12-13
[#3711] Check properties file encoding script

kardan:
This PR does not enable anything in CI. It's only the start on a script to check encodings.

### DIFF
--- a/ci/check-property-encoding.sh
+++ b/ci/check-property-encoding.sh
@@ -39,7 +39,7 @@ check() {
 
 echo "Checking for non ISO-8859-1 encoded .properties files:"
 check
-if [ $EXIT_CODE = 0 ];
+if [[ $EXIT_CODE = 0 ]]
 then
     echo "Properties files has the correct encoding"
 fi
--- a/ci/check-property-encoding.sh
+++ b/ci/check-property-encoding.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+#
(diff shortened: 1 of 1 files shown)


iperdomo (changes requested):
```
$ shellcheck ci/check-property-encoding.sh 

In ci/check-property-encoding.sh line 21:
DIRECTORY=$(cd `dirname $0` && pwd)
               ^----------^ SC2046: Quote this to prevent word splitting.
               ^----------^ SC2006: Use $(...) notation instead of legacy backticked `...`.
                        ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
DIRECTORY=$(cd $(dirname "$0") && pwd)


In ci/check-property-encoding.sh line 29:
            DETECTED_ENCODING=$(uchardet ${property_file})
                                         ^--------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
            DETECTED_ENCODING=$(uchardet "${property_file}")

For more information:
  https://www.shellcheck.net/wiki/SC2046 -- Quote this to prevent word splitt...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2006 -- Use $(...) notation instead of le...
```

I don't think we need to deal with `dirname $0` and those fancy detect directory. We can expect the script to be used from the root folder. So: `find GAE/ -name '*.properties`

iperdomo (approved):

Issue #3773, opened 2021-04-19
Error generating Data cleaning export for akvoflow-23


janagombitova:
Reported via support https://flowhelp.reamaze.com/admin/conversations/error-generating-slash-exporting-reports

* On akvoflow-23 the partner cannot generate the Data cleaning export for this dataset: Nepal > NPL_1059 Siraha Livelihood > Endline Survey NPL 1059 > Endline Assessment NPL 1059 > New form (4131210133)
* The survey is published and the structure looks ok - the only thing 'odd' is that the form is translated (what should not be an issue), but in the original text the 1st question also holds the translation after a `/` - but I also think this should not be an issue - but wanted to point it out 
* generating a Data cleaning export for other forms on the instance works fine

PR #3776, opened 2021-04-19
[#3773] prevent crash when question text is missing

valllllll2000:
Shouldn't happen but somehow there is a question with a null text field
#### The solution
Prevent crashes in the reports for null question titles
NA
#### Reviewer Checklist
* [ ] Formatted the code


kardan (approved):
LGTM

Issue #3597, opened 2020-06-11
Delete monitoring submissions with registration submission


janagombitova:
In Flow data deletion is relatively limited. You can delete one submission at a time, so you cannot delete all data to a survey form in one go. You have to go through all the individual submissions. In some cases partners contacted us with a list of surveys they wanted the data to be deleted. 

Sometimes partners want to remove old data from their instance or have done a test round of the survey form, want to delete all the submission so they can start the actual data collection. 

We never invested more in deletion because of the risk **what if someone deletes an entire dataset by mistake**. Some products handle such cases - where you cannot bring the thing you delete back again by asking the user for a double confirmation where you need to type something in. Plus stating this cannot be undone. 

Then we have the case of monitoring. What is someone tries to **delete registration form data but there are monitoring forms with data too**. Do we not allow to delete the registration form data until the monitoring forms' data are deleted? or do we ask if the user wants to delete everything in one go? This is a problem also when the user deletes just one registration form submission. In that case the monitoring form submission is still stored, thus the data point exists but without having the registration form data to define it. Perhaps we should follow the pattern we have in Lumen, that when deleting a dataset we show linked visuals and dashboards and delete them all.

Kiarii:
I do not see and assume there cant be any use cases for 'orphan' data, given a reg form provides a datapoint's 'parent' data and 'mon' form child data. 

Deleting parent data therefore should by this assumption include or bundle up the deletion of respective child data. The deletion of such data should be clearly communicated in the UI.

We can still maintain that no bulk deletion is done - so users can only delete datapoints (data) one at a time. Ideally if we provided bulk deletion, we might want to consider providing a 2 or 4 week grace period in which the deleted data sits in a deleted folder, and can be recovered, and auto-deletes completely after the period has elapsed.. 🤺

janagombitova:
@Kiarii indeed 'orphan' submissions - so monitoring submissions without a data point - are useless conceptually but also practically. They cause issues in analysis and in Lumen.

Ideally if a registration form submission is about to be deleted we also remove the linked monitoring form submissions and somehow show the user before doing so.

janagombitova:
Im going to rename this issue to handle the monitoring deletion and create a new one to handle bulk deletion so we keep things separate

Kiarii:
In this of a single delete we could use a modal dialog to notify the user (and ask them to confirm) that all monitoring data corresponding to a datapoint being deleted will also be deleted..

[...]

PR #3781, opened 2021-04-30
[#3597] added script to count orphaned instances

valllllll2000:
#### The solution
added script to count orphaned instances and add them to a csv file
#### Reviewer Checklist
* [ ] Formatted the code


scripts/data/src/org/akvo/gae/remoteapi/FixOrphanedSubmissions.java
```diff
@@ -0,0 +1,152 @@
...
+            System.out.println("No orphaned SurveyInstances found");
+        }
+    }
+
+    private Set<Long> fetchExistingDataPointInstances(DatastoreService ds, Map<Long, Set<Long>> mappedBySurveyId) {
+        Set<Long> dataPointIds = new HashSet<>();
+        Set<Entity> allDataPoints = new HashSet<>();
+        for (Long surveyId : mappedBySurveyId.keySet()) {
+            Query.Filter f = new Query.FilterPredicate("surveyId", Query.FilterOperator.EQUAL, surveyId);
+            Query.Filter f2 = new Query.FilterPredicate("surveyedLocaleId", Query.FilterOperator.IN, mappedBySurveyId.get(surveyId));
```
muloem, line 104: I think in this case we probably do not need to filter by the datapoint Ids that have been retrieved. We want all the form instances that are present for an existing registration form form, and then extract a set of datapoint ids from them. so that we can compare if indeed for every form instance that exists, there is also a datapoint.

scripts/data/src/org/akvo/gae/remoteapi/FixOrphanedSubmissions.java
```diff
@@ -0,0 +1,152 @@
...
+            System.out.printf("Found a total of %d orphaned instances\n", surveyInstances.size());
+            if (doIt) {
+                batchDelete(ds, surveyInstances);
+            }
+        } else {
+            System.out.println("No orphaned SurveyInstances found");
+        }
+    }
+
+    private Set<Long> fetchExistingDataPointInstances(DatastoreService ds, Map<Long, Set<Long>> mappedBySurveyId) {
```
muloem, line 99:
```suggestion
    private Set<Long> findDataPointsIdsWithExistingRegistrationForm(DatastoreService ds, Map<Long, Set<Long>> mappedBySurveyId) {
```

this is what I understood this method is trying to do

scripts/data/src/org/akvo/gae/remoteapi/FixOrphanedSubmissions.java
```diff
@@ -0,0 +1,152 @@
...
+                mappedBySurveyId.put(surveyId, datapoints);
+            } else {
+                mappedBySurveyId.get(surveyId).add(dataPointId);
+            }
+        }
+
+        Set<Long> dataPointsFound = fetchExistingDataPointInstances(ds, mappedBySurveyId);
+        dataPointIds.removeAll(dataPointsFound);
+
+        List<Entity> instances = getSurveyInstances(ds, new ArrayList<>(dataPointIds));
```
muloem, line 76:
```suggestion
        List<Entity> instances = getSurveyInstancesToDelete(ds, new ArrayList<>(dataPointIds));
```

scripts/data/src/org/akvo/gae/remoteapi/FixOrphanedSubmissions.java
```diff
@@ -0,0 +1,152 @@
...
+                sb.append(
+                        String.format("%s,%s,%s", instance.getProperty("surveyId"), instance.getProperty("surveyedLocaleIdentifier"), key.getId()))
+                        .append("\n");
+            }
+            FileUtils.write(f, sb.toString(), true);
+        }
+        if (surveyInstances.size() > 0) {
+            System.out.printf("Found a total of %d orphaned instances\n", surveyInstances.size());
+            if (doIt) {
+                batchDelete(ds, surveyInstances);
```
muloem, line 92: I guess you will also need to delete the datapoints after this right?

valllllll2000: added datapoint deletion

scripts/data/src/org/akvo/gae/remoteapi/FixOrphanedSubmissions.java
```diff
@@ -0,0 +1,156 @@
...
+                dataPointIds.add(surveyedLocaleId);
+            }
+        }
+        return dataPointIds;
+    }
+
+    private List<Entity> getSurveyInstancesToDelete(DatastoreService ds, List<Long> surveyedLocaleIds) {
+        if (!surveyedLocaleIds.isEmpty()) {
+            Query.Filter f1 = new Query.FilterPredicate("surveyedLocaleId", Query.FilterOperator.IN, surveyedLocaleIds);
+            Query q = new Query("SurveyInstance");
```
muloem, line 124: I think here you can also save a little by setting it to `setKeysOnly()` so you do not retrieve any other properties. I see you log some form instance properties but unless you think its absolutely necessary, I would suggest we remove them.

muloem (approved):

PR #3800, opened 2021-05-13
Issue/3799 fix dev deploy

valllllll2000:
Fix the dev-deploy script
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/dev-deploy.sh
+++ b/dev-deploy.sh
@@ -1,16 +1,48 @@
 #!/usr/bin/env bash
 
-set -eu
+# see README at root of the project
+# usage: build with `docker-compose up --build -d && docker-compose logs -f
+# make sure FLOW_GH_TOKEN is setup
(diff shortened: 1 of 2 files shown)


dev-deploy.sh
```diff
@@ -1,16 +1,48 @@
...
 
-java -cp /google-cloud-sdk/platform/google_appengine/google/appengine/tools/java/lib/appengine-tools-api.jar \
-     com.google.appengine.tools.admin.AppCfg \
-     update ./target/appengine-staging
+archive_name="${FLOW_VERSION}.zip"
+(
+    cd target
+    rm -rf appengine-staging/WEB-INF/appengine-web.xml
+    zip "${archive_name}" -q -r appengine-staging/*
+)
```
muloem, line 48: I recall this section was copying the files to the deployments bucket but in this case do you need it since you are copying directly to the instance?

valllllll2000: you are right, I will remove

muloem (approved):

Issue #3810, opened 2021-05-19
Geocells library


valllllll2000:
# Context
Geocells library is using a repository that has been discontinued

# Problem or idea
`[ERROR] Failed to execute goal on project akvo-flow: Could not resolve dependencies for project org.akvo.flow:akvo-flow:war:1: Failed to collect dependencies at com.beoui:geocell:jar:0.0.10.1: Failed to read artifact descriptor for com.beoui:geocell:jar:0.0.10.1: Could not transfer artifact com.beoui:geocell:pom:0.0.10.1 from/to bintray.com (https://dl.bintray.com/meetup/maven/): Authorization failed for https://dl.bintray.com/meetup/maven/com/beoui/geocell/0.0.10.1/geocell-0.0.10.1.pom 403 Forbidden -> [Help 1]`

# Solution or next step
Add the jar manually

PR #3811, opened 2021-05-19
[#3810] add geocells library manually

valllllll2000:
also fix the error with log4j by removing it

#### The solution
Update after @iperdomo commented:
* We can no longer use maven it's broken, that is what is causing the issue: see https://dl.bintray.com/meetup/maven/com/beoui/geocell/0.0.10.1/geocell-0.0.10.1.jar it says Forbidden
This is a very old library and it seems all the links of Google doc from that are broken, not sure why but maybe because they have now another tool for this https://cloud.google.com/blog/products/data-analytics/geospatial-insights-bigquery-gis-and-data-studio-choropleth
Here is another java Geocell library which explains quite well the Geocells: https://bitbucket.org/molson/geocell-java/src/master/

* After removing that library, I got the error about all the log4j classes not found. Even after adding the dependency strangely it did not work, I had to do some changes because the code was expecting an older version of log4j. In other parts of flow, we use java.util.logging.Logger so I decided to use it here too and avoid adding an extra dependency.
#### Reviewer Checklist
* [ ] Formatted the code


iperdomo:
* Why do we need to add *manually* if can be managed by Maven? 🤔 
https://mvnrepository.com/artifact/com.beoui/geocell/0.0.10.1

* Why there is lots of changes related to *logging* ?

my five cents

muloem (approved):

PR #3815, opened 2021-05-20
[#3597] disable ok button while number of items is still loading

valllllll2000:
The dialog appears but users can accept before knowing that monitoring instances will also be deleted
#### The solution
while loading:
Screenshot from 2021-05-21 13-42-44

when loading completed:
Screenshot from 2021-05-21 13-42-46

#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/Dashboard/app/js/templates/application/application.handlebars
+++ b/Dashboard/app/js/templates/application/application.handlebars
@@ -28,8 +28,7 @@
       <br/><br/>
       <div class="buttons menuCentre">
         <ul>
-          {{#if FLOW.dialogControl.showOK}} <li><a {{action "doOK" target="FLOW.dialogControl"}} class="ok smallBtn">{{t _ok}}</a></li>{{/if}}
-          {{#if FLOW.dialogControl.showOKDisabled}} <li><a class="ok smallBtn disabled">{{t _ok}}</a></li>{{/if}}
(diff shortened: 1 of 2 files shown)


Dashboard/app/js/templates/application/application.handlebars
```diff
@@ -29,6 +29,7 @@
       <div class="buttons menuCentre">
         <ul>
           {{#if FLOW.dialogControl.showOK}} <li><a {{action "doOK" target="FLOW.dialogControl"}} class="ok smallBtn">{{t _ok}}</a></li>{{/if}}
+          {{#if FLOW.dialogControl.showOKDisabled}} <li><a class="ok smallBtn disabled">{{t _ok}}</a></li>{{/if}}
```
muloem, line 32: I think in this case you may want to use the framework facility `{{bindAttr }}` that is for exactly this kind of use case.

see an example here 
https://github.com/akvo/akvo-flow/blob/529c2aee5f55c3d109b909c238639d0172f90392/Dashboard/app/js/templates/application/application.handlebars#L21

valllllll2000: ok Fixed, thanks a lot for your help on this one @muloem

muloem (approved):

Issue #3791, opened 2021-05-11
Batch upload images


valllllll2000:
# Context
One of our partners needs to batch upload many images at once. All the images will belong to the same forms and the rest of the data for that form would have already been imported via spreadsheets. The form contains up to 40 photo questions. The first question should be a photo of the farmer but the rest of the pictures can be randomly assigned to any photo question in order.

The survey is here: TREE PLANTING IMPACT / planting site registration and monitoring

TODO: find out if the images all go to one form or multiple forms

# Given
All the images will go to one form
One form can have up to 45 image questions most of them non mandatory
The first image should be the photo of the farmer
All the other images order does not matter: it does not matter to which form question they are assigned

Partner imports data via spreadsheet and then would batch upload all images
They will place each group of images in the corresponding folder which have as name the instance to which these images should be assigned. Then all these images will be zipped together and uploaded.

# Possible steps
Fetch formId from the surveyInstanceId
For each folder, get a list of image questions
Upload images to s3
Assign each image to its question and instance

Errors: if the surveyInstanceId does not match anything in our datastore we need to show the error to the user so they can retry

# Doubts
Where in the UI do we add that feature?
How can it be later made usable by others?

janagombitova:
I am not sure if it should be here or as a separate issue but we also need to ensure that the photo geotags are read and stored as if the photos came from the Flow app.

Kiarii:
Talked further with Emanuel, we thought the batch image upload API will be applicable for single image upload - which could later be added to the Inspect data tab - so users can edit / change images if that be a need. We were left with some open questions whose response might suggest whether to consider supporting the front end stuff for single image upload/edit: 

- Has any other partner besides Ecosia asked for an image edit support
- Inspite of all the media data collected via flow, has there been an overarching use or analysis of such data. From the talks with Ecosia so far, have they hinted why it is so important for them to have image data, and what sort of value analysis wise it provides for them?

As for technical constraints, Mulo will look into applicable folder size limits i.,e. if a datapoint/submission can have some 45 images and on a moderate estimate there is some 20 datapoints/submissions that makes 45*20 = 900 images. So would we have hiccups uploading such a moderately sized folder? Do we need to curtail what the user can upload at any one go, how? 

On the UI we need to consider UI how to communicate 3 states: 
- image upload (and unpacking process)
- upload (and unpacking failure); what points of failure do we consider
- How long do we expect the 900 images to take to upload and unpack - how do we communicate that
- image upload success

[...]

PR #3816, opened 2021-05-25
[#3791] one image upload

valllllll2000:
#### The solution
Upload one image to s3 and insert the answer to the store
Added tests
#### Reviewer Checklist
* [ ] Formatted the code


GAE/src/org/akvo/flow/rest/ImageUploadRestService.java
```diff
@@ -0,0 +1,219 @@
...
+            return null;
+        }
+    }
+
+    private void saveQuestionAnswer(Question question, SurveyInstance formInstance, String fileName) {
+        Long questionGroupId = question.getQuestionGroupId();
+        QuestionGroup group = new QuestionGroupDao().getByKey(questionGroupId);
+        boolean repeatable = group.getRepeatable();
+        if (repeatable) {
+            saveRepeatedQuestionAnswer(question, formInstance, fileName);
```
muloem, line 128: I think we should leave out the handling of repeat question group for now.  In this case, there are no photos in the repeated groups I believe. and its one that needs a little more designing on how to handle it.

GAE/src/org/akvo/flow/rest/ImageUploadRestService.java
```diff
@@ -0,0 +1,219 @@
...
+        createAndSaveQuestionAnswer(question, formInstance, fileName, questionAnswerStoreDao, iteration);
+    }
+
+    private void createAndSaveQuestionAnswer(Question question, SurveyInstance formInstance, String fileName, QuestionAnswerStoreDao questionAnswerStoreDao, int iteration) {
+        QuestionAnswerStore store = new QuestionAnswerStore();
+        store.setSurveyId(formInstance.getSurveyId());
+        store.setSurveyInstanceId(formInstance.getKey().getId());
+        store.setQuestionID(question.getKey().getId() + "");
+
+        //TODO: should the date be same as collection date on the form Instance??
```
muloem, line 165: If we are able to get the metadata of when the image was taken maybe thats better.

muloem (commented):
Looks good! 👍 

A couple of comments inline to consider and we also need to handle the extraction and saving of the geo data if its there in the image but we can maybe do that in a follow up PR

GAE/src/org/akvo/flow/rest/ImageUploadRestService.java
valllllll2000: I will add this together with geotag extraction in second PR

GAE/src/org/akvo/flow/rest/ImageUploadRestService.java
```diff
@@ -0,0 +1,183 @@
...
+@Controller
+@RequestMapping("/image_upload")
+public class ImageUploadRestService {
+
+    private static final Logger log = Logger
+            .getLogger(ImageUploadRestService.class.getName());
+
+    @RequestMapping(method = RequestMethod.POST, value = "/question/{questionId}/instance/{instanceId}")
+    @ResponseBody
+    public Response uploadImage(@PathVariable("questionId") String questionId,
```
muloem, line 49: I believe its possible to already convert to a `Long` here.. Same for the `formInstanceId`

muloem:
```suggestion
    public Response uploadImage(@PathVariable("questionId") Long questionId,
```

GAE/src/org/akvo/flow/rest/ImageUploadRestService.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        if (file == null || file.isEmpty()) {
+            return new Response(400, "File is not valid");
+        }
+        String fileExtension = getFileType(file);
+        if (fileExtension == null) {
+            return new Response(400, "File type is not valid: only jpg and png are accepted");
+        }
+        //TODO: shall we resize? does the calling task handle resize
```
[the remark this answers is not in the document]
valllllll2000: https://github.com/akvo/akvo-flow/issues/3817

GAE/src/org/akvo/flow/rest/ImageUploadRestService.java
valllllll2000: done

muloem (approved):

PR #3836, opened 2021-06-22
Issue/3835 publish backend

valllllll2000:
#### The solution
Rest service to publish a form without a Task
#### Reviewer Checklist
* [ ] Formatted the code

### DIFF
--- a/GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
+++ b/GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
@@ -36,6 +36,7 @@ public class FormAssemblyRestService {
     private static final Logger log = Logger.getLogger(FormAssemblyRestService.class.getName());
     private static final String SURVEY_UPLOAD_DIR = "surveyuploaddir";
     private static final String SURVEY_UPLOAD_URL = "surveyuploadurl";
+    public static final String BUCKET = "s3bucket";
 
     @PostMapping(consumes = "application/json")
     @ResponseBody
@@ -47,6 +48,8 @@ public class FormAssemblyRestService {
         long formId = form.getObjectId();
         String xmlAppId = props.getProperty("xmlAppId");
         String appStr = (xmlAppId != null && !xmlAppId.isEmpty()) ? xmlAppId : SystemProperty.applicationId.get();
+        //TODO: fetch translations
+        //TODO: fetch cascades
         XmlForm jacksonForm = new XmlForm(form, appStr, alias, survey.getCode());
         String formXML;
         try {
@@ -86,23 +89,19 @@ public class FormAssemblyRestService {
                 .stream().flatMap(Collection::stream).collect(Collectors.toList());
     }
 
-    /**  Upload a zipped form file twice to S3 under different filenames.
-     * @param fileName1
-     * @param fileName2
-     * @param formXML
-     * @return
+    /**
+     * Upload a zipped form file twice to S3 under different filenames.
      */
     public UploadStatusContainer uploadFormXML(String fileName1, String fileName2, String formXML) {
         Properties props = System.getProperties();
-        String bucketName = props.getProperty("s3bucket");
+        String bucketName = props.getProperty(BUCKET);
         String directory = props.getProperty(SURVEY_UPLOAD_DIR);
+        String uploadUrl = props.getProperty(SURVEY_UPLOAD_URL);
 
         UploadStatusContainer uc = new UploadStatusContainer();
         uc.setUploadedZip1(uploadZippedXml(formXML, bucketName, directory, fileName1));
         uc.setUploadedZip2(uploadZippedXml(formXML, bucketName, directory, fileName2));
-        uc.setUrl(props.getProperty(SURVEY_UPLOAD_URL)
-                + props.getProperty(SURVEY_UPLOAD_DIR)
-                + "/" + fileName1 + ".zip");
+        uc.setUrl(uploadUrl + directory + "/" + fileName1 + ".zip");
         return uc;
     }
 
--- a/GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
+++ b/GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
@@ -1,0 +1,123 @@
+package org.akvo.flow.rest;
+
+import com.gallatinsystems.common.domain.UploadStatusContainer;
+import com.gallatinsystems.common.util.S3Util;
+import com.gallatinsystems.common.util.ZipUtil;
+import com.gallatinsystems.survey.dao.SurveyDAO;
+import com.gallatinsystems.survey.dao.SurveyGroupDAO;
+import com.gallatinsystems.survey.dao.SurveyUtils;
+import com.gallatinsystems.survey.domain.Question;
+import com.gallatinsystems.survey.domain.QuestionGroup;
+import com.gallatinsystems.survey.domain.Survey;
+import com.gallatinsystems.survey.domain.SurveyGroup;
+import com.gallatinsystems.survey.domain.WebForm;
+import com.google.appengine.api.utils.SystemProperty;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import org.akvo.flow.xml.PublishedForm;
+import org.akvo.flow.xml.XmlForm;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@Controller
+@RequestMapping("/form_assembly")
+public class FormAssemblyRestService {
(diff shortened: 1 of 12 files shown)


GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
```diff
@@ -0,0 +1,440 @@
...
+                CascadeResource cascadeResource = cascadeResourceDao.getByKey(question.getCascadeResourceId());
+                if (cascadeResource != null) {
+                    question.setCascadeResource(cascadeResource.getResourceId());
+                    question.setLevelNames(cascadeResource.getLevelNames());
+                }
+            }
+        }
+    }
+
+    private void attachFormTranslations(Survey form) {
```
muloem, line 287: Since all the transations have a `parentId`, wouldnt it be simpler/more clear to create a map like `parentId -> (languageCode -> translationList)` and then process it from there, rather than having specific methods like `getFormTranslations()`, `getGroupTranslations()`?

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
```diff
@@ -0,0 +1,440 @@
...
+        String alias = props.getProperty("alias");
+        String xmlAppId = props.getProperty("xmlAppId");
+        String appStr = (xmlAppId != null && !xmlAppId.isEmpty()) ? xmlAppId : SystemProperty.applicationId.get();
+        XmlForm jacksonForm = new XmlForm(form, survey, appStr, alias);
+        try {
+            Long formId = form.getObjectId();
+            return new FormUploadXml(Long.toString(formId), //latest version in plain filename
+                    formId + "v" + form.getVersion(), //archive copy
+                    PublishedForm.generate(jacksonForm));
+        } catch (IOException e) {
```
muloem, line 116: I think rather than catching the exception, rather throw an exception at the method level and if an exception happens then the server will return a 500 to the client and we can deal with showing the user an error message there.

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
```diff
@@ -0,0 +1,440 @@
...
+    private HashMap<String, Translation> getGroupTranslationsMap(List<Translation> translations, long id) {
+        List<Translation> translationsForForm = getTranslationsForParent(translations, id);
+        return mapTranslations(translationsForForm);
+    }
+
+    private HashMap<String, Translation> mapTranslations(@Nullable List<Translation> translations) {
+        HashMap<String, Translation> mappedTranslations = new HashMap<>();
+        if (translations != null) {
+            for (Translation t : translations) {
+                mappedTranslations.put(t.getLanguageCode(), t);
```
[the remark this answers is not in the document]
valllllll2000: It is already like this in the current implementation. Not sure how but somehow it works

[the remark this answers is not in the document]
valllllll2000: https://github.com/akvo/akvo-flow/blob/master/GAE/src/com/gallatinsystems/survey/dao/SurveyDAO.java#L124

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
valllllll2000: Actually I think it would be a better idea to return a Response object like here https://github.com/akvo/akvo-flow/blob/master/GAE/src/org/akvo/flow/rest/ImageUploadRestService.java#L185, what do you think? I can put a 500 code.

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
valllllll2000: that's a really good idea, I have mapped them by parentId -> list of translations because the questions want list but all the others want a map

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
valllllll2000: After giving it more thought I realized this is not how we are supposed to return errors and the ImageUploadRestService is doing it wrong. I have changed to using ResponseStatusException and also returning the SurveyDto when success.

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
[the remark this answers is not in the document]
muloem: ok. I think it works because we currently dont have any translations for survey descriptions.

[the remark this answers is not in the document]
muloem: Are you using this mapping for any other entities that may have multiple translations that are in use?

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
```diff
@@ -0,0 +1,411 @@
...
+                    formUploadXml.getXmlContent());
+            if (uc.getUploadedZip1() && uc.getUploadedZip2()) {
+                formUploadSuccess(survey, form);
+                log.info("Completed form assembly for " + formId);
+                surveyDto.setStatus(Survey.Status.PUBLISHED.toString());
+                return surveyDto;
+            } else {
+                String message = "Failed to upload assembled form, id " + formId;
+                log.severe(message + "\n" + uc.getMessage());
+                throw new ResponseStatusException(INTERNAL_SERVER_ERROR, message, new Exception(uc.getMessage()));
```
muloem, line 100: Nice. thats even better! 👍

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
```diff
@@ -0,0 +1,411 @@
...
+                    int order = groupDto.getOrder() != null ? groupDto.getOrder() : i;
+                    groupMap.put(order, group);
+                }
+                i++;
+            }
+        }
+        return groupMap;
+    }
+
+    private QuestionGroup mapToGroup(QuestionGroupDto groupDto) {
```
muloem, line 185: Maybe you want to group this and other mapping methods in the respective mapper? i.e. `QuestionDtoMapper`, `QuestionGroupDtomapper`, `QuestionOptionDtoMapper`

GAE/test/org/akvo/flow/xml/FlowXmlObjectReaderTests.java
```diff
@@ -104,70 +104,70 @@
...
         assertEquals("1.0", testFormDto.getVersion());
-        assertNotEquals(null, testFormDto.getQuestionGroupList());
+        assertNotNull(testFormDto.getQuestionGroupList());
         assertEquals(1, testFormDto.getQuestionGroupList().size());
 
         QuestionGroupDto qg = testFormDto.getQuestionGroupList().get(0);
         assertEquals(1, qg.getOrder());
         assertEquals("The Empty Group", qg.getName());
-        assertEquals(null, qg.getQuestionMap());
+        assertNull(qg.getQuestionList());
```
muloem, line 158: This made me wonder, in the app, is it ok if you get a `questionList` intead of a `questionMap`?  Will everything still work as expected?

GAE/war/WEB-INF/rest-servlet.xml
```diff
@@ -35,6 +35,7 @@
     <bean name="/devices" class="org.waterforpeople.mapping.app.web.rest.DeviceRestService" lazy-init="true" />
     <bean name="/messages" class="org.akvo.flow.rest.MessageRestService" lazy-init="true" />
     <bean name="/image_upload" class="org.akvo.flow.rest.ImageUploadRestService" lazy-init="true" />
+    <bean name="/form_assembly" class="org.akvo.flow.rest.FormAssemblyRestService" lazy-init="true" />
```
muloem, line 38:
```suggestion
    <bean name="/form_publish" class="org.akvo.flow.rest.FormAssemblyRestService" lazy-init="true" />
```

GAE/test/org/akvo/flow/xml/FlowXmlObjectReaderTests.java
valllllll2000: The app gets the xml file, but in any case for json we also would need the list. The map is only used to have ordered questions as far as I can tell.

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
valllllll2000: My idea was to do that when adding the tests, extract those into different delegate classes to be able to easily test them.

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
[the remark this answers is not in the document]
valllllll2000: so: questions have list (due to the issue we had as there are too many types), question groups have: QUESTION_GROUP_DESC, QUESTION_GROUP_NAME but we use a map and it seems to work, maybe the desc one is not used? Question option has only one type QUESTION_OPTION and uses a map.

GAE/src/org/akvo/flow/rest/FormAssemblyRestService.java
muloem: > After giving it more thought I realized this is not how we are supposed to return errors and the ImageUploadRestService is doing it wrong. I have changed to using ResponseStatusException and also returning the SurveyDto when success.

I had also thought to make a mental note (better yet an issue) to also handle this properly in the `ImageUploadRestService`

muloem (approved):

PR #3879, opened 2021-09-27
Issue 3878/form list with count of form instances script

dedenbangkit:
We want to analyze the number of form instances including survey forms and their path.
Task assigned by @joycarpediem

#### The solution
Generate CSV file with below structure:

| Instance Name | Form ID | Form Name | Survey Name | Total Form Instances | Last Submission Date | Folder Path |
|------|---------|-------------|--------------|----------------------|-------------|-------------|
| uat1 | 123456 | Test Registration Form | Test Survey | 50 | 2021-12-31 | Emmanuel Test > Testing |

#### Usage
- Single Instance
  ```bash
  $ ./count-forminstances.sh [appengine-id]
  Processing [alias-name]
  alias-name form instance counts written
  filename: /tmp/form-instance-counts_alias-name.csv
  ```
- All Instances
  The `--all` argument will prompt you to type process code

  ```bash
  $ ./count-forminstances.sh --all
  Please type the process code 9dc7d95bae: 9dc7d95bae
  Processing [first-alias-name]
  [first-alias-name] form instance counts written
  filename: /tmp/form-instance-counts_[first-alias-name].csv
  ...
  Processing [last-alias-name]
  [last-alias-name] form instance counts written
  filename: /tmp/form-instance-counts_[last-alias-name].csv
  -------------------------------
  joining tmp files
  done: /tmp/all-form-instance-counts.csv
  ```
Screen Shot 2021-09-28 at 02 54 37

#### Reviewer Checklist
* [x] Formatted the code
* [x] Added a documentation (if relevant)

### DIFF
--- a/scripts/data/src/org/akvo/gae/remoteapi/CountFormInstances.java
+++ b/scripts/data/src/org/akvo/gae/remoteapi/CountFormInstances.java
@@ -25,24 +25,28 @@ import com.google.appengine.api.datastore.Query.CompositeFilterOperator;
 import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterOperator;
 import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Query.SortDirection;
 import com.google.appengine.repackaged.org.apache.commons.io.FileUtils;
 import java.io.File;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Date;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 /*
  */
-public class GetDatapointCounts implements Process {
+public class CountFormInstances implements Process {
 
   private Map<Long, String> sgName = new HashMap<>();
   private Map<Long, String> sgType = new HashMap<>();
   private Map<Long, Long> sgParents = new HashMap<>();
 
   private Map<Long, String> surveyNames = new HashMap<>();
+  private Map<Long, String> surveyDates = new HashMap<>();
   private Map<Long, Long> surveyParents = new HashMap<>();
   private Map<Long, Long> surveyCounts = new HashMap<>();
   private Map<Long, List<Long>> surveyToGroups = new HashMap<>();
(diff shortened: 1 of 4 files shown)


scripts/data/src/org/akvo/gae/remoteapi/CountFormInstances.java
```diff
@@ -0,0 +1,183 @@
...
+  private String iName = "";
+
+  @Override
+  public void execute(DatastoreService ds, String[] args) throws Exception {
+
+    if (args.length != 1) {
+      throw new IllegalArgumentException("Missing instance name");
+    }
+
+    iName = args[0];
```
muloem, line 63: Does this work? I kind of recall that you would need to pass the instance id twice at the commandline if you want it to appear here. Otherwise its stripped off in the `RemoteAPI.main()` method.

scripts/data/src/org/akvo/gae/remoteapi/CountFormInstances.java
```diff
@@ -0,0 +1,183 @@
...
+          String strDate = dateFormat.format(surveyDate);
+          surveyCounts.put(surveyId, count);
+          surveyNames.put(surveyId, surveyName);
+          surveyDates.put(surveyId, strDate);
+          surveyParents.put(surveyId, surveyGroup);
+          surveyToGroups.put(surveyId, new ArrayList<Long>());
+        }
+      }
+    }
+  }
```
muloem, line 182: My suggestion here would be things.

Second I would use the `setKeysOnly()` for the SurveyInstance query. Since you only want to count you do not need any other info from the Survey instance other than how many are there. If you do it the way it is now, it will increase the number of reads and that could increase our monthly bill for the instances considerably.  setKeysOnly does not count towards the reads/cost.

scripts/data/src/org/akvo/gae/remoteapi/CountFormInstances.java
dedenbangkit: yes, because this script is only work with my bash script, please check: https://github.com/akvo/akvo-flow/blob/0cf577c2f8be72dec6c3283dac45677e2b426e5d/scripts/data/count-forminstances.sh#L21-L22

Why do we need iName? because we need to have instance_name for the first column of CSV file

scripts/data/src/org/akvo/gae/remoteapi/CountFormInstances.java
dedenbangkit: Let's wait for @joycarpediem to answer.

dedenbangkit: @joycarpediem confirmed that for now, we don't need the last updated data. I'm going to comment on the line then merge @muloem

PR #3928, opened 2022-07-29
Code quality and readability review 

lutzseverino:
# Code quality and readability review
This PR contains certain quality improvements that don't tackle on changing the approach, but rather readability.

## Problem
Code clarity and readability could be improved upon on some parts of the code I'm going to work on.

## Solution
Automatically format with the IDE, cleanup the logic, and many other micro-improvements that make the code more readable and of better quality overall.

#### Use of ternary operators

This logic may be unnecessary or overcomplicated, but I'm not worrying about that right now.
```java
public String getValue() {
    if (value != null) {
        return value;
    }

    if (valueText != null) {
        return valueText.getValue();
    }

    return null;
}
```
Replacing that with a ternary one-liner improves readability and makes it easier for a future contributor to apply a different approach or verify this one.

```java
public String getValue() {
    return value == null ? (valueText == null ? null : valueText.getValue()) : value;
}
``` 
#### Switches

If-else chains to switches…
```java
if ("buildMap".equals(action)) {
    // empty
    } else if ("purgeExpiredSurveys".equals(action)) {
        purgeExpiredSurveys();
    } else if ("purgeOrphanJobQueueRecords".equals(action)) {
        purgeOrphanJobQueueRecords();
// ...
```
```java
switch (action) {
    case "buildMap":
        // empty
        break;
    case "purgeExpiredSurveys":
        purgeExpiredSurveys();
        break;
    case "purgeOrphanJobQueueRecords":
        purgeOrphanJobQueueRecords();
// ...
```

#### Line-breaks

And convenient line-breaks for a massive readability boost.
```java
private void extractImageFileGeotags() {
    Calendar deadline = Calendar.getInstance();
    deadline.add(Calendar.MONTH, ONE_MONTH_AGO);
    log.info("Starting scan for image answers, newer than: " + deadline.getTime());
    QuestionAnswerStoreDao qaDao = new QuestionAnswerStoreDao();
    String cursor = "";
    int json = 0;
    int nonjson = 0;
    Media media;
    // ...
```
```java
private void extractImageFileGeotags() {
    Date date = this.getModifiedDate(Calendar.MONTH, -1);

    log.info("Starting scan for image answers, newer than: " + date);

    int json = 0;
    int nonJson = 0;

    QuestionAnswerStoreDao qaDao = new QuestionAnswerStoreDao();
    String cursor = "";
    Media media;
    // ...
```
#### Extract repeated code
Additionally, some code repetition was present and removed. Not everywhere though, some instances were not worth to change due to possible changes in the future.
This logic was present several times, including the `#getTime` call. This can all be effectively abstracted.
```java
private void purgeOldMessages() {
    Calendar deadline = Calendar.getInstance();
    deadline.add(Calendar.YEAR, ONE_YEAR_AGO);
    log.info("Starting scan for Message entries older than: " + deadline.getTime());
// ...
```
```java
private void purgeOldMessages() {
    Date date = this.getModifiedDate(Calendar.YEAR, -1);

    log.info("Starting scan for Message entries older than: " + date);
```
A new approach may still be desired, but once again, that's not what I'm focusing on right


frequential:
Created a draft to ask the following.

The `BaseDAO` contains a [piece of code that is repeated 3 times](https://github.com/frequential/akvo-flow/blob/cleanup/GAE/src/com/gallatinsystems/framework/dao/BaseDAO.java#L271).
```java
if (cursorString != null && !cursorString.trim().equalsIgnoreCase(Constants.ALL_RESULTS)) {
    Cursor cursor = Cursor.fromWebSafeString(cursorString);
    Map<String, Object> extensionMap = new HashMap<>();
    extensionMap.put(JDOCursorHelper.CURSOR_EXTENSION, cursor);
    query.setExtensions(extensionMap);
}
 ```
If it were to be extracted, what would be an appropriate method name?

File: GAE/src/org/waterforpeople/mapping/domain/QuestionAnswerStore.java [hunk 1]
```diff
@@ -35,7 +35,7 @@ public class QuestionAnswerStore extends BaseDomain {
     private static final long serialVersionUID = 3726562582080475960L;
 
     @Persistent
-    private Long arbitratyNumber;
+    private Long arbitraryNumber;
```
frequential, line 38: This change may not be possible for the database.

muloem: This property will be removed. You can mark that with a TODO.

GAE/src/org/waterforpeople/mapping/app/web/CronCommanderServlet.java
```diff
@@ -236,182 +263,196 @@ private void purgeOrphanJobQueueRecords() {
...
 
+
     /**
      * scans for and deletes Message entries that are more than one year old
      */
     private void purgeOldMessages() {
-        Calendar deadline = Calendar.getInstance();
-        deadline.add(Calendar.YEAR, ONE_YEAR_AGO);
-        log.info("Starting scan for Message entries older than: " + deadline.getTime());
+        Date date = this.getModifiedDate(Calendar.YEAR, -1);
```
muloem, line 432:
```suggestion
        Date cutOffDate = this.getModifiedDate(Calendar.YEAR, -1);
```

muloem (changes requested):

